PR #3, opened 2026-01-11
Add Browserbase support for remote browser over CDP

pkiv:
This is great! Our customers would love to use this, so I added a simple Browserbase integration to connect to remote browsers if desired

When BROWSERBASE_API_KEY and BROWSERBASE_PROJECT_ID env vars are set, connect to a Browserbase session via CDP instead of launching a local browser.

Tested with every command. Looks like `session` and `session list` aren't implemented yet. I have a draft implementation here: https://github.com/browserbase/agent-browser/pull/1

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1440,8 +1440,6 @@ export class BrowserManager {
       this.cdpSession = null;
     }
 
-    // CDP: only disconnect, don't close external app's pages
-    if (this.cdpPort !== null) {
     if (this.browserbaseSessionId && this.browserbaseApiKey) {
       await this.closeBrowserbaseSession(this.browserbaseSessionId, this.browserbaseApiKey).catch(
         (error) => {


maxjendrall:
I'd suggest making this PR CDP url agnostic and not biased towards using browserbase.

ctate:
Hey @pkiv! Thanks for this PR and for being an early contributor 🙏

I love the idea of making Browserbase integration seamless. After reviewing this alongside some other PRs (we just merged #24 for generic CDP support), I think we can make this even better.

Two thoughts:

1. Explicit provider flag instead of auto-detection

The implicit auto-detection when env vars are set could surprise users. What do you think about an explicit flag?

```bash
# Explicit flag
agent-browser -p browserbase open https://example.com

# Or env var for CI/scripts
export AGENT_BROWSER_PROVIDER=browserbase
```

This makes behavior discoverable and follows our existing patterns (--session / AGENT_BROWSER_SESSION).

2. Skip the SDK dependency

Since we just merged CDP support (#24), we could keep agent-browser dependency-free by using Browserbase's REST API directly:

```ts
const res = await fetch('https://api.browserbase.com/v1/sessions', {
  method: 'POST',
  headers: {
    'Authorization': `Bearer ${apiKey}`,
    'Content-Type': 'application/json'
  },
  body: JSON.stringify({ projectId })
});
const { connectUrl } = await res.json();
// Reuse existing CDP connection
```

This means zero added dependencies for users, and lower friction to try Browserbase.

If you'd prefer to keep the SDK for richer features down the road, we could add it as an optionalDependency instead - users would just need to pnpm add @browserbasehq/sdk to enable it.

Kylejeong2:
Hey @ctate, I'll update the code to use the API directly instead of the SDK to keep you guys at zero deps.

ctate:
Looks great! Can you rebase on `main`?

Kylejeong2:
> Looks great! Can you rebase on `main`?

rebased!

ctate:
Thanks @Kylejeong2 - I test this locally and running into an error:

```
Build fails with: src/browser.ts(1486,1): error TS1005: '}' expected
```

Are you seeing this on your end?

Kylejeong2:
> Thanks @Kylejeong2 - I test this locally and running into an error:
> 
> ```
> Build fails with: src/browser.ts(1486,1): error TS1005: '}' expected
> ```
> 
> Are you seeing this on your end?

Fixed and built locally, it works!

ctate:
@Kylejeong2 Thank you so much! Looks great

Issue #8, opened 2026-01-12
CLI stripping --interactive and similar snapshot flags


coty:
The --interactive flag for the snapshot command is silently ignored, while the short form -i works correctly.

 # Steps to Reproduce

 ## Open a page
  agent-browser open example.com

 ## These produce different results:
  agent-browser snapshot -i --json
  agent-browser snapshot --interactive --json

  # Expected Behavior

  Both commands should produce identical output, showing only interactive elements (links, buttons, inputs).

  # Actual Behavior

  - snapshot -i → Returns only interactive elements (correct)
  - snapshot --interactive → Returns full snapshot including non-interactive elements (incorrect)

  # Root Cause

  In cli/src/flags.rs, the clean_args function strips all arguments starting with -- before the command parser sees them:

```
  if !arg.starts_with("--") && arg != "-f" {
      result.push(arg.clone());
  }
```
  
This was intended to remove global flags like --json and --headed, but it also removes command-specific flags like --interactive, --compact, --depth, and --selector.

 # Affected Flags

  All long-form snapshot flags are broken:
  - --interactive (works as -i)
  - --compact (works as -c)
  - --depth (works as -d)
  - --selector (works as -s)

ctate:
Fixed in https://github.com/vercel-labs/agent-browser/pull/9

Thanks again

coty:
Resolved in #9

PR #9, opened 2026-01-12
Fix CLI stripping --interactive and similar snapshot flags

coty:
## Summary

Fixes Issue #8.

  - Fixed `clean_args()` stripping all `--` prefixed arguments instead of just global flags
  - Long-form snapshot flags (`--interactive`, `--compact`, `--depth`, `--selector`) now work correctly

  ## Problem

  The `clean_args` function was designed to strip global flags before passing args to the command parser, but it was too aggressive:

  ```rust
  // Before: strips ALL --prefixed args
  if !arg.starts_with("--") && arg != "-f" {
      result.push(arg.clone());
  }
```

  This caused snapshot --interactive to be silently ignored while snapshot -i worked fine.

  ## Solution

  Changed to explicitly list global flags to strip, allowing command-specific flags through:

  ```rust
  // After: only strips known global flags
  const GLOBAL_FLAGS: &[&str] = &["--json", "--full", "--headed", "--debug"];

  if GLOBAL_FLAGS.contains(&arg.as_str()) || arg == "-f" {
      continue;
  }
  result.push(arg.clone());
```

  ## Testing

  ### Both now produce identical output:
  agent-browser snapshot -i --json
  agent-browser snapshot --interactive --json

  All existing tests pass (84/84).

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -4,7 +4,7 @@ version = 4
 
 [[package]]
 name = "agent-browser"
-version = "0.1.0"
+version = "0.4.0"
 dependencies = [
  "libc",
  "serde",


coty:
NB: This also update Cargo.lock to match the version in Cargo.toml because that was updated while buildign and testing.

ctate:
Great catch! Thank you for the fix

PR #16, opened 2026-01-12
feat: add --proxy flag for browser proxy configuration

nmwcode:
## Summary

Adds a `--proxy` flag to route all browser traffic through a proxy server.

## Changes

- Add `--proxy` flag parsing in CLI (Rust)
- Parse proxy URLs with credentials (http://user:pass@host:port)
- Pass proxy config to Playwright's newContext()
- Add Zod validation for proxy schema

## Usage

```bash
# Simple proxy
agent-browser --proxy http://proxy.com:8080 open example.com

# With authentication
agent-browser --proxy http://user:pass@proxy.com:8080 open example.com
```

Supports HTTP, HTTPS, and SOCKS5 protocols.

## Background

Playwright requires proxy configuration at context creation time and does not respect HTTP_PROXY/HTTPS_PROXY environment variables.

## Testing

Tested with authenticated and unauthenticated proxies. Verified IP changes using httpbin.org/ip.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -24,37 +24,32 @@ use install::run_install;
 use output::{print_command_help, print_help, print_response};
 
 fn parse_proxy(proxy_str: &str) -> serde_json::Value {
-    // Parse URL format: http://user:pass@host:port or http://host:port
     let Some(protocol_end) = proxy_str.find("://") else {
         return json!({ "server": proxy_str });
     };
     let protocol = &proxy_str[..protocol_end + 3];
-
-    // Check for credentials (user:pass@host format)
     let rest = &proxy_str[protocol_end + 3..];
+
     let Some(at_pos) = rest.rfind('@') else {
         return json!({ "server": proxy_str });
     };
 
     let creds = &rest[..at_pos];
     let server_part = &rest[at_pos + 1..];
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for adding proxy support - this is a useful feature!

A few things to address:

1. Proxy parsing edge case

As the bot mentioned, username-only proxies silently drop the username:

```
let Some(colon_pos) = creds.find(':') else {
  // Found @ but no : in credentials - return server without credentials
  let server_part = &rest[at_pos + 1..];
  return json!({ "server": format!("{}{}", protocol, server_part) });
};
```

For http://username@host:port, this returns just `{"server": "http://host:port"}` - the username is lost. Either preserve it or return an error explaining that password is required.

2. Missing tests

The parse_proxy() function should have unit tests covering:
- Basic proxy without auth: http://host:port
- Proxy with auth: http://user:pass@host:port
- Username only: http://user@host:port
- No protocol: host:port
- SOCKS5: socks5://host:port

3. Missing help output

Add `--proxy` to the help text in cli/src/output.rs so users can discover it.

nmwcode:
@ctate done!

ctate:
Pulled this down, proxy flag works as expected. Nice addition.

PR #18, opened 2026-01-12
fix typo

ctate:


ngchenkai:
**Summary: Dashboard live viewport — headless/headed & externally-driven browsers**

**Q1: Does the dashboard's live viewport work for headless sessions, headed, or both?**
Both. The viewport is fed by CDP `Page.startScreencast` (`cli/src/native/stream/cdp_loop.rs:299`), which captures from the page's render pipeline, not the physical screen. `handle_screencast_start` (`cli/src/native/actions.rs:8030`) never checks headless/headed or who launched the browser; the only engine restriction is Lightpanda being headless-only (`cli/src/native/browser.rs:81`). The idle-timeout docs also group "headed" and "user-attached" browsers with dashboard input (`cli/src/output.rs:3820`).

**Q2: Can the dashboard show a browser driven by another tool (e.g. browser-use/browser-harness)?**
Yes, via attach-not-launch. Chrome supports multiple concurrent CDP clients, so:
1. Other tool launches/drives Chrome with remote debugging on (e.g.) 9222
2. `agent-browser connect 9222` (or `--cdp <port|url>`, or `--auto-connect` which discovers via `DevToolsActivePort` then probes 9222/9229 — `cli/src/native/cdp/chrome.rs:956`)
3. `agent-browser dashboard start` → live viewport of the same tabs

Caveats: the command activity feed only logs commands that went through agent-browser; two drivers can race on tab focus/input; Chrome M144+ may prompt per attaching client.

**Q3: Connecting to a headless browser specifically**
```bash
google-chrome --headless=new --remote-debugging-port=9222 --user-data-dir=/tmp/profile --no-first-run about:blank
agent-browser connect 9222
agent-browser dashboard start
```
(`--user-data-dir` is effectively required: Chrome 136+ blocks debugging on the default profile.) Attached browsers are exempt from the default idle shutdown (`browser_process.is_none()`), so the daemon won't kill an externally-launched browser.

PR #31, opened 2026-01-12
feat: Add popup/new window detection for JavaScript-opened pages

cipherchabon:
## Summary

This PR adds the ability to detect and interact with pages opened by JavaScript (`window.open()`, `target="_blank"`, etc.), which was previously not possible with the existing tab management system.

### Problem

When automating web applications that open new windows/tabs via JavaScript, the current implementation cannot detect these new pages. The `tab list` command only shows explicitly created tabs, missing any popups triggered by user interactions.

### Solution

Added a `context.on('page')` event listener that captures all new pages created within a browser context, along with new commands to wait for and manage these popups.

### Core Implementation (`src/browser.ts`)
- Added `PopupEvent` interface for tracking popup metadata
- Added `context.on('page')` listener in `launch()` to detect JavaScript-opened pages
- Implemented `handlePopup()` to track new pages with their opener URL and timestamp
- Added `waitForPopup(timeout?)` - waits for and returns the next popup
- Added helper methods: `getPopupHistory()`, `getLastPopup()`, `clearPopupHistory()`, `hasPopups()`
- Fixed `newTab()`, `newWindow()`, and `launch()` to prevent explicit page creations from being recorded as popups (using `isCreatingExplicitPage` flag)

### Types (`src/types.ts`)
- Added `PopupWaitCommand`, `PopupListCommand`, `PopupClearCommand` interfaces
- Added `PopupInfo`, `PopupWaitData`, `PopupListData` response types

### Protocol (`src/protocol.ts`)
- Added Zod schemas for `popup_wait`, `popup_list`, `popup_clear` commands

### Actions (`src/actions.ts`)
- Added handlers: `handlePopupWait`, `handlePopupList`, `handlePopupClear`

### CLI (`cli/src/commands.rs`)
- Added new commands:
  - `popup wait [timeout]` - Wait for a popup to be opened
  - `popup list` - List all detected popups
  - `popup clear` - Clear popup history

## Usage

```bash
# Click something that opens a popup, then capture it
click "button#open-popup"
popup wait 5000

# List all detected popups
popup list

# Switch to the popup tab
tab 1

# Clear popup history
popup clear
```

## Test Plan

- [x] All existing tests pass (84/84)
- [x] Tested with real-world site that uses `target="_blank"` links
- [x] Verified explicit `newTab()` calls don't get recorded as popups
- [x] Verified `waitForPopup()` correctly times out when no popup appears

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,31 @@ agent-browser tab close [n]           # Close tab
 agent-browser window new              # New window
 ```
 
+### Popup Detection
+
+Detect and interact with JavaScript-opened windows (window.open, target="_blank"):
+
+```bash
+agent-browser popup wait [timeout]    # Wait for popup (default 30s)
+agent-browser popup list              # List detected popups
+agent-browser popup clear             # Clear popup history
+```
+
+**Example workflow:**
+
+```bash
+# Click a link that opens in new tab
+agent-browser click "a[target='_blank']"
+
+# Wait for the popup to open
+agent-browser popup wait
+# Returns: {"index": 1, "url": "https://...", "switched": true}
+
+# Now on the new tab - interact normally
+agent-browser snapshot -i
+agent-browser click @e1
+```
+
 ### Frames
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ agent-browser window new              # New window
 Detect and interact with JavaScript-opened windows (window.open, target="_blank"):
 
 ```bash
-agent-browser popup wait [timeout]    # Wait for popup (default 30s)
+agent-browser popup wait [timeout]    # Wait for popup (default 5s)
 agent-browser popup list              # List detected popups
 agent-browser popup clear             # Clear popup history
 ```
@@ -217,7 +217,7 @@ agent-browser click "a[target='_blank']"
 
 # Wait for the popup to open
 agent-browser popup wait
-# Returns: {"index": 1, "url": "https://...", "switched": true}
+# Returns: {"index": 1, "url": "https://...", "title": "Page Title"}
 
 # Now on the new tab - interact normally
 agent-browser snapshot -i
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -347,6 +347,6 @@ checksum = "589f6da84c646204747d1270a2a5661ea66ed1cced2631d546fdfb155959f9ec"
 
 [[package]]
 name = "zmij"
-version = "1.0.12"
+version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2fc5a66a20078bf1251bde995aa2fdcc4b800c70b5d92dd2c62abc5c60f679f8"
+checksum = "bd8f3f50b848df28f887acb68e41201b5aea6bc8a8dacc00fb40635ff9a72fea"
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -643,6 +643,17 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
             }
         }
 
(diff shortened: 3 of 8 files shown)


cipherchabon:
Thanks for the review feedback! I've addressed all 4 issues:

1. **Race condition fix**: Replaced `isCreatingExplicitPage` boolean flag with a `WeakSet<Page>` to track explicitly created pages. This handles async 'page' events correctly since the page object identity is preserved.

2. **Memory leak fix**: Added cleanup logic in `waitForPopup()` to remove orphaned resolvers from `popupResolvers` array when timeout fires.

3. **README timeout**: Fixed documentation to show correct default (5s instead of 30s).

4. **README example**: Fixed example output to show `title` field instead of non-existent `switched` field.

All 107 tests pass.

ctate:
Hey @cipherchabon! Nice work on this - popup detection is a great addition.

Tested it locally and it works well. Just found two small issues:

1. popup wait without timeout errors

```
$ agent-browser popup wait
{"success":false,"error":"Validation error: timeout: Expected number, received null"}
```

The README says it defaults to 5s, but looks like the schema requires a number. Should either make timeout optional with a default, or update the docs (likely the former).

2. Popup commands missing from `--help`

The commands work but don't show up in agent-browser `--help`. Might want to add them to output.rs.

Other than that, looks good to merge!

cipherchabon:
Thanks for the review @ctate! Both issues have been fixed:

1. popup wait timeout validation error
Fixed in commands.rs - the CLI was sending "timeout": null when not provided, but Zod expected the field to be omitted entirely. Now the timeout field is only included when explicitly provided, so the TypeScript side correctly uses the 5s default.

2. Missing from --help
Added popup commands to both --help output and popup --help detailed documentation.

Also added 5 unit tests for the popup commands.

ctate:
Looks good! Do you mind resolving conflicts so I can merge this in?

cipherchabon:
sure!

ctate:
Hey @cipherchabon - so sorry. There's merge conflicts again 😅

cipherchabon:
No problem @ctate! done

ctate:
@cipherchabon I see in commit e79cc13 you switched from the isCreatingExplicitPage flag to the WeakSet approach. However, I'm still seeing the bug after testing the latest code:

```
$ agent-browser popup clear
$ agent-browser tab new
$ agent-browser popup list --json
# Shows: explicit tab incorrectly in popup list ❌
```

The WeakSet approach still has the race condition because the page is added to explicitlyCreatedPages after context.newPage(), but the event fires during the await.

The original flag-based approach from earlier commits was actually correct. Could you restore it?

```
// This works (synchronous check):
this.isCreatingExplicitPage = true;  // Set BEFORE async call
const page = await context.newPage();
this.explicitlyCreatedPages.add(page);  // Keep WeakSet too for safety
this.isCreatingExplicitPage = false;

// In handlePopup():
if (this.isCreatingExplicitPage || this.explicitlyCreatedPages.has(newPage)) {
  return;
}
```

The flag provides synchronous protection during the async operation, while the WeakSet provides long-term tracking.

cipherchabon:
Thanks for catching this @ctate! You're absolutely right about the race condition.

Fixed in a990592 - I restored the flag-based approach as you suggested:

this.isCreatingExplicitPage = true;     // Set BEFORE async call
const page = await context.newPage();
this.explicitlyCreatedPages.add(page);  // WeakSet for long-term tracking
this.isCreatingExplicitPage = false;

And handlePopup() now checks the flag first:

if (this.isCreatingExplicitPage) {
  return;  // Skip - explicit page creation in progress
}

The flag provides synchronous protection during the async window, while the WeakSet handles long-term tracking without memory leaks.

ctate:
Thanks for reporting this @cipherchabon. This has been fixed

PR #45, opened 2026-01-13
Fix CLI/protocol mismatches for select, frame main, and headers

Nicenonecb:
## Description
Align select to send values instead of value, matching the protocol and execution layer (commands.rs).
Map frame main to the correct mainframe action to avoid command mismatches (commands.rs).
Allow headers on navigate in the protocol schema so open --headers validates correctly (protocol.ts).
Add/extend parsing tests to cover the corrected behaviors (commands.rs, protocol.test.ts).
## Test
Cargo.toml
protocol.test.ts

wechat_2026-01-13_104305_780
wechat_2026-01-13_104316_236

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -153,13 +153,18 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
         "select" => {
             let sel = rest.get(0).ok_or_else(|| ParseError::MissingArguments {
                 context: "select".to_string(),
-                usage: "select <selector> <value>",
+                usage: "select <selector> <value...>",
             })?;
             let _val = rest.get(1).ok_or_else(|| ParseError::MissingArguments {
                 context: "select".to_string(),
-                usage: "select <selector> <value>",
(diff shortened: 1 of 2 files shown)


ctate:
Nice catches on these CLI/protocol mismatches! The frame main → mainframe and navigate headers fixes look correct.

One question on the select change:
```
Ok(json!({ "id": id, "action": "select", "selector": sel, "values": rest[1..].join(" ") }))
```

The protocol accepts either a string or array for values:
```
values: z.union([z.string(), z.array(z.string())]),
```

By joining with space, select #menu opt1 opt2 opt3 sends `"opt1 opt2 opt3"` as a single string. Should this instead send `["opt1", "opt2", "opt3"]` as an array for multi-select dropdowns? Or is single-string the intended behavior here?

Nicenonecb:
@ctate  Thanks for the catch. I updated `select` to send a string for a single value and an array for multiple values, matching the protocol; also updated the CLI usage and tests. 
Let me know if this matches your expectations or if you’d like any further adjustments.

ctate:
This looks great! Thank you

PR #63, opened 2026-01-13
feat: `--video [dir]` support

tmm:
wires up `--video` flag (via `recordVideo`) for video recording

```bash
agent-browser --video /tmp/videos open example.com
agent-browser close  # video saved as .webm
```


vesper8:
video recording now available via https://github.com/vercel-labs/agent-browser/pull/116

not yet tagged in new release but just tried it by building from source and it works

PR #79, opened 2026-01-14
fix(cli): fix windows daemon startup and port calculation inconsistency

0okay:
Fixes inconsistent port calculation on Windows (using unsigned_abs) and improves daemon startup reliability by avoiding cmd /c.

Issue #56, opened 2026-01-13
windows Daemon failed


TPLink32:
"C:\Users\Administrator\AppData\Roaming\npm\node_modules\agent-browser\bin\agent-browser-win32-x64.exe" open https://google.com --debug

C:\Users\Administrator\Desktop\agent>[0m Daemon failed to start

EricLeeK:
same

0okay:
+1

pehcastro:
+1, daemon error.

only was able to even run it with this workaround https://github.com/vercel-labs/agent-browser/issues/25#issuecomment-3747532705 

started with `npm install -g agent-browser` then had to point git bash inside the .ps1 file

<img width="662" height="431" alt="Image" src="https://github.com/user-attachments/assets/dab8eb16-05c0-45a0-86ab-96bf5a2a927c" />

tried local install (clone + build), it don't output a index.js, build has no trace logs.

then clone + build:windows docker also fails middle build

xiansheng50:
+1

alice-chowy:
+1

Chekhovin:
+1

Mtartup:
+1

hashimwarren:
+1

golamt313:
+1

bearyue:
+1

zx9597446:
+1

kamikakushi:
same here
<img width="1777" height="911" alt="Image" src="https://github.com/user-attachments/assets/a45f55b8-b7ef-4cbe-b3d4-20711366c015" />

tercel53:
+1

mgalihpp:
+1

toanpi:
+1

slender9168:
+1

coofly:
+1

thameraltoimi:
hello, i found a workaround. just start the daemon directly, then you can run agent-browser commands normally:
  ```
node $env:APPDATA\npm\node_modules\agent-browser\dist\daemon.js

CXRui:
<img width="1023" height="556" alt="Image" src="https://github.com/user-attachments/assets/24312d2b-30eb-45bc-acb8-ba36c4b23e5e" />

giulio-leone:
Duplicate of #89. Windows daemon failures addressed in PRs #440, #448.

ctate:
These Windows daemon failures should be addressed in #582, which fixes three root causes:

[...]

PR #88, opened 2026-01-14
fix "Daemon failed to start" error on Windows

JustinLeei:
The original code cast hash.abs() to u16 before taking modulo, which truncates large i32 values and causes incorrect port calculations. Now takes modulo on the full i32 value first, then safely casts to u16.


JustinLeei:
fix this issue #56 #88

ctate:
Thanks so much for tracking down this port calculation bug

This turned out to be a duplicate of https://github.com/vercel-labs/agent-browser/pull/79 which we just merged. Both PRs identified the same casting issue - your analysis was spot on about the truncation problem.

Really appreciate you taking the time to contribute

PR #92, opened 2026-01-14
fix: respect AGENT_BROWSER_HEADED env var for headed mode

### LINKED ISSUE #90: Windows CLI fails to start daemon + headed mode ignored
## Summary

Two bugs affecting Windows users that prevent the agent-browser from working correctly:

1. **Headed mode is ignored** - The `AGENT_BROWSER_HEADED=1` environment variable has no effect
2. **Windows CLI fails with "Daemon failed to start"** - Path escaping issues when spawning the Node.js daemon

---

### Location
`src/daemon.ts` (compiled as `dist/daemon.js`, around line 180)

### Problem
The daemon has `headless: true` hardcoded, completely ignoring the `AGENT_BROWSER_HEADED` environment variable:

```typescript
// Current code
const browser = await chromium.launch({
  headless: true,  // Always headless, ignoring env var
  // ...
});
```

### Expected Behavior
Setting `AGENT_BROWSER_HEADED=1` should launch the browser in headed mode for debugging.

### Fix
```typescript
const browser = await chromium.launch({
  headless: process.env.AGENT_BROWSER_HEADED !== '1',
  // ...
});
```

---

### Location
`cli/src/connection.rs`

### Problem
On Windows, the Rust CLI spawns the Node.js daemon using `cmd /c`, which causes path escaping issues:

```rust
// Current approach (problematic on Windows)
Command::new("cmd")
    .arg("/c")
    .arg(&cmd_string)  // cmd_string = "node \"C:\path\to\daemon.js\""
```

This results in the path being mangled. The error message shows:

```
Cannot find module 'C:\Users\username\project\"C:\Users\username\AppData\Roaming\npm\node_modules\@anthropic-ai\agent-browser\dist\daemon.js"'
```

Notice how:
- The CWD gets prepended to the path
- The quotes become part of the filename instead of escaping the path

### Steps to Reproduce
1. Install agent-browser on Windows: `npm install -g @anthropic-ai/agent-browser`
2. Run any command that starts the daemon
3. Observe "Daemon failed to start" error

### Fix
Call Node.js directly instead of through `cmd.exe`:

```rust
// Instead of:
Command::new("cmd")
    .arg("/c")
    .arg(&cmd_string)

// Use:
Command::new("node")
    .arg(&daemon_path)
    .spawn()
```

This avoids the shell escaping issues entirely.

---

## Environment
- OS: Windows 10/11
- agent-browser version: latest from npm
- Node.js: v20+

## Impact
These bugs completely prevent Windows users from:
1. Using the agent-browser CLI at all (Bug 2)
2. Debugging browser automation in headed mode (Bug 1)

---

*Found while integrating agent-browser with Claude Code on Windows.*

Sanchay-T:
## Summary

- Fixes hardcoded `headless: true` in the auto-launch section of `daemon.ts`
- Now respects the `AGENT_BROWSER_HEADED` environment variable
- Setting `AGENT_BROWSER_HEADED=1` will launch the browser in headed mode

## The Bug

In `src/daemon.ts`, line 203, the `headless` option was hardcoded to `true` in the auto-launch section. This meant the `AGENT_BROWSER_HEADED` environment variable was being ignored when the browser was auto-launched.

## The Fix

Changed:
```typescript
headless: true,
```

To:
```typescript
headless: process.env.AGENT_BROWSER_HEADED !== '1',
```

This is consistent with how the environment variable is documented and expected to work.

Fixes #90

## Test plan

- [ ] Set `AGENT_BROWSER_HEADED=1` and verify browser launches in headed mode
- [ ] Without the env var set, verify browser still launches in headless mode (default)
- [ ] Test on Windows where this issue was originally reported

### DIFF
--- a/src/daemon.ts
+++ b/src/daemon.ts
@@ -200,7 +200,7 @@ export async function startDaemon(options?: { streamPort?: number }): Promise<vo
             await browser.launch({
               id: 'auto',
               action: 'launch',
-              headless: true,
+              headless: process.env.AGENT_BROWSER_HEADED !== '1',
               executablePath: process.env.AGENT_BROWSER_EXECUTABLE_PATH,
               extensions: extensions,
             });


ctate:
Good catch! Tested and confirmed it works - headed mode now respects the env var during auto-launch.

PR #94, opened 2026-01-14
feat: add --version flag to CLI

mking-keeper:
## Summary
- Give users the ability to check the current version of the tool
- The version flag format (`--version`, `-V`) matches `rustc` and `ruff`.

## Usage
```
$ agent-browser --version
agent-browser 0.5.0
```

## Test plan
- [x] Run `cargo test` 
- [x] Build CLI with `cargo build --release`
- [x] Run `./target/release/agent-browser --version` and verify it outputs `agent-browser 0.5.0`
- [x] Run `./target/release/agent-browser -V` and verify it outputs `agent-browser 0.5.0`

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -21,7 +21,7 @@ use commands::{gen_id, parse_command, ParseError};
 use connection::{ensure_daemon, send_command};
 use flags::{clean_args, parse_flags};
 use install::run_install;
-use output::{print_command_help, print_help, print_response};
+use output::{print_command_help, print_help, print_response, print_version};
 
 fn run_session(args: &[String], session: &str, json_mode: bool) {
     let subcommand = args.get(1).map(|s| s.as_str());
@@ -99,11 +99,7 @@ fn main() {
     let clean = clean_args(&args);
 
     let has_help = args.iter().any(|a| a == "--help" || a == "-h");
-
-    if clean.is_empty() {
-        print_help();
-        return;
-    }
+    let has_version = args.iter().any(|a| a == "--version" || a == "-V");
 
     if has_help {
         if let Some(cmd) = clean.get(0) {
@@ -111,6 +107,16 @@ fn main() {
                 return;
             }
         }
+        print_help();
+        return;
+    }
+
+    if has_version {
+        print_version();
+        return;
+    }
+
+    if clean.is_empty() {
         print_help();
         return;
     }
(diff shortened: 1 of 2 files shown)


cli/src/flags.rs
```diff
@@ -79,7 +79,7 @@ pub fn clean_args(args: &[String]) -> Vec<String> {
     let mut skip_next = false;
 
     // Global flags that should be stripped from command args
-    const GLOBAL_FLAGS: &[&str] = &["--json", "--full", "--headed", "--debug"];
+    const GLOBAL_FLAGS: &[&str] = &["--json", "--full", "--headed", "--debug", "--version"];
```
n33pm: I don't think we need it as a global flag because it exits immediately

cli/src/flags.rs
```diff
@@ -93,7 +93,7 @@ pub fn clean_args(args: &[String]) -> Vec<String> {
             continue;
         }
         // Only strip known global flags, not command-specific flags
-        if GLOBAL_FLAGS.contains(&arg.as_str()) || arg == "-f" {
+        if GLOBAL_FLAGS.contains(&arg.as_str()) || arg == "-f" || arg == "-V" {
```
n33pm: > The GNU Coding Standards require `--version` but don't mandate a short form, emphasizing long options for clarity [gnu.org](https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html). The POSIX utility guidelines allow single alphanumeric characters (including uppercase) for options, with -W being the only reserved capital letter [opengroup.org](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html).

cli/src/flags.rs
```diff
@@ -207,4 +207,9 @@ mod tests {
         assert_eq!(flags.session, "test");
         assert_eq!(flags.executable_path, Some("/custom/chrome".to_string()));
     }
+
+    #[test]
+    fn test_clean_args_removes_version_flag() {
+        assert_eq!(clean_args(&args("--version open example.com")), vec!["open", "example.com"]);
+    }
```
n33pm: IMHO, this makes no sense; we return from main immediately after printing the version.

cli/src/flags.rs
mking-keeper: great point! updated

cli/src/flags.rs
mking-keeper: I think that's a fair point. I added `-V` to be consistent with `rustc` and `ruff`, which support the short version flag (consistency with ruff ecosystem)

ctate:
Looks great. Thank you!

PR #98, opened 2026-01-14
fix: track dynamically opened tabs in non-CDP mode

### LINKED ISSUE #74: tab command didn't list all tabs
open a website, click an element which opens a new tab. so, we could see two tabs. (with headed mode)
but tab command only output 1 tab.

Aster110:
## Summary

  Fixes #74 - `tab` command doesn't list tabs opened via `window.open()` or `target="_blank"` links.

  ## Problem

  In non-CDP launch mode, `setupContextTracking()` was only called for CDP connections, so the `context.on('page')` event listener was never registered. This caused dynamically opened tabs to not be added to the `pages` array.

  ## Solution

  Add `setupContextTracking(context)` after initial page setup in the standard launch path (line 704), ensuring new pages are tracked regardless of launch mode.

  ## Test

  ```bash
  agent-browser open "https://www.baidu.com" --headed
  agent-browser tab  # Shows 1 tab
  agent-browser eval "window.open('https://www.bing.com', '_blank'); 'done'"
  agent-browser tab  # Now correctly shows 2 tabs

  Before fix: Only showed 1 tab
  After fix: Correctly shows 2 tabs
  ```

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -701,6 +701,7 @@ export class BrowserManager {
     this.pages.push(page);
     this.activePageIndex = 0;
     this.setupPageTracking(page);
+    this.setupContextTracking(context);
   }
 
   /**


ctate:
Good catch on the window.open() bug - tested and it works.

But the fix breaks 3 tests because newTab() now double-tracks pages (once from the event listener, once manually). Need to update newTab() to not manually push since setupContextTracking handles it. Can you plz fix?

PR #101, opened 2026-01-14
fix(cli): allow null path in screenshot command validation

sheing-google:
The Rust CLI sends `null` for the `path` argument when it is not provided, but the Zod schema only accepted `undefined`. This change updates the `screenshotSchema` to allow `null` values for `path`, enabling the `screenshot` command to work without a file path argument (outputting to stdout).


ctate:
Tested both with and without path. Works great. Thanks!

Issue #87, opened 2026-01-14
CDP connection times out when Android browser has tabs with empty URL


edxeth:
# CDP connection times out when Android browser has tabs with empty URL

When connecting to an Android device via CDP (Chrome/Brave with remote debugging enabled through ADB port forwarding), agent-browser times out if the browser has any tabs with an empty URL.

## Environment

- Android device with Chrome/Brave
- ADB port forwarding: `adb forward tcp:9223 localabstract:chrome_devtools_remote`
- agent-browser v0.5.0

## Steps to reproduce

1. Set up ADB port forwarding to your Android browser's CDP socket

2. Verify CDP is accessible:
   ```sh
   curl http://localhost:9223/json/version
   ```

3. Have at least one tab that reports empty url/title in the target list:
   ```sh
   curl http://localhost:9223/json/list
   ```
   Look for entries like:
   ```json
   {"title": "", "url": "", "type": "page", ...}
   ```

4. Run agent-browser:
   ```sh
   agent-browser --cdp 9223 snapshot -i
   ```
   **Result:** hangs indefinitely, times out

5. Navigate that empty tab to any URL (or close it), then retry step 4

   **Result:** works correctly

## Root cause

Playwright's `connectOverCDP` uses `Target.setAutoAttach` which attaches to all page targets. When it attaches to the empty-URL tab and sends commands like `Page.enable`, `Page.getFrameTree`, etc., those commands never receive responses, causing the connection to hang until timeout.

Tabs with `about:blank` work fine; the issue is specifically with tabs that have completely empty URLs, which is a state that can occur on Android in certain conditions (new tab that hasn't loaded yet, system pages, etc.).

## Workaround

Ensure all tabs have loaded a URL before connecting.

sheing-google:
https://github.com/vercel-labs/agent-browser/pull/102

giulio-leone:
This is addressed in PR #445 which improves CDP connection handling. Empty-URL tabs (like `chrome://newtab`) are now properly filtered during connection, preventing timeout issues.

This issue can likely be closed.

PR #102, opened 2026-01-14
fix: Prevent CDP timeout on empty URL tabs

sheing-google:
When connecting to a browser via CDP, particularly on Android, tabs with an empty URL can cause Playwright commands to hang indefinitely. This leads to a timeout in agent-browser.

Address: https://github.com/vercel-labs/agent-browser/issues/87

This commit fixes the issue by filtering out any pages that have an empty `page.url()` during the CDP connection process. This prevents agent-browser from attempting to interact with these problematic tabs, resolving the timeout while preserving normal pages.

Added a unit test to verify that pages with empty URLs are correctly ignored. Also increased the timeout for a flaky screencast test to improve test suite stability.

image


ctate:
Smart fix. Thanks!

PR #116, opened 2026-01-15
feat: Add video recording with Playwright native video

davekiss:
# Summary

Adds record start, record stop, and record restart commands for capturing browser sessions to video.

## Why context-based recording instead of `--video` launch flag? (https://github.com/vercel-labs/agent-browser/pull/63)

Playwright supports a `--video` flag at browser launch, but this doesn't work well with agent-browser's daemon architecture:

  - If the daemon is already running, the flag is silently ignored (the launch command no-ops)
  - New windows/contexts don't inherit the video setting
  - Recording is tied to context lifecycle - no way to start/stop mid-session

You _could_ work around this by saving state, closing the browser, relaunching with `--video`, and loading state back - but that's 5 extra commands, temp file management, and browser restart latency.

Instead, this record start/stop approach requires just 2 commands, handles state transfer automatically, and avoids the restart overhead.

## Why Playwright native instead of FFmpeg?

I initially implemented video recording using FFmpeg + CDP screencast. It provided more control (start/stop anytime, MP4 output) but required users to install FFmpeg. For a library targeting many developers, that dependency is significant friction. Playwright's native `recordVideo` context option requires zero additional dependencies and produces good-quality WebM output.

## Why auto-capture URL and transfer state?

The typical workflow is: explore a site, plan your actions, then record a clean execution. To minimize commands:

  - If no URL is provided, we capture the current page's URL
  - We transfer cookies and localStorage automatically

  This lets agents start recording with a single command: `record start ./demo.webm`

## Changes

  - `src/browser.ts`: Recording lifecycle management with context creation/destruction
  - `src/protocol.ts`, src/types.ts: Schema and types for recording commands
  - `src/actions.ts`: Command handlers
  - `cli/src/commands.rs`: CLI parsing for record start|stop|restart
  - `cli/src/output.rs`: Response formatting and help text
  - `skills/agent-browser/SKILL.md`: Documentation

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -957,6 +957,7 @@ mod tests {
             executable_path: None,
             extensions: Vec::new(),
             cdp: None,
+            proxy: None,
         }
     }
 
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -174,16 +174,23 @@ pub fn ensure_daemon(
     let exe_path = env::current_exe().map_err(|e| e.to_string())?;
     let exe_dir = exe_path.parent().unwrap();
 
-    let daemon_paths = [
+    let mut daemon_paths = vec![
         exe_dir.join("daemon.js"),
         exe_dir.join("../dist/daemon.js"),
         PathBuf::from("dist/daemon.js"),
     ];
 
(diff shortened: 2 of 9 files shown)


ctate:
Amazing. Thank you so much!

PR #138, opened 2026-01-17
feat: add Browser Use cloud browser as available provider

reformedot:


src/browser.ts
```diff
@@ -642,6 +644,80 @@ export class BrowserManager {
...
+      throw new Error('BROWSER_USE_API_KEY is required when using browseruse as a provider');
+    }
+
+    const response = await fetch('https://api.browser-use.com/api/v2/browsers', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-Browser-Use-API-Key': browserUseApiKey,
+      },
+      body: JSON.stringify({}),
```
LarsenCundric, line 756: Do we want to expose proxies, profiles and timeout?

LarsenCundric: Unsure if this aligns with the Vercel team, but I think it could be useful 😄

ctate:
Thank you so much for adding this!

PR #142, opened 2026-01-17
feat: add 'get styles' command for computed styles extraction

dannote:
**Problem**

Analyzing element styles requires writing custom JS in `eval`:

```bash
agent-browser eval "JSON.stringify(Array.from(document.querySelectorAll('button')).map(el => getComputedStyle(el)...))"
```

**Solution**

New `get styles <selector>` command returns computed styles, bounding box, and text for matched elements:

```bash
agent-browser get styles button
agent-browser get styles @e1
```

**Example**

```
[0] button "Вход через email"
    box: 228x44 at (820, 535)
    font: 16px 500 PT Root UI
    color: oklch(0.371 0 0)
    background: rgb(255, 255, 255)
    border-radius: 12px

[1] button "Нет аккаунта?"
    box: 228x20 at (820, 603)
    font: 14px 400 PT Root UI
    color: oklch(0.556 0 0)
    background: rgba(0, 0, 0, 0)
```

Text is included to identify elements when multiple match the selector.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -549,7 +549,7 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
 }
 
 fn parse_get(rest: &[&str], id: &str) -> Result<Value, ParseError> {
-    const VALID: &[&str] = &["text", "html", "value", "attr", "url", "title", "count", "box"];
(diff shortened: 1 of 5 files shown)


ctate:
Nice addition - tested and working well. Thanks!

PR #166, opened 2026-01-19
fix(windows): prevent CMD window popup when spawning daemon

### LINKED ISSUE #159: appear cmd-process during running
【Environment】：windows11
【Question】：During operation, there is a black window for the cmd program, and it does not automatically terminate the cmd process.

Gerkinfeltser:
## Summary

Fixes #159 - CMD window appearing when running agent-browser on Windows.

## Problem

When spawning the Node.js daemon on Windows, a visible CMD window would appear and persist. This happens because `DETACHED_PROCESS` and `CREATE_NO_WINDOW` flags conflict when used together for console applications like `node.exe`.

## Solution

Use only `CREATE_NO_WINDOW` (0x08000000) instead of combining `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`.

## Testing

- Built and tested on Windows 11
- Verified that `agent-browser open <url>` no longer spawns a visible CMD window
- Daemon still runs correctly in the background

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -256,11 +256,11 @@ pub fn ensure_daemon(
             cmd.env("AGENT_BROWSER_EXTENSIONS", extensions.join(","));
         }
 
-        // CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS
-        const CREATE_NEW_PROCESS_GROUP: u32 = 0x00000200;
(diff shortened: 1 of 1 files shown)


Wadeck:
For the record, I tested the patch locally, compiled, it works perfectly fine, no longer the annoying terminal opening!
On Windows 10 !

Thanks!

PR #180, opened 2026-01-20
fix: use ~/.agent-browser for socket files instead of TMPDIR

### LINKED ISSUE #163: agent-browser fails to start in Claude Code environment due to TMPDIR mismatch
## Environment

- **OS**: macOS (Darwin 23.2.0, arm64)
- **agent-browser version**: 0.6.0
- **Node.js version**: v20.19.2
- **Environment**: Claude Code CLI (sandboxed environment)
- **TMPDIR**: `/tmp/claude` (set by Claude Code)

## Description

When running `agent-browser` commands in the Claude Code environment, the browser fails to launch with the error:

```
✗ Browser not launched. Call launch first.
```

This occurs even after running `agent-browser install` successfully.

## Root Cause

The issue is caused by a mismatch in temporary directory paths:

1. **Claude Code environment** sets `TMPDIR=/tmp/claude` for sandboxing
2. **agent-browser daemon** creates socket files in `$TMPDIR`:
   - Socket: `$TMPDIR/agent-browser-default.sock`
   - PID file: `$TMPDIR/agent-browser-default.pid`
3. When the daemon process and CLI client use different `TMPDIR` values, they cannot communicate via the Unix socket

## Steps to Reproduce

1. Set up Claude Code environment (or manually set `TMPDIR=/tmp/claude`)
2. Install agent-browser: `npm install -g agent-browser`
3. Install browser binaries: `agent-browser install`
4. Try to open a page: `agent-browser open https://example.com`
5. Observe error: `✗ Browser not launched. Call launch first.`

## Expected Behavior

The agent-browser daemon should start automatically and the CLI should connect successfully, regardless of the `TMPDIR` value.

## Actual Behavior

The daemon fails to start or the CLI cannot connect to the daemon due to socket path mismatch.

## Workaround

Explicitly set `TMPDIR=/tmp` when running agent-browser commands:

```bash
TMPDIR=/tmp agent-browser open https://example.com
TMPDIR=/tmp agent-browser snapshot -i
```

Or create a shell alias:
```bash
alias agent-browser='TMPDIR=/tmp agent-browser'
```

### Option 1: Use a fixed socket location
Instead of using `$TMPDIR`, use a fixed location like `/tmp` for socket files on Unix systems. This ensures consistency across different environments.

### Option 2: Store TMPDIR in daemon state
When the daemon starts, store the `TMPDIR` value in a known location (e.g., `/tmp/agent-browser-default.tmpdir`), so the CLI can read it and connect to the correct socket path.

### Option 3: Support multiple socket locations
The CLI could try multiple common locations:
1. `$TMPDIR/agent-browser-default.sock`
2. `/tmp/agent-browser-default.sock`
3. `/var/tmp/agent-browser-default.sock`

### Option 4: Environment variable for socket path
Add an environment variable like `AGENT_BROWSER_SOCKET_PATH` to explicitly specify the socket location, overriding the default `$TMPDIR` behavior.

## Additional Context

This issue affects any environment that modifies `TMPDIR`, including:
- Claude Code CLI
- Docker containers with custom temp directories
- CI/CD environments with isolated temp directories
- Any sandboxed execution environment

## Related Code

The socket path is determined in `dist/daemon.js`:

```javascript
export function getSocketPath(session) {
    const sess = session ?? currentSession;
    if (isWindows) {
        return String(getPortForSession(sess));
    }
    return path.join(os.tmpdir(), `agent-browser-${sess}.sock`);
}
```

Where `os.tmpdir()` respects the `TMPDIR` environment variable.

## Verification

After applying the workaround (`TMPDIR=/tmp`), the daemon starts successfully:

```bash
$ ls -la /tmp/agent-browser*
-rw-r--r--  1 user  wheel  5 Jan 19 20:04 agent-browser-default.pid
srwxr-xr-x  1 user  wheel  0 Jan 19 20:04 agent-browser-default.sock
```

And commands work as expected:
```bash
$ TMPDIR=/tmp agent-browser open https://example.com
✓ Example Domain
  https://example.com
```

mmhiyoko:
## Summary

  Fix socket path to use `~/.agent-browser/` instead of `TMPDIR`. Fixes #163.

  Different tools (tmux, VSCode, IntelliJ) set different `TMPDIR` values, causing CLI and daemon to look
  for sockets in different locations.

  Socket directory priority:
  1. `AGENT_BROWSER_SOCKET_DIR` (explicit override)
  2. `$XDG_RUNTIME_DIR/agent-browser` (Linux standard)
  3. `~/.agent-browser/` (fallback, like Docker Desktop)

  ## Test plan
  - [x] `npm test` passes
  - [x] Manual test: different TMPDIR values use same socket
  - [x] Manual test: `--session` option works with different TMPDIR
  - [x] Manual test: `AGENT_BROWSER_SOCKET_DIR` override works

  ## Notes

  ### Breaking Change
  Socket files are now stored in `~/.agent-browser/` instead of `$TMPDIR/`.
  Old CLI + new daemon (or vice versa) won't work together.

  ### Backward Compatibility Considered
  I considered adding fallback to check old paths (similar to docker-modem's approach),
  but decided against it because:
  - v0.x allows breaking changes
  - CLI and daemon are installed together, reducing version mismatch risk
  - Keeps the code simple

  Let me know if you'd prefer backward compatibility added.

  🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -82,16 +82,20 @@ impl Connection {
 }
 
 /// Get the base directory for socket/pid files.
-/// Priority: AGENT_BROWSER_SOCKET_DIR > XDG_RUNTIME_DIR > ~/.agent-browser
+/// Priority: AGENT_BROWSER_SOCKET_DIR > XDG_RUNTIME_DIR > ~/.agent-browser > tmpdir
 pub fn get_socket_dir() -> PathBuf {
-    // 1. Explicit override
+    // 1. Explicit override (ignore empty string)
     if let Ok(dir) = env::var("AGENT_BROWSER_SOCKET_DIR") {
-        return PathBuf::from(dir);
+        if !dir.is_empty() {
+            return PathBuf::from(dir);
+        }
     }
 
-    // 2. XDG_RUNTIME_DIR (Linux standard)
+    // 2. XDG_RUNTIME_DIR (Linux standard, ignore empty string)
     if let Ok(runtime_dir) = env::var("XDG_RUNTIME_DIR") {
-        return PathBuf::from(runtime_dir).join("agent-browser");
+        if !runtime_dir.is_empty() {
+            return PathBuf::from(runtime_dir).join("agent-browser");
+        }
     }
 
     // 3. Home directory fallback (like Docker Desktop's ~/.docker/run/)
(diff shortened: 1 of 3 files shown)


ctate:
Great call, thanks for your contribution!

PR #106, opened 2026-01-15
feat: add Claude marketplace plugin

mking-keeper:
## Summary
- Purpose: Allow claude code users to install the skill as a claude plugin 
- Note that without this, each user of claude code may need to set up their own plugin if they want to install the agent browser skill using claude's plugin system 
- I focused on making this as minimal an update as possible 

## References (for inspo)
- [Anthropic's marketplace for claude plugins](https://github.com/anthropics/claude-plugins-official/) 
- [Upstash's marketplace for context7](https://github.com/upstash/context7)
- [Anthropic's claude marketplace docs](https://code.claude.com/docs/en/plugin-marketplaces)

## Usage
```
/plugin marketplace add vercel-labs/agent-browser
/plugin install agent-browser
```

## Testing
- I checked that I can install the marketplace (from my local clone) in `claude` (claude code cli)
image

- I checked that the skill is discovered and used by `claude` upon user request
image

PR #181, opened 2026-01-20
docs: add Claude Code plugin installation instructions

tomdale:
## Summary
- Add instructions for installing the agent-browser skill as a Claude Code plugin using the marketplace
- Mark plugin installation as the recommended approach
- Keep manual installation methods as alternatives

This follows up on #106 which added Claude marketplace plugin support.

## Test plan
- Verify the `/plugin marketplace add` and `/plugin install` commands work as documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -643,13 +643,22 @@ Core workflow:
 
 ### Claude Code Skill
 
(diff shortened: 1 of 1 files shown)


ctate:
Great followup - tysm

PR #188, opened 2026-01-21
fix: add .exe extension for Windows source binary path

mypengpengli:
The copy-native.js script was looking for 'agent-browser' but on Windows the compiled binary is 'agent-browser.exe', causing the copy to fail.


ctate:
Clean fix, thanks!

PR #202, opened 2026-01-22
doc: fix BrowserManager  import in  serverless example

yannliao:
updates BrowserManager  import issue in serverless example.


ctate:
Good catch! The fix is correct, but there's a second instance on line 557 that also needs updating:

```typescript
// Line 557 - Programmatic API section
import { BrowserManager } from 'agent-browser';
```

Could you update that one as well? Both examples should use 'agent-browser/dist/browser'.

PR #213, opened 2026-01-23
Fix keyboard events for Backspace/Delete keys

gdaybrice:
## Summary
- Add `keyCode` parameter to `injectKeyboardEvent` to pass `windowsVirtualKeyCode` and `nativeVirtualKeyCode` to CDP
- Update `InputKeyboardMessage` interface to include `keyCode`
- Pass `keyCode` from WebSocket message to the keyboard event handler

## Problem
Backspace and Delete keys were not working when using the streaming server. CDP's `Input.dispatchKeyEvent` requires `windowsVirtualKeyCode` for non-character keys like Backspace (keyCode 8) and Delete (keyCode 46) to function properly.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1314,6 +1314,7 @@ export class BrowserManager {
     key?: string;
     code?: string;
     text?: string;
+    keyCode?: number;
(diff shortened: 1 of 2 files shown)


Alecyrus:
？？？

Issue #214, opened 2026-01-23
Skill does not use agent-browser session by default, causing issues when parallel agents are testing


freddyhaddad:
If you install the claude skill of agent-browser and attempts to do many tests at the same time in parallel on your machine, they overlap with 1 browser. Shouldn't we update the skill to use agent-browser session each time it starts a new test or task that involves the browser to mitigate that?

freddyhaddad:
Solved and ready to merge #215

digitall-it:
The problem is more serious than parallel agents — it affects **any ongoing user session**, including work completely unrelated to testing.

**Root cause:** The agent-browser daemon is **persistent across all Claude conversations**. It does not restart between sessions. This means the default session (no `--session` flag) is a single shared browser that survives indefinitely.

**Real-world impact:**
A user had pfSense (a local network admin panel) open in the default session. An AI agent running browser tasks — without a named session — navigated away from pfSense, disrupting active work. The agent had no way to know the default session was already in use.

**The issue is not just parallelism** — even a single sequential agent can destroy a user's active browser session if it uses the default session, because the daemon started in a previous conversation (or manually by the user) persists.

**Proposed fix:** The official skill should **mandate named sessions** for every command. The current official SKILL.md uses the default session in all examples, which is the source of the problem.

Minimal safe pattern:
```bash
# Every command should carry --session <name>
agent-browser --session task-xyz open https://...
agent-browser --session task-xyz snapshot -i
agent-browser --session task-xyz close   # closes ONLY this session
```

Adding a warning or even a hard error when an agent tries to use the default session (if other sessions are active) could also help.

**Environment:** agent-browser v0.15.1, macOS, Claude Code

PR #215, opened 2026-01-23
Add session management guidance to prevent browser overlap

freddyhaddad:
## Summary

- Adds "Session management" section to `SKILL.md` instructing Claude to always use unique descriptive session names
- Requires `--session <name>` on every command to ensure commands go to the correct browser
- Adds context-aware cleanup rules (auto-close for autonomous tasks, ask for user-guided interactions)

**Problem:** When running parallel browser tasks, they all used the default session and shared one browser instance, causing interference.

**Solution:** The session isolation mechanism already exists (`--session` flag), but the skill didn't instruct Claude to use it. This PR adds that guidance.

## Test plan

- [ ] Install the updated skill
- [ ] Run multiple parallel browser tasks
- [ ] Verify each task uses a unique session name
- [ ] Verify no browser overlap between parallel tasks

### DIFF
--- a/skills/agent-browser/SKILL.md
+++ b/skills/agent-browser/SKILL.md
@@ -23,6 +23,38 @@ agent-browser close             # Close browser
 3. Interact using refs from the snapshot
 4. Re-snapshot after navigation or significant DOM changes
 
+## Session management
+
+**Always use a unique descriptive session name** for every browser task:
+
+```bash
+agent-browser --session login-test open https://example.com/login
+agent-browser --session checkout-flow open https://shop.example.com/cart
+```
+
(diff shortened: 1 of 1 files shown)


freddyhaddad:
Solves #214

Railly:
Thanks @freddyhaddad. #1589 shipped the session guidance proposed here and expanded it to require a named session before the first browser command.

The current skill explains that the default session is shared across agents and conversations, derives a stable task session ID, and covers parallel sessions plus shared Chrome tab pinning.

Closing this PR as superseded by #1589. Thank you for identifying the gap in #214 and proposing the original fix here.

PR #83, opened 2026-01-14
fix(cli): Output screenshot as base64 string when no path provided

jaydenfyi:
Closes #82

PR #247, opened 2026-01-24
chore(cli): save screenshots to tmp dir when no path provided

hewliyang:
## Summary

Reverts the base64 stdout behavior from #83 and instead saves screenshots to a temp directory when no path is provided.

## Problem

`agent-browser screenshot` outputting base64 to stdout is not useful for agentic use cases, it's a massive blob of unreadable text that causes context rot. 

Agents will figure out in the next step to give a filename anyway or pipe it into a file.

## Solution

When no path is provided, screenshots now save to `~/.agent-browser/tmp/screenshots/` with a generated filename and return the path:

```
$ agent-browser screenshot
✓ Screenshot saved to /Users/m1a1/.agent-browser/tmp/screenshots/screenshot-2026-01-24T09-40-15-656Z-q2jua0.png
```

## Changes

- `src/actions.ts`: Generate temp path when no path provided
- `src/daemon.ts`: Extract `getAppDir()` from `getSocketDir()` for cleaner semantics
- `cli/src/output.rs`: Remove base64 stdout handling, update help text
- Updated docs (README, quick-start, SKILL.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ agent-browser scroll <dir> [px]       # Scroll (up/down/left/right)
 agent-browser scrollintoview <sel>    # Scroll element into view (alias: scrollinto)
 agent-browser drag <src> <tgt>        # Drag and drop
 agent-browser upload <sel> <files>    # Upload files
-agent-browser screenshot [path]       # Take screenshot (--full for full page, saves to tmp if no path)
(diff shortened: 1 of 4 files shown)


ctate:
Good call. The base64 output was indeed problematic for agentic use cases, and saving to a temp directory with a returned path is a much better UX. Thank you!

PR #306, opened 2026-01-28
feat: concurrent tab targeting in CDP mode

possibilities:
## Problem

When multiple agents share a browser via `--cdp`, they fight over the global active tab. Tab addressing is index-based (shifts when tabs open/close), and each CLI invocation reconnects statelessly -- so there's no way for an agent to "hold onto" its tab across commands.

In practice this means two agents interleaving `tab 1` / `open <url>` commands will stomp on each other.

## Solution

This PR adds stable, per-tab targeting for CDP mode:

- **`--target <targetId>` flag** -- any command can specify which CDP target to operate on
- **targetId in tab commands** -- `tab new`, `tab list`, and `tab switch` all return the CDP `targetId`
- **`tab <targetId>`** -- switch to a tab by targetId instead of index
- **Page-specific CDP URLs** -- `--cdp "ws://host:port/devtools/page/<id>"` auto-extracts the target
- **Concurrency-safe** -- uses `AsyncLocalStorage` to bind `getPage()` to the target within each async context, so interleaved commands don't cross wires

Agent workflow:
```sh
agent-browser --cdp 9222 tab new --json
# → {"targetId":"A8C7634B...","index":1,"total":2}

agent-browser --cdp 9222 --target A8C7634B open https://example.com
agent-browser --cdp 9222 --target A8C7634B snapshot -i
```

## Changes

10 files, +485/-22. Touches both the Rust CLI and the TS daemon:

| Area | What |
|------|------|
| `src/browser.ts` | `AsyncLocalStorage` scoping, targetId cache, `getTargetId`/`resolveTarget`/`withTarget`/`switchToTarget` |
| `src/actions.ts` | Target-scoped `executeCommand` wrapper |
| `src/protocol.ts` | `targetId` on base schema, `tab_switch` accepts index or targetId |
| `src/types.ts` | `targetId` on `BaseCommand`, `TabInfo`, `TabNewData`, `TabSwitchData` |
| `cli/src/flags.rs` | `--target` flag |
| `cli/src/commands.rs` | `tab <targetId>` parsing |
| `cli/src/main.rs` | `extract_page_target` helper, page-specific URL support, targetId injection |
| `cli/src/output.rs` | targetId in tab list display, help text |

## Tests

- 11 new TS tests covering targetId lifecycle, caching, resolution, `withTarget` scoping, and cache cleanup on close
- 4 new TS protocol tests for targetId on base command and tab_switch variants
- 6 new Rust tests for `--target` flag parsing, `tab <targetId>` command parsing, and `extract_page_target`

Happy to iterate on any of this -- let me know if there's anything you'd like changed or if I can help get it across the line.

### DIFF
--- a/cli/src/flags.rs
+++ b/cli/src/flags.rs
@@ -50,7 +50,7 @@ pub fn parse_flags(args: &[String]) -> Flags {
         user_agent: env::var("AGENT_BROWSER_USER_AGENT").ok(),
         provider: env::var("AGENT_BROWSER_PROVIDER").ok(),
         ignore_https_errors: false,
-        target: env::var("AGENT_BROWSER_TARGET").ok(),
+        target: None,
     };
 
     let mut i = 0;
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -1672,7 +1672,6 @@ Environment:
   AGENT_BROWSER_SESSION          Session name (default: "default")
   AGENT_BROWSER_EXECUTABLE_PATH  Custom browser executable path
   AGENT_BROWSER_PROVIDER         Cloud browser provider
-  AGENT_BROWSER_TARGET           CDP target ID for tab targeting
   AGENT_BROWSER_STREAM_PORT      Enable WebSocket streaming on port (e.g., 9223)
 
 Examples:


possibilities:
For a bit of context on where this came from -- I'm working on [possibilities/browser](https://github.com/possibilities/browser), a fast headful browser with anti-detection running on Apple Silicon. The main use case is multiple AI agents each working in their own tab in parallel, sharing cookies/auth/localStorage in one browser instance. Without stable tab targeting the agents kept stepping on each other, so this was the missing piece to make that work reliably.

It's pretty awesome to see a bunch of claude codes using agent browser to treat a single native browser all at the same time. There's small app in possibilities/browser that let's me view a screencast of any tab.

possibilities:
I'm a little unsure of this part:

Page-specific CDP URLs -- --cdp "ws://host:port/devtools/page/<id>" auto-extracts the target

I couldn't get ws URLs to work with my setup and I like that I can use the port only to connect with cdp otherwise. Maybe I'll strip this part out?

possibilities:
Was fun idea but in practice isn't going to be feasible. Primarily because there is a shared ref namespace shared by all tabs. Maybe there's a way to address this at agent-browser level but wasn't a road I felt like going down. Thanks for this great tool!

Issue #318, opened 2026-01-29
Feature Request: Proper iframe context isolation for snapshot, eval, and screenshot


erinnovations:
## Description

The `frame` command exists and reports success, but subsequent commands (`snapshot`, `eval`, `screenshot`) continue to operate on the parent frame rather than the isolated iframe context.

## Current Behavior

```bash
# Navigate to page with iframes
agent-browser open https://example.com/dashboard

# Switch to iframe (reports success)
agent-browser frame "iframe#preview-panel"
# ✓ Done

# Snapshot still shows parent frame elements
agent-browser snapshot -i
# Shows: Settings, Profile, Logout (parent frame nav buttons)
# Expected: Elements inside the iframe

# Eval still returns parent context
agent-browser eval "window.location.href"
# Returns: "https://example.com/dashboard" (parent URL)
# Expected: "https://example.com/embed/preview/123" (iframe URL)

# Screenshot captures full page, not isolated iframe
agent-browser screenshot ./iframe.png
# Captures entire page including parent frame
```

## Expected Behavior

After `frame <selector>`, all subsequent commands should operate within the iframe context:

```bash
agent-browser frame "iframe#preview-panel"

# Snapshot should only show elements inside the iframe
agent-browser snapshot -i
# Shows only iframe's interactive elements

# Eval should execute in iframe context
agent-browser eval "window.location.href"
# Returns iframe's URL

# Screenshot should capture only the iframe's viewport
agent-browser screenshot ./iframe.png
# Captures only the iframe content, cropped to iframe bounds

# Return to parent
agent-browser frame main
```

## Use Case

Testing applications with split-view layouts (e.g., admin panel + preview iframe, IDE + live preview, email builder + preview) where you need to:

1. Inspect elements inside a specific iframe without noise from the parent
2. Take focused screenshots of iframe content for documentation/debugging
3. Execute JavaScript in the correct context
4. Get accurate element refs for clicking/interacting

## Current Workarounds

1. **Navigate directly to iframe URL** - Works but loses the parent page context:
   ```bash
   agent-browser eval "document.querySelector('iframe').src"
   agent-browser tab new <iframe-url>
   ```

2. **Coordinate-based clicking** - Clicks do pass through to iframe content, but fragile and requires manual coordinate calculation

## Suggested Implementation

Following Playwright's frame handling pattern:

```bash
# Option 1: Current command with fixed isolation
agent-browser frame "iframe#preview-panel"
[...]

PR #323, opened 2026-01-30
fix: Use frames as reference for all activity

nidhi-nair:
Fixes https://github.com/vercel-labs/agent-browser/issues/318

Make use of frames for all activity. The main frame is used by default, and agents are allowed to enter nested iframe context isolated from other iframes as well. Command for switching to main frame remains the same.


nidhi-nair:
@ctate would you be able to take a look at this PR to fix an issue with iframes?

PR #391, opened 2026-02-07
fix: prevent native file picker dialog during uploads

### LINKED ISSUE #192: feat: Add file download/save capability for files open in tabs
## Summary

Add the ability to download/save files (PDFs, images, etc.) that are already displayed in a browser tab to disk.

## Use Case

When a tab is viewing a file directly (e.g., `https://example.com/document.pdf` or an image URL), provide a way to save that content to the local filesystem.

### Detecting File Type in Tab

1. **URL pattern matching** - Check for `.pdf`, `.png`, `.jpg`, etc. extensions
2. **MIME type via CDP** - Use `Network.responseReceived` event which includes `response.mimeType`
3. **Chrome PDF viewer detection** - Check for `embed[type="application/x-google-chrome-pdf"]` element

### Saving the Content

**Option A: Re-fetch via `page.evaluate()` (most reliable for already-displayed files)**
```javascript
const base64Data = await page.evaluate(async () => {
  const response = await fetch(window.location.href);
  const blob = await response.blob();
  return new Promise((resolve) => {
    const reader = new FileReader();
    reader.onloadend = () => resolve(reader.result.split(',')[1]);
    reader.readAsDataURL(blob);
  });
});
fs.writeFileSync('file.pdf', Buffer.from(base64Data, 'base64'));
```

**Option B: `Network.getResponseBody` via CDP**
- Must be called during/immediately after `loadingFinished` event before response is garbage collected
- Good for intercepting during navigation, less useful for already-loaded tabs

**Option C: `Page.getResourceContent` for cached resources**
```javascript
const { content, base64Encoded } = await client.send('Page.getResourceContent', {
  frameId,
  url: tabUrl
});
```

### Proposed API

Could be a new MCP tool like:
```
mcp__claude-in-chrome__save_file
  - tabId: number
  - outputPath: string (destination on disk)
  - filename?: string (optional, derive from URL if not provided)
```

Or extend existing tools to support this.

## Considerations

- Large files may need streaming (`Fetch.takeResponseBodyAsStream`)
- Files behind auth may require cookie/session preservation during re-fetch
- Should detect and handle different content types appropriately

NikhilGaddam:
## Fix file upload with relative paths

The upload command fails when relative file paths are passed because Playwright's `setInputFiles` expects absolute paths. This resolves relative paths to absolute before passing them to `setInputFiles`.

The previous approach of using CDP `DOM.setFileInputFiles` to avoid the native file dialog is unnecessary — Playwright's `setInputFiles` already bypasses the native dialog entirely. The CDP approach actually introduced a regression: cached CDP sessions become stale after page navigations, causing `Target page, context or browser has been closed` errors.

This keeps the fix minimal: just add path resolution while staying with Playwright's built-in API.

Fixes #192

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -826,14 +826,23 @@ async function handleUncheck(command: UncheckCommand, browser: BrowserManager): 
 }
 
 async function handleUpload(command: UploadCommand, browser: BrowserManager): Promise<Response> {
-  const locator = browser.getLocator(command.selector);
   const files = Array.isArray(command.files) ? command.files : [command.files];
+
   try {
-    await locator.setInputFiles(files);
+    const absoluteFiles = files.map((file) => {
+      if (path.isAbsolute(file)) {
+        return file;
+      }
+      return path.resolve(process.cwd(), file);
+    });
+
+    const locator = browser.getLocator(command.selector);
+    await locator.setInputFiles(absoluteFiles);
+
+    return successResponse(command.id, { uploaded: absoluteFiles });
   } catch (error) {
     throw toAIFriendlyError(error, command.selector);
   }
-  return successResponse(command.id, { uploaded: files });
 }
 
 async function handleDoubleClick(


NikhilGaddam:
Fixed the CSS selector issue\! Now using `DOM.describeNode` to get the node ID directly from the element's remote object when dealing with refs. This completely avoids building CSS selectors from element attributes, so no more issues with special characters in IDs/names or the problematic `input[type="file"]` fallback that could match the wrong element.

Blakeinstein:
I would love this functionality. Currently my agent is just stuck once the native file dialog opens with no way to get rid of it.

PR #399, opened 2026-02-09
Fix clippy lints

mlafeldt:
Fix all linter warnings reported by `cargo clippy --all-targets --all-features`, using Rust 1.93.0.


ctate:
Thanks for cleaning this up!

PR #405, opened 2026-02-10
Action timeouts and improved Snapshot behavior

glade-at-gigwell:
* Fix crashes with undefined timeouts in Playwright which manifest in Rust IPC Errors.
* Add ARIA tag support into the snapshot json
* Remove `snapshot` raw text from the json object unless the --raw parameter is specified to improve payload sizes.


glade-at-gigwell:
Sorry, wasn't intended to submit this to the vercel repo yet.

PR #410, opened 2026-02-10
fix: preserve chrome-extension:// and chrome:// URL schemes in CLI

### LINKED ISSUE #409: open command mangles chrome-extension:// URLs by auto-upgrading to HTTPS
When using `agent-browser open` with a `chrome-extension://` URL, the scheme is silently rewritten to `https://chrome-extension//...` (colon stripped, scheme upgraded), causing `ERR_NAME_NOT_RESOLVED`.

This prevents navigating to extension pages (popup, side panel, options) which is a core requirement for Chrome extension testing.

  ## Reproduction

  ```bash
  # Load an extension and try to navigate to its page
  agent-browser \
    --extension /path/to/unpacked/extension \
    --session test \
    --headed \
    open "chrome-extension://abcdefghijklmnop/popup.html"
```

  Expected: Browser navigates to `chrome-extension://abcdefghijklmnop/popup.html`

  Actual:
  ✗ page.goto: `net::ERR_NAME_NOT_RESOLVED` at `https://chrome-extension//abcdefghijklmnop/popup.html`

  The `://` is stripped and the URL is upgraded to https://, producing an invalid URL.

  ## Workarounds attempted
  
  | Method | Result |
  |--------|--------|
  | `agent-browser open chrome-extension://...` | Rewritten to `https://chrome-extension//...` |
  | `agent-browser tab new chrome-extension://...` | `ERR_BLOCKED_BY_CLIENT` |
  | `agent-browser eval "location.href = 'chrome-extension://...'"` | `ERR_BLOCKED_BY_CLIENT` |
  | `agent-browser eval "window.open('chrome-extension://...')"` | Tab created but shows `chrome-error://chromewebdata/` |

  None of the agent-browser commands can navigate to `chrome-extension://` URLs.

  ## Working Playwright approach

  Using Playwright directly (which agent-browser wraps) works perfectly:
```
  const { chromium } = require('playwright');

  const context = await chromium.launchPersistentContext('/tmp/test', {
    args: [
      `--disable-extensions-except=${extPath}`,
      `--load-extension=${extPath}`,
    ],
  });

  let sw = context.serviceWorkers()[0];
  if (!sw) sw = await context.waitForEvent('serviceworker');
  const extensionId = sw.url().split('/')[2];

  const page = context.pages()[0];
  await page.goto(`chrome-extension://${extensionId}/popup.html`); // Works!
```

  Reference: https://playwright.dev/docs/chrome-extensions

  This confirms the issue is in agent-browser's URL processing, not a Playwright or Chromium limitation.

##  Suggested fix

  Skip the HTTPS auto-upgrade for URLs with the chrome-extension:// scheme. These are legitimate navigation targets when `--extension` is used and should be passed through to Playwright's `page.goto()` unmodified.

  Similarly, `chrome://` URLs (e.g., `chrome://extensions`) are also affected by the upgrade logic and could be preserved.

rsiddle:
The CLI's URL normalization was auto-prepending https:// to any URL whose scheme wasn't in the allowlist (http, https, about, data, file). This caused chrome-extension:// URLs to become
https://chrome-extension//... which fails with ERR_NAME_NOT_RESOLVED, preventing navigation to extension pages (popup, side panel, options).

Add chrome-extension:// and chrome:// to the open command's scheme allowlist, and update the record start/restart commands to preserve any URL that already contains :// instead of only checking for http.

Fixes #409

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -91,6 +91,8 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
                 || url_lower.starts_with("about:")
                 || url_lower.starts_with("data:")
                 || url_lower.starts_with("file:")
+                || url_lower.starts_with("chrome-extension://")
+                || url_lower.starts_with("chrome://")
             {
                 url.to_string()
             } else {
@@ -759,8 +761,8 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
                     let url = rest.get(2);
                     let mut cmd = json!({ "id": id, "action": "recording_start", "path": path });
                     if let Some(u) = url {
-                        // Add https:// prefix if needed
-                        let url_str = if u.starts_with("http") {
+                        // Add https:// prefix if needed (preserve special schemes)
+                        let url_str = if u.starts_with("http") || u.contains("://") {
                             u.to_string()
                         } else {
                             format!("https://{}", u)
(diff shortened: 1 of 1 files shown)


ctate:
Thanks @rsiddle!

PR #425, opened 2026-02-12
feat: add selector command to extract stable CSS selectors from refs

davekiss:
Adds a new `selector` command that extracts stable CSS selectors from element refs (@e1, @e2, etc.) for use in deterministic scripts across sessions.

### Usage:
  `agent-browser selector @e1`          # Get best stable selector
  `agent-browser selector @e1 --all`  # Get all candidate selectors

### Selector priority (by stability):
  1. `data-testid`
  2. `id` (if not auto-generated)
  3. `aria-label`
  4. Test attributes (`data-cy`, `data-qa`, `data-test`, `data-e2e`)
  5. Unique class combinations (filters utility classes)
  6. Structural path with `nth-of-type` (last resort)

### DIFF
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -1485,6 +1485,42 @@ Examples:
 "##
         }
 
+        // === Selector ===
+        "selector" => {
+            r##"
+agent-browser selector - Extract stable CSS selectors from element refs
+
+Usage: agent-browser selector <@ref> [--all]
+
+Extracts stable CSS selectors from element refs (@e1, @e2, etc.) for use
+in deterministic scripts across sessions. This is useful when you want to
+convert a snapshot ref into a reusable selector for automation scripts.
+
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the contribution! The selector extraction approach and stability priority system are well thought out. A few things to address before merging:

**Bug: nonexistent ref produces an internal error**

`selector @e999` returns a raw Playwright CSS parsing error instead of the intended "Element not found" message. `getLocator` falls through to treating the unknown ref as a CSS selector, and `locator.count()` fails before the `count === 0` check. Wrap it in a try/catch or check the ref map before calling `getLocator`.

**Missing docs site pages**

Per `AGENTS.md`, `docs/src/app/` must be updated for user-facing features. The `selector` command needs to be added to `commands/page.mdx` and `selectors/page.mdx`.

PR #502, opened 2026-02-18
feat: Add browserless.io as a browser provider

joelgriffith:
Apologies on the markdown changes, it's a bit better formatted IMO, but happy to change that.

I did also write a little boolean-env parsing utility since we've seen a lot of folks prefer the `0` and `1` format as well as `true` and `false`.

Happy to make any changes needed.

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1335,6 +1335,7 @@ export class BrowserManager {
       }
 
       this.browser = browser;
+      this.browserlessStopUrl = session.stop;
       context.setDefaultTimeout(getDefaultTimeout());
       this.contexts.push(context);
       this.setupContextTracking(context);


joelgriffith:
Been about two weeks for this PR, anyone able to approve the run at least on the Vercel Check? Apologies on the direct ping @ctate

ctate:
Thanks for the PR @joelgriffith - and sorry about the delay!

I've reviewed the code and tested it locally. Builds clean, all unit tests pass, E2E smoke test of the core workflow works.

I found a few issues that need addressing before merge:

1. Missing ensureDomainFilter(context) and sanitizeExistingPages([page]) calls. Other providers call both of these after setting up the context. The connectToBrowserless method skips them, which could bypass domain filtering. The post-connect setup should follow the same order as the other providers.

2. Hardcoded context.setDefaultTimeout(60000). Should use getDefaultTimeout() (defaults to 25000ms) like the other providers. This respects the AGENT_BROWSER_DEFAULT_TIMEOUT env var.

3. browserlessApiToken is stored but never used. Unlike other providers, the browserless cleanup only needs the stop URL. You can remove this field or, if you plan to use it later, add a comment.

4. parseBooleanEnvVar doesn't use `this`. Consider making it a standalone utility function or static method. The existing Kernel provider handles boolean env vars inline, so this could be a nice refactor to share across providers, but scoping it as a BrowserManager instance method isn't quite right.

5. Missing Documentation: skills/agent-browser/SKILL.md, docs/src/app/commands/page.mdx, docs/src/app/cdp-mode/page.mdx, skills/agent-browser/references/commands.md

Happy to help with any of these. Looking forward to the updated version!

joelgriffith:
Thanks! I added the feedback you mentioned, I didn't find a relevant entry in `skills/agent-browser/SKILL.md` or in `skills/agent-browser/references/commands.md`. Did I miss something there?

Otherwise feels like it's in line with the other providers, thanks again @ctate

joelgriffith:
@ctate sorry again for the ping. Feedback addressed!

ctate:
Thanks for the update! One bug to fix: `this.browserlessStopUrl` is never assigned in the success path of `connectToBrowserless()` (around line 1306). The other providers all store their session identifiers, but the Browserless path is missing the equivalent. This means `close()` will never call `closeBrowserlessSession()`, so sessions leak until TTL.

Add after `this.browser = browser;`:
```ts
this.browserlessStopUrl = session.stop;
```

joelgriffith:
> Thanks for the update! One bug to fix: `this.browserlessStopUrl` is never assigned in the success path of `connectToBrowserless()` (around line 1306). The other providers all store their session identifiers, but the Browserless path is missing the equivalent. This means `close()` will never call `closeBrowserlessSession()`, so sessions leak until TTL.
> 
> Add after `this.browser = browser;`:
> 
> ```ts
> this.browserlessStopUrl = session.stop;
> ```

Great catch, fixed!

ctate:
Awesome - thanks @joelgriffith!

PR #508, opened 2026-02-19
Add page range support to PDF command

the-vampiire:
## Summary

Adds `--pages` flag to the `pdf` command to specify which pages to include in the generated PDF, using Playwright's native `pageRanges` option.

## Changes

- **CLI Parser** (`cli/src/commands.rs`): Added `-p/--pages` flag parser
- **Protocol** (`src/protocol.ts`): Added `pageRanges` field to `pdfSchema` validation
- **Types** (`src/types.ts`): Added optional `pageRanges` field to `PdfCommand` interface
- **Actions** (`src/actions.ts`): Pass `pageRanges` to Playwright's `page.pdf()` when specified

## Usage

```bash
# Single page
agent-browser pdf output.pdf --pages "1"

# Page range
agent-browser pdf output.pdf --pages "1-3"

# Multiple ranges
agent-browser pdf output.pdf --pages "1-5, 8, 11-13"
```

## Testing

Tested with Chrome via CDP:
- Single page (`--pages "1"`): ✅ Generates 1-page PDF
- Page range (`--pages "1-3"`): ✅ Generates 3-page PDF
- Specific page (`--pages "5"`): ✅ Generates 1-page PDF with page 5

## Implementation Notes

The `pageRanges` parameter uses Playwright's standard format (e.g., "1-5, 8, 11-13") and is passed directly to Playwright's `page.pdf()` method.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -404,11 +404,35 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
             )
         }
         "pdf" => {
-            let path = rest.first().ok_or_else(|| ParseError::MissingArguments {
+            let mut cmd = json!({ "id": id, "action": "pdf" });
+            let obj = cmd.as_object_mut().unwrap();
+            let mut path: Option<&str> = None;
+            let mut i = 0;
+
+            while i < rest.len() {
+                match rest[i] {
+                    "-p" | "--pages" => {
+                        if let Some(p) = rest.get(i + 1) {
+                            obj.insert("pageRanges".to_string(), json!(p));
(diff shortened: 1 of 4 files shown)


the-vampiire:
Closing this PR after testing revealed that Playwright's page.pdf() and CDP's Page.printToPDF don't replicate Chrome's native print dialog behavior. They render the viewport as-is rather than applying Chrome's full print layout engine, which means pageRanges only works for already-paginated content. For use cases like Google Forms responses that expand to 15 pages in Chrome's print dialog but render as 1 page via CDP, this feature doesn't provide the expected value. Thanks for the learning experience\!

Issue #526, opened 2026-02-23
"Cannot load state while browser is running. Close browser first, then relaunch with loaded state."


ro0sterjam:
`agent-browser state load` does not seem to work.

Whenever I try to call it, it tells me that the browser needs to be closed first before launching with the loaded state. But as you can see in the output below it is already closed.

```
➜  workspace git:(main) ✗ agent-browser session list
No active sessions
➜  workspace git:(main) ✗ agent-browser state load ~/workspace/tmp/auth-state.json
✗ Cannot load state while browser is running. Close browser first, then relaunch with loaded state.
➜  workspace git:(main) ✗ agent-browser session list
Active sessions:
→ default
➜  workspace git:(main) ✗ agent-browser close
✓ Browser closed
➜  workspace git:(main) ✗ agent-browser state load ~/workspace/tmp/auth-state.json
✗ Cannot load state while browser is running. Close browser first, then relaunch with loaded state.
```

```
➜  workspace git:(main) ✗ agent-browser --version
agent-browser 0.13.0
```

PR #527, opened 2026-02-23
fix state load failing when no browser is running

ctate:
`state load` always fails with "Cannot load state while browser is running" even when no browser is running, making the command completely unusable (#526).

The daemon's auto-launch logic starts a browser before `state_load` gets to handle the command. This adds `state_load` to the exclusion list alongside `launch` and `close`, so `handleStateLoad` can perform its own launch with the state file.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,6 +2,10 @@
 
 Instructions for AI coding agents working with this codebase.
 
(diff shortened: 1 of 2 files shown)


ro0sterjam:
Wow that was fast! Thank you.

Issue #322, opened 2026-01-30
macOS: EAGAIN (os error 35) crashes CLI instead of retrying


mikewiendels:
## Description

On macOS, the CLI crashes with `Failed to read: Resource temporarily unavailable (os error 35)` when the Unix socket read encounters an EAGAIN condition. This happens during normal operation, particularly with rapid sequential commands.

## Environment

- **OS:** macOS (Darwin 25.2.0, Apple Silicon)
- **agent-browser version:** 0.7.6
- **Node.js:** v22.x

## Steps to Reproduce

1. Run a sequence of browser automation commands in quick succession
2. Especially after interactions that trigger React re-renders or async operations
3. The daemon occasionally doesn't respond immediately, causing EAGAIN

Example test sequence that triggers it:
```bash
agent-browser open https://example.com
agent-browser click [data-ai-id="some-button"]
agent-browser eval document.querySelector('[data-ai-id="result"]')?.textContent
# ^ This sometimes fails with error 35
```

## Error Output

```
✗ Failed to read: Resource temporarily unavailable (os error 35)
```

## Root Cause

In `cli/src/connection.rs`, the `send_command` function doesn't handle EAGAIN:

```rust
reader
    .read_line(&mut response_line)
    .map_err(|e| format!("Failed to read: {}", e))?;
```

On macOS, error 35 is EAGAIN/EWOULDBLOCK. This occurs when:
- The socket read would block because data isn't available yet
- The daemon is still processing the previous command
- There's momentary socket congestion

## Suggested Fix

Add retry logic for EAGAIN errors:

```rust
use std::io::ErrorKind;

// In send_command:
let mut retries = 0;
const MAX_RETRIES: u32 = 10;

loop {
    match reader.read_line(&mut response_line) {
        Ok(_) => break,
        Err(e) if e.raw_os_error() == Some(35) || e.kind() == ErrorKind::WouldBlock => {
            retries += 1;
            if retries >= MAX_RETRIES {
                return Err(format!("Failed to read after {} retries: {}", MAX_RETRIES, e));
            }
            thread::sleep(Duration::from_millis(50));
            continue;
        }
        Err(e) => return Err(format!("Failed to read: {}", e)),
    }
}
```

## Workaround

For now, users can:
1. Add `agent-browser wait 200` between rapid commands
2. Restart the daemon if it gets into a bad state: `pkill -f "node.*agent-browser"`

## Related

[...]

PR #529, opened 2026-02-23
fix(daemon): add backpressure control and command serialization to prevent IPC EAGAIN

shohu:
## Summary

Fixes the daemon-side root causes of EAGAIN (os error 35 on macOS / error 11 on Linux) that crash the Rust CLI during IPC reads.

This complements #329 (CLI-side retry logic) by preventing the conditions that trigger EAGAIN in the first place.

### Changes

**`src/browser.ts`** — Configurable Playwright timeout via environment variable
- Add `getDefaultTimeout()` helper that reads `AGENT_BROWSER_DEFAULT_TIMEOUT` env var
- Replace all hardcoded `setDefaultTimeout(60000)` calls (5 locations) with `setDefaultTimeout(getDefaultTimeout())`
- CDP and recording contexts (10s timeout) are **not** affected
- Default remains 60s when env var is unset — fully backward compatible

**`src/daemon.ts`** — Backpressure-aware writes + command serialization
- Add `safeWrite()` helper that waits for `drain` event when `socket.write()` returns `false`, with proper cleanup on `close`/`error`
- Serialize command execution per socket via a queue to prevent concurrent `socket.write()` calls that cause kernel buffer contention
- Add `.catch()` guards for unhandled rejection safety

### Root Cause Analysis

Two daemon-side issues combine to cause EAGAIN:

1. **Playwright timeout > CLI IPC timeout**: `setDefaultTimeout(60000)` means Playwright can block for up to 60s, but the Rust CLI times out earlier. The daemon never responds, and the CLI's `read_line()` hits EAGAIN.

2. **Uncontrolled `socket.write()` concurrency**: Multiple async command handlers can call `socket.write()` in parallel. When payloads are large (e.g., `snapshot` responses), the kernel buffer fills up and subsequent writes/reads fail with EAGAIN.

### Testing

- All 363 existing tests pass (9 skipped)
- Prettier formatting verified
- TypeScript compilation clean
- Manually tested with a heavy React Native/Expo app (~1000+ DOM nodes): 10 consecutive `snapshot` commands complete without os error 35

### Usage

```bash
# Set a shorter timeout (e.g., 30s) to prevent daemon hangs
export AGENT_BROWSER_DEFAULT_TIMEOUT=30000
```

Refs #322

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -28,6 +28,22 @@ import {
   ENCRYPTION_KEY_ENV,
 } from './state-utils.js';
 
+/**
+ * Returns the default Playwright timeout for standard operations.
+ * Can be overridden via the AGENT_BROWSER_DEFAULT_TIMEOUT environment variable.
+ * CDP and recording contexts use a shorter fixed timeout (10s) and are not affected.
+ */
+function getDefaultTimeout(): number {
+  const envValue = process.env.AGENT_BROWSER_DEFAULT_TIMEOUT;
+  if (envValue) {
+    const parsed = parseInt(envValue, 10);
+    if (!isNaN(parsed) && parsed > 0) {
+      return parsed;
+    }
+  }
+  return 60000;
+}
+
 // Screencast frame data from CDP
 export interface ScreencastFrame {
   data: string; // base64 encoded image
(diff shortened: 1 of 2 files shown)


ctate:
This is so good - thanks @shohu

PR #531, opened 2026-02-23
cli: only warn about --annotate when explicitly passed via CLI

### LINKED ISSUE #530: Config-sourced --annotate warns on every non-screenshot command
When `annotate` is set in the user-level config (`~/.agent-browser/config.json`), every non-screenshot command emits a warning:

```
$ agent-browser close
⚠ --annotate only applies to the screenshot command
✓ Browser closed
```

Config:
```json
{
  "annotate": true
}
```

The purpose of a user-level config is to set persistent defaults. Options that don't apply to the current command should be silently ignored when they come from config — the warning only makes sense when the user explicitly passes `--annotate` on the CLI.

The codebase already handles this for other flags via `cli_*` tracking fields on the `Flags` struct (e.g. `cli_executable_path`, `cli_allow_file_access`), but `annotate` is missing one.

lukasmalkmus:
The `--annotate` warning fires on every non-screenshot command when `annotate` is set in the user-level config file (`~/.agent-browser/config.json`). This is noisy for users who set it as a persistent default — the whole point of a config file.

Adds `cli_annotate` tracking to the `Flags` struct, matching the existing `cli_*` pattern used for `executable_path`, `extensions`, `profile`, etc. The warning now only fires when `--annotate` is explicitly passed as a CLI flag.

Fixes #530

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -83,7 +83,7 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
     let rest: Vec<&str> = args[1..].iter().map(|s| s.as_str()).collect();
     let id = gen_id();
 
-    if flags.annotate && cmd != "screenshot" {
+    if flags.cli_annotate && cmd != "screenshot" {
         eprintln!(
             "{} --annotate only applies to the screenshot command",
             color::warning_indicator()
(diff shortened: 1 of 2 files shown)


ctate:
LGTM - thanks @lukasmalkmus

PR #563, opened 2026-02-28
fix: correct misleading output for cookies clear and tab close

### LINKED ISSUE #556: [Bug] Misleading output messages: `cookies clear`, `tab close`, and duplicated error output
## Summary

Three output messages are incorrect or duplicated, causing confusion for both human users and AI agents that parse CLI output to determine success/failure:

1. `cookies clear` prints **"Request log cleared"** — wrong message
2. `tab close` prints **"Browser closed"** — suggests the entire browser was closed when only a tab was
3. **Errors are printed twice** — daemon-level errors appear duplicated in the output

## Environment

- **agent-browser version:** 0.15.1
- **OS:** macOS darwin 25.3.0
- **Install method:** `npm install -g agent-browser`

---

## Bug 1 — `cookies clear` prints wrong message

```bash
agent-browser --session s1 open https://example.com
agent-browser --session s1 cookies clear
```

**Actual output:**
```
✓ Request log cleared
```

**Expected output:**
```
✓ Cookies cleared
```

The cookies ARE cleared (the operation succeeds), but the message belongs to a different command (`network requests --clear`). An AI agent reading this output may conclude cookies were not cleared and retry or report an error.

---

## Bug 2 — `tab close` prints "Browser closed"

```bash
agent-browser --session s1 open https://example.com
agent-browser --session s1 tab new https://example.org
agent-browser --session s1 tab close 1
```

**Actual output:**
```
✓ Browser closed
```

**Expected output:**
```
✓ Tab closed
```

The browser is still running and the session is still active. An AI agent seeing "Browser closed" may incorrectly assume the session is dead and attempt to restart it.

---

## Bug 3 — Errors printed twice

When a command fails (e.g. strict mode violation, unknown selector), the error message is emitted twice in the output:

```bash
agent-browser --session s1 get html "button"
# If multiple buttons exist:
```

**Actual output:**
```
✗ strict mode violation: locator('button') resolved to 5 elements
✗ strict mode violation: locator('button') resolved to 5 elements
```

**Expected output:**
```
✗ strict mode violation: locator('button') resolved to 5 elements
```

The error appears to be printed once by the daemon and once echoed by the CLI layer. The operation fails only once (correct), but the duplicate output confuses agents that count error lines or parse stderr vs stdout.

---

## Impact

These cosmetic bugs are particularly harmful in AI agent workflows where the CLI output is the only feedback mechanism. Wrong messages cause agents to misinterpret the result of an operation, potentially triggering unnecessary retries, incorrect state tracking, or false failure reports.

giulio-leone:
## Summary

Fixes two misleading CLI output messages reported in #556 that confuse both human users and AI agents parsing CLI output.

### Bug 1 — `cookies clear` prints wrong message

**Before:** `✓ Request log cleared`
**After:** `✓ Cookies cleared`

**Root cause:** Both `cookies_clear` and `requests --clear` return `{ cleared: true }`. The output handler matched this generic shape without checking which action produced it, always printing the `requests` message.

**Fix:** Use the `action` parameter (already passed to `print_response_with_opts`) to distinguish `cookies_clear` from other clear operations.

### Bug 2 — `tab close` prints wrong message

**Before:** `✓ Browser closed`
**After:** `✓ Tab closed`

**Root cause:** Both `close` (browser) and `tab_close` return a response with a `closed` field. The output handler matched the generic shape without checking the action.

**Fix:** Use the `action` parameter to distinguish `tab_close` from `close`.

### Bug 3 — Duplicate error output

Investigated the full error path (daemon `executeCommand` → `errorResponse` → socket write → CLI `print_response_with_opts`) and confirmed only one error message is emitted per failed command. The daemon's stderr is piped to `Stdio::null()` when auto-started, `send_command_once` reads exactly one response line, and the `processQueue` serialization prevents duplicate processing. Could not reproduce from code analysis — may be environment-specific or version-dependent.

### Testing

- 256/256 Rust CLI tests pass (2 consecutive clean runs)
- 376/376 applicable TypeScript tests pass (3 test files fail due to pre-existing Playwright environment issues, unrelated to this change)

Closes #556

### DIFF
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -318,10 +318,14 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
             }
             return;
         }
-        // Cleared requests
+        // Cleared (cookies or request log)
         if let Some(cleared) = data.get("cleared").and_then(|v| v.as_bool()) {
             if cleared {
-                println!("{} Request log cleared", color::success_indicator());
+                let label = match action {
+                    Some("cookies_clear") => "Cookies cleared",
+                    _ => "Request log cleared",
+                };
+                println!("{} {}", color::success_indicator(), label);
                 return;
             }
         }
@@ -382,9 +386,13 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
             }
             return;
         }
-        // Closed
+        // Closed (browser or tab)
         if data.get("closed").is_some() {
-            println!("{} Browser closed", color::success_indicator());
+            let label = match action {
+                Some("tab_close") => "Tab closed",
+                _ => "Browser closed",
+            };
+            println!("{} {}", color::success_indicator(), label);
             return;
         }
         // Recording start (has "started" field)


ctate:
@giulio-leone Thanks for fixing these!

PR #564, opened 2026-02-28
fix: treat EPERM from kill(pid, 0) as process exists

neilmix:
## Summary

- `kill(pid, 0)` is used to check if the daemon process is alive, but all three call sites treat any failure as "process not running"
- Per POSIX, `EPERM` means the process **exists** but the caller lacks permission to signal it, while `ESRCH` means it does not exist
- Inside a macOS sandbox with `(allow signal (target self))`, `kill` returns `EPERM` for all other PIDs, causing the CLI to incorrectly conclude the daemon is dead
- This leads to deletion of the running daemon's socket/PID files and spawning of a duplicate daemon

### Changes

**`cli/src/connection.rs` — `is_daemon_running`:** Check `errno` after a failed `kill(pid, 0)`; only return `false` when `errno == ESRCH`.

**`cli/src/main.rs` — `run_session` (session list):** Same fix for the inline liveness check.

**`src/daemon.ts` — `isDaemonRunning`:** Check `err.code === 'EPERM'` before cleaning up stale files.

## Test plan

- [x] `cargo test` — all 256 existing tests pass
- [ ] Run CLI inside a macOS sandbox that restricts signals to `(target self)` and verify the daemon is correctly detected as running
- [ ] Verify that stale PID files (where the process is genuinely gone) are still cleaned up correctly

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -159,7 +159,13 @@ fn is_daemon_running(session: &str) -> bool {
     if let Ok(pid_str) = fs::read_to_string(&pid_path) {
         if let Ok(pid) = pid_str.trim().parse::<i32>() {
             unsafe {
-                return libc::kill(pid, 0) == 0;
+                if libc::kill(pid, 0) == 0 {
+                    return true;
+                }
+                // EPERM means the process exists but we lack permission to
+                // signal it (e.g. inside a macOS sandbox). Only ESRCH means
+                // the process is genuinely gone.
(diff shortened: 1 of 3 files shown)


neilmix:
Full disclosure: this PR was AI generated. I discovered this issue while trying to use agent-browser for web testing within my sandbox utility, sandbash.  https://github.com/neilmix/sandbash

ctate:
@neilmix Thank you for the fix!

PR #572, opened 2026-02-28
fix(windows): improve daemon error messages and add debug logging

kiss-kedaya:
## Problem

On Windows, when the daemon fails to start, the error message incorrectly shows a Unix socket path instead of the TCP port number that Windows actually uses.

## Solution

- Fix error message to show port number on Windows instead of socket path
- Add debug logging for daemon startup (command, session, port)
- Helps diagnose daemon startup failures on Windows

## Changes

- Modified cli/src/connection.rs to use platform-specific error messages
- Added debug logging with println! for Windows daemon startup

## Testing

Tested on Windows 11 with agent-browser 0.15.1:
- Error messages now correctly show port numbers
- Debug logs help identify startup issues
- Daemon successfully starts and connects

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -400,6 +400,10 @@ pub fn ensure_daemon(
 
         // On Windows, call node directly. Command::new handles PATH resolution (node.exe or node.cmd)
         // and automatically quotes arguments containing spaces.
+        eprintln!("[DEBUG] Starting daemon: node {:?}", daemon_path);
+        eprintln!("[DEBUG] Session: {}", session);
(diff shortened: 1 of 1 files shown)


trikiman:
```
PS E:\Projects\saleapp> agent-browser open example.com                                
✗ Daemon failed to start (socket: C:\Users\rust-\.agent-browser\default.sock)


```

it is this problem? cus i idk how to use this agent its always failed

trikiman:
for fix this i need to run `node "$env:APPDATA\npm\node_modules\agent-browser\dist\daemon.js"`
 in diff terminal

PR #573, opened 2026-03-01
fix: resolve unnamed element refs matching multiple elements

### LINKED ISSUE #500: [Bug] Selector "@e18" matched 8 elements. Run 'snapshot' to get updated refs, or use a more specific CSS selector.
Image

```
    - text: First name
    - textbox "First name" [ref=e13]:
      - /placeholder: Enter first name
    - paragraph: Please enter a first name.
    - text: Last name
    - textbox "Last name" [ref=e14]:
      - /placeholder: Enter last name
    - paragraph: Please enter a last name.
    - text: Date of birth (MM/DD/YYYY) *
    - textbox "Date of birth (MM/DD/YYYY) *" [ref=e15]:
      - /placeholder: MM/DD/YYYY
    - button "Choose date" [ref=e16]
    - paragraph: Please enter a date of birth.
    - text: State *
    - button "Help icon Button" [ref=e17]
    - button [ref=e18]
    - paragraph: Please select a state.
    - button "Search" [ref=e19]
    - link "Clear search" [ref=e20]:
```

```
agent-browser click @e18          
✗ Selector "@e18" matched 8 elements. Run 'snapshot' to get updated refs, or use a more specific CSS selector.
```

When there are multiple buttons and I want to click the button without name.

jin-2-kakaoent:
## Summary

Fixes #500

When a page has one unnamed button among several named buttons (e.g., `button [ref=e18]` alongside `button "Search"`, `button "Choose date"`, etc.), clicking `@e18` fails with:

```
✗ Selector "@e18" matched 8 elements. Run 'snapshot' to get updated refs, or use a more specific CSS selector.
```

## Root Cause

When the ARIA snapshot is parsed, each line is matched against a regex:

```typescript
const match = line.match(/^(\s*-\s*)(\w+)(?:\s+"([^"]*)")?(.*)$/);
const [, prefix, role, name, suffix] = match;
```

For `- button "Search"`, `name = "Search"`. For `- button` (unnamed), the capture group doesn't match, so `name = undefined`.

This `undefined` flows through three steps that together cause the bug:

**1. `buildSelector('button', undefined)`** — the old code checked `if (name)`, which is falsy for `undefined`, so it returned `getByRole('button')` with **no name filter**.

**2. `removeNthFromNonDuplicates()`** — this function only preserves `nth` for refs that share the same `role:name` key. The key `"button:"` has count=1 (only one unnamed button), so `nth` is deleted.

**3. `getLocatorFromRef()`** — the old code checked `if (refData.name)`, which is falsy for `undefined`, so it built `page.getByRole('button')` with no `.nth()`.

The resulting locator `page.getByRole('button')` means "any element with role button" — it matches **all 8 buttons** on the page, causing Playwright's strict mode violation.

## Fix

An unnamed `<button></button>`'s accessible name is actually `""` (empty string), not `undefined`. The browser's Accessible Name Computation runs and finds no text content, resulting in `""`. The previous code incorrectly used `undefined` to represent this, which in Playwright's API means "don't filter by name at all."

Now interactive elements with no name store `name: ""`, producing:

```typescript
page.getByRole('button', { name: "", exact: true })
```

This tells Playwright to match only buttons whose accessible name is **exactly the empty string** — which is only the unnamed button. Named buttons like `"Search"` or `"Choose date"` don't match.

**Before:**
```
name: undefined → getByRole('button') → matches ALL 8 buttons → ✗ error
```

**After:**
```
name: "" → getByRole('button', { name: "", exact: true }) → matches 1 button → ✓ success
```

## Test plan

- [x] New test: unnamed button among named buttons resolves and clicks correctly
- [x] `pnpm test` — all 481 tests pass
- [x] E2E verification: `getByRole('button', { name: "", exact: true })` matches exactly 1 element in the #500 scenario

### DIFF
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -929,7 +929,7 @@ async function handleSnapshot(
   });
 
   // Simplify refs for output (just role and name)
-  const simpleRefs: Record<string, { role: string; name?: string }> = {};
+  const simpleRefs: Record<string, { role: string; name: string }> = {};
   for (const [ref, data] of Object.entries(refs)) {
     simpleRefs[ref] = { role: data.role, name: data.name };
   }
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -226,12 +226,10 @@ export class BrowserManager {
     }
 
     // Build locator with exact: true to avoid substring matches
-    let locator: Locator;
-    if (refData.name !== undefined) {
-      locator = page.getByRole(refData.role as any, { name: refData.name, exact: true });
-    } else {
-      locator = page.getByRole(refData.role as any);
-    }
+    let locator: Locator = page.getByRole(refData.role as any, {
+      name: refData.name,
+      exact: true,
+    });
 
     // If an nth index is stored (for disambiguation), use it
     if (refData.nth !== undefined) {
--- a/src/snapshot.ts
+++ b/src/snapshot.ts
@@ -23,7 +23,7 @@ export interface RefMap {
   [ref: string]: {
     selector: string;
     role: string;
-    name?: string;
+    name: string;
     /** Index for disambiguation when multiple elements have same role+name */
     nth?: number;
   };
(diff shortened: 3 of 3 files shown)


ctate:
@jin-2-kakaoent Thanks for the thorough PR! LGTM

PR #577, opened 2026-03-01
fix: typo in skill-creator `SKILL.md`

vladdoster:
`auxilary` => `auxiliary`


ctate:
Thanks for the PR - just went ahead and removed `skill-creator` since it can cause confusing DX

PR #601, opened 2026-03-03
Fix native mode on WSL/Docker and improve CDP error handling

rudygt:
## Summary

Native mode (`--native`) fails silently on WSL and Docker environments. The root cause is missing Chrome sandbox flags, but the debugging experience is also poor due to swallowed errors and hardcoded timeouts.

This PR fixes 4 issues:

- **Add `--no-sandbox` and `--disable-dev-shm-usage` to default Chrome launch args** — required for WSL, Docker, and most headless CI environments. Without these, Chrome silently hangs during `Page.navigate` even though it starts up fine.

- **Auto-discover Puppeteer and Playwright bundled Chrome on Linux** — when no system Chrome is installed (common in dev environments), `find_chrome()` now checks `~/.cache/puppeteer/chrome/` (preferred, unpatched Chrome) and `~/.cache/ms-playwright/` (fallback). Latest version is preferred.

- **Propagate WebSocket death to pending CDP commands immediately** — when Chrome crashes or closes the WebSocket, all pending commands now get an actionable error like `"WebSocket closed by browser: 1001"` instead of silently waiting 30 seconds and returning `"CDP command timed out: Page.navigate"`. A `ws_alive` flag also short-circuits new commands on a dead connection.

- **Add `AGENT_BROWSER_DEBUG_LOG` env var** — redirects native daemon stderr to a file for diagnostics (defaults to `/dev/null` as before). Also adds `AGENT_BROWSER_CDP_TIMEOUT` and honors `AGENT_BROWSER_DEFAULT_TIMEOUT` in the native daemon.

## Test plan

- [x] `cargo test` — 366 tests pass, 0 failures
- [x] `cargo check` — clean compilation
- [x] Manual test: `agent-browser --native open https://example.com` on WSL2 (previously timed out, now works)
- [x] Manual test: snapshot and screenshot work in native mode
- [x] Manual test: auto-discovers Puppeteer Chrome without `--executable-path`
- [ ] Test on Docker (same flags, should also fix native mode there)
- [ ] Test on macOS (no-op for sandbox/shm flags, discovery paths are Linux-only)

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -154,6 +154,10 @@ fn build_chrome_args(options: &LaunchOptions) -> Result<ChromeArgs, String> {
         args.push("--no-sandbox".to_string());
     }
 
+    if !args.iter().any(|a| a == "--disable-dev-shm-usage") {
+        args.push("--disable-dev-shm-usage".to_string());
+    }
+
     Ok(ChromeArgs {
         args,
         temp_user_data_dir,
--- a/cli/src/native/cdp/client.rs
+++ b/cli/src/native/cdp/client.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 
 use futures_util::{SinkExt, StreamExt};
@@ -26,6 +26,7 @@ pub struct CdpClient {
     next_id: AtomicU64,
     pending: PendingMap,
     event_tx: broadcast::Sender<CdpEvent>,
+    ws_alive: Arc<AtomicBool>,
     _reader_handle: tokio::task::JoinHandle<()>,
 }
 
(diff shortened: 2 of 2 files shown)


ctate:
Thanks for this! We've since landed #602 and #614 which cover the Chrome launch args, `find_chrome()` fallback, and daemon stderr surfacing, so there will be conflicts on those files.

The CDP client resilience work in `client.rs` is really valuable though, the `ws_alive` tracking and pending command cleanup would be a big improvement over the current "wait 30s per command" behavior on disconnect. The `--disable-dev-shm-usage` addition is also a good catch.

Would you be up for rebasing onto main and narrowing this down to the `client.rs` changes + the `--disable-dev-shm-usage` flag? Happy to help if you run into any issues.

rudygt:
sure! I will try to get it updated today

rudygt:
Rebased onto main and narrowed to the two changes not covered by #602 / #614:

1. **`client.rs`** — `ws_alive` tracking + pending command drain on WebSocket death
2. **`chrome.rs`** — `--disable-dev-shm-usage` added to launch args

Force-pushed to this same branch to keep the conversation context.

PR #611, opened 2026-03-04
fix: suppress spurious --native warning when set via env var

hewliyang:
## Problem

When `AGENT_BROWSER_NATIVE=1` is set via environment variable, every command after the first shows a spurious warning:

```
⚠ --native ignored: daemon already running. Use 'agent-browser close' first to restart with new options.
```

This is a false positive — the native Rust daemon is already running (confirmed via `ps aux` showing `agent-browser-darwin-arm64`, not `node daemon.js`). The daemon inherited the env var at spawn time, so re-warning on every subsequent command is incorrect & confuses the agent.

## Steps to reproduce

1. `export AGENT_BROWSER_NATIVE=1`
2. `agent-browser close` (ensure clean state)
3. `agent-browser open https://example.com` — ✅ no warning
4. `agent-browser snapshot -i` — ⚠️ spurious warning
5. Verify with `ps aux | grep agent-browser | grep -v grep` — daemon is the native binary, not Node.js

Screenshot 2026-03-04 at 13 23 10

## Root cause

`cli/src/main.rs:466` uses `flags.native.then_some("--native")` in the ignored flags check. `flags.native` is true from both the env var and CLI flag, so it always warns when the daemon is already running — regardless of whether the daemon is already in native mode.

## Fix

Add `cli_native` flag (consistent with existing `cli_*` pattern for `cli_executable_path`, `cli_profile`, etc.) to distinguish CLI origin from env var origin. `cli_native` tracks whether `--native` was passed on the command line at all (not its value), so both `--native` and `--native false` correctly warn when the daemon is already running.

## Testing

Built and verified:
- `AGENT_BROWSER_NATIVE=1` env var + subsequent commands → **no warning** 
- `--native` CLI flag on already-running daemon → **warning fires correctly** 
- `--native false` CLI flag on already-running daemon → **warning fires correctly**

### DIFF
--- a/cli/src/flags.rs
+++ b/cli/src/flags.rs
@@ -570,7 +570,7 @@ pub fn parse_flags(args: &[String]) -> Flags {
             "--native" => {
                 let (val, consumed) = parse_bool_arg(args, i);
                 flags.native = val;
-                flags.cli_native = val;
+                flags.cli_native = true;
                 if consumed {
                     i += 1;
                 }


ctate:
@hewliyang Perfect - thank you!

PR #618, opened 2026-03-04
！issue in windows

hankl:


ctate:
Thanks for the PR! Could you share more detail on the issue you're seeing?

- What are the exact steps to reproduce the error?
- What does the error message look like?

Thanks for taking the time to contribute!

hankl:
> Thanks for the PR! Could you share more detail on the issue you're seeing?
> 
> * What are the exact steps to reproduce the error?
> * What does the error message look like?
> 
> Thanks for taking the time to contribute!

 Run claude code v2.1.38 on windows terminal 1.25.622.0  with agent browser skill：  

❯   ⎿  Error: Bash command failed for pattern "!` characters (history expansion), backticks, and `": [stderr]
       /usr/bin/bash: eval: line 1: syntax error near unexpected token `history'
       /usr/bin/bash: eval: line 1: `characters (history expansion), backticks, and < /dev/null'

@ctate

hankl:
Main branch has simplified the SKILL.md and solved the issue.

PR #619, opened 2026-03-04
fix(native): use reqwest for CDP port discovery instead of broken hand-rolled HTTP client

hewliyang:
## Summary

`reqwest_get_string()` in `cli/src/native/cdp/chrome.rs` hand-rolls HTTP/1.1 over raw TCP despite `reqwest` being an existing dependency (`Cargo.toml`). Might be a remnant of some AI slop. The hand-rolled implementation has two bugs that cause `agent-browser --cdp <port>` to always timeout when `AGENT_BROWSER_NATIVE=1`.

## Bugs

**1. URL path parsing** — `url.find('/')` matches the first `/` in `http://`, producing an invalid request path:
```
Expected: GET /json/version HTTP/1.1
Actual:   GET //127.0.0.1:9222/json/version HTTP/1.1
```

**2. `read_to_end()` hangs** — Chrome's DevTools HTTP server ignores `Connection: close` and keeps the socket open. `read_to_end()` waits for EOF that never comes, hitting the 2-second timeout in `discover_cdp_url`.

The error then gets mangled by `to_ai_friendly_error` into `"Operation timed out. The page may still be loading or the element may not exist."` - masking the real cause.

## Fix

Replace 49 lines of broken hand-rolled TCP with `reqwest::get()`:

```rust
async fn reqwest_get_string(url: &str) -> Result<String, String> {
    let resp = reqwest::get(url).await.map_err(|e| e.to_string())?;
    resp.text().await.map_err(|e| e.to_string())
}
```

## Repro

```bash
# Launch Chrome with remote debugging
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --remote-debugging-port=9222 \
  --user-data-dir="$HOME/.agent-browser/chrome-profile" \
  --no-first-run &

# Verify Chrome responds
curl -s http://localhost:9222/json/version | head -2

# This times out (native mode)
AGENT_BROWSER_NATIVE=1 agent-browser --cdp 9222 open "about:blank"
# ✗ Operation timed out.

# Without native mode works fine (Playwright handles HTTP correctly)
agent-browser --cdp 9222 open "about:blank"
# ✓ about:blank
```

## Note

`to_ai_friendly_error` shadows any error containing "timeout" into a generic page-related message, even for non-page failures like this HTTP fetch. Worth improving separately.

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -381,55 +381,8 @@ pub async fn discover_cdp_url(port: u16) -> Result<String, String> {
 }
 
 async fn reqwest_get_string(url: &str) -> Result<String, String> {
-    let client = tokio::net::TcpStream::connect(
-        url.strip_prefix("http://")
-            .unwrap_or(url)
-            .split('/')
-            .next()
-            .unwrap_or("127.0.0.1:9222"),
-    )
-    .await
-    .map_err(|e| e.to_string())?;
-
-    let path = url
-        .find('/')
-        .and_then(|i| url[i..].find('/').map(|j| &url[i + j..]))
-        .unwrap_or("/json/version");
-
-    let host = url
-        .strip_prefix("http://")
-        .unwrap_or(url)
-        .split('/')
-        .next()
-        .unwrap_or("127.0.0.1");
-
-    let request = format!(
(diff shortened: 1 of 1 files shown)


ctate:
@hewliyang Great fix - thank you!

PR #620, opened 2026-03-04
chore(ci): add cargo fmt check & fix existing violations

hewliyang:
TypeScript CI has `prettier --check` but Rust CI only runs `cargo test`. This adds `cargo fmt --check` and fixes the 8 pre-existing formatting violations on main to prevent future drift in *.rs


ctate:
This is great. Thanks @hewliyang!

PR #621, opened 2026-03-04
Fix issue 279: make browser commands honor frame context

### LINKED ISSUE #279: Feature Request: Add frameLocator support for cross-origin iframe interaction
## Problem

When trying to interact with elements inside cross-origin iframes (such as Apple's authentication iframe on icloud.com), agent-browser cannot access or interact with the iframe content.

## Current Behavior

- `agent-browser snapshot` does not include elements from iframes
- Mouse clicks via `agent-browser mouse move/down/up` click on the iframe but don't focus internal elements
- Keyboard input via `agent-browser press` doesn't reach elements inside the iframe
- `agent-browser fill "iframe >> input" "text"` doesn't work for cross-origin iframes
- `agent-browser eval` cannot access iframe content due to cross-origin security restrictions

## Expected Behavior

Playwright supports iframe interaction via `page.frameLocator()` which properly handles cross-origin frames:

```javascript
const frame = page.frameLocator('iframe[name="auth-widget"]');
await frame.locator('input[type="text"]').fill('email@example.com');
```

## Proposed Solution

Add frameLocator support to agent-browser, potentially via:

1. A new `--frame` flag for commands:
   ```bash
   agent-browser fill --frame "iframe[name='auth']" "input" "text"
   agent-browser click --frame "iframe" "@e1"
   ```

2. Or a frame selector syntax:
   ```bash
   agent-browser fill "frame:iframe[name='auth'] >> input" "text"
   ```

3. Or frame-specific commands:
   ```bash
   agent-browser frame "iframe[name='auth']" fill "input" "text"
   ```

## Use Case

This is critical for automating authentication flows (Apple, Google, OAuth providers) where login forms are embedded in cross-origin iframes for security.

## Environment

- agent-browser version: latest (via npx)
- macOS
- Tested with iCloud authentication (idmsa.apple.com iframe)

## Workaround

Currently, users must use Playwright directly or tools like `mcp__next-devtools__browser_eval` which properly expose frameLocator functionality.

jeffscottward:
## Summary
- Route frame-sensitive commands and snapshots through the active frame instead of main page.
- Update locator resolution for click/fill/get-by/snapshot-related flows using getLocator/getFrame.
- Extend snapshot generation to accept Playwright Frame for iframe-aware snapshots.
- Add regression coverage for frame context in browser manager actions.

Closes #279

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -1,4 +1,5 @@
 use serde_json::{json, Value};
+use std::collections::HashMap;
 use std::env;
 use std::io::Write;
 use std::sync::Arc;
@@ -105,6 +106,7 @@ pub struct DaemonState {
     pub tracked_requests: Vec<TrackedRequest>,
     pub request_tracking: bool,
     pub active_frame_id: Option<String>,
+    pub frame_context_ids: HashMap<String, i64>,
     /// Shared slot for stream server to receive CDP client when browser launches.
     pub stream_client: Option<Arc<RwLock<Option<Arc<CdpClient>>>>>,
 }
@@ -139,6 +141,7 @@ impl DaemonState {
             tracked_requests: Vec::new(),
             request_tracking: false,
             active_frame_id: None,
+            frame_context_ids: HashMap::new(),
             stream_client: None,
         }
     }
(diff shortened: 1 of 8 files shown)


ctate:
Thanks for tackling this, iframe support has been a big gap. The approach is solid and the test coverage is great.

A few commands still use `page` when they should go through `getFrame()`:
- `handleEvaluate` (line 949), probably the most important one since it's the escape hatch for custom JS
- `handleScroll` without a selector (line 1013), `page.evaluate('window.scrollBy(...)')` won't scroll inside a frame
- `handleWait` without a selector (line 969), `page.waitForLoadState` is page-level, `Frame` has it too
- `handleContent` without a selector (line 1055), `Frame` also has `.content()`

Could you update those? No need to worry about the native path for now, we can handle that separately.

jeffscottward:
@ctate

ctate:
Hey @jeffscottward - thanks for the update. Can you rebase plz?

jeffscottward:
@ctate  I did and then there were conflicts hopefully this resolved it. Otherwise at this point I would say good luck and you can close mine and re-implement it's core ideas in a clean branch PR

PR #624, opened 2026-03-04
fix: handle relative URLs in Websocket domain filter script

classiemilio:
### Summary

Fixes https://github.com/vercel-labs/agent-browser/issues/764

When using allowedDomains with a dev server that uses webpack, the domain filter's _checkUrl() function throws TypeError: Failed to construct 'URL': Invalid URL on relative URLs like /__webpack_hmr (webpack's hot module reloading endpoint). This is because `new URL(url)` requires an absolute URL — relative paths need a base URL to resolve against.

This PR passes location.href as the base URL so relative URLs resolve correctly against the current page origin. Also adds a console.warn when URL parsing fails to aid debugging.

 ### Changes

  - Use location.href as base in _checkUrl() to support relative URLs
  - Log a warning with the offending URL on parse failure


classiemilio:
@ctate We ran into this issue trying to leverage agent-browser with a domain allowlist. Trivial fix to ensure relative URLs (in our case the webpack HMR URL was causing issues) get properly validated.

classiemilio:
> Good fix. Passing location.href as the base URL to new URL() is the correct way to handle relative URLs. Since _checkUrl is shared by WebSocket, EventSource, and sendBeacon, fixing it once covers all three.
> 
> **1. Native path WebSocket handler needs the same fix** -- Per AGENTS.md, browser automation changes must be made in both paths. The native implementation in cli/src/native/network.rs already uses location.href for EventSource (line 197) and sendBeacon (line 208), but the WebSocket handler at line 187 still uses new URL(url) without a base. This should be new URL(url, location.href) to match.
> 
> **2. Missing test coverage** -- Consider adding a test to src/domain-filter.test.ts verifying the script contains location.href (e.g. expect(script).toContain('new URL(url, location.href)'))
> 
> **Nit:** The inline comment in the injected script adds payload to every page. The second argument to new URL() is well-understood enough that it could be removed.

@ctate Thanks for the prompt review! Updated with those changes.

classiemilio:
@ctate friendly bump. This is blocking our team's adoption of agent-browser. Thanks.

classiemilio:
@ctate when you have a moment, can I get a re-review on this one? This is blocking our team's adoption of agent-browser. We're trying to replace chrome-devtools-mcp with it and the domain filter support is a main reason for that move.

ctate:
@classiemilio sorry for the delay. I need to set up better alerts. Code looks great - can you plz fix the merge conflicts? I'll publish this asap

PR #625, opened 2026-03-04
Remove BROWSERBASE_PROJECT_ID requirement

derekmeegan:
The Browserbase API no longer requires a project ID to create sessions. It is now inferred from the API key. Remove the env var requirement from both the TypeScript daemon and Rust CLI, and update docs accordingly.


ctate:
Looks great - thank you @derekmeegan!

ctate (approved):

PR #641, opened 2026-03-06
fix: add channel parameter for Chrome extensions content script injection

### LINKED ISSUE #640: [Bug] Chrome extensions not loading - content scripts not injected
## Description

Chrome extensions are not working with agent-browser. Content scripts are not being injected into pages, even though the extension appears to load (background pages are visible in CDP).

Interestingly, using Playwright directly **does work** when `channel: 'chromium'` is specified, suggesting the issue is specific to how agent-browser configures Playwright.

## Reproduction

```bash
# 1. Create a minimal test extension
mkdir -p /tmp/test-extension
cat > /tmp/test-extension/manifest.json <<'EOF'
{
  "manifest_version": 3,
  "name": "Test Extension",
  "version": "1.0",
  "content_scripts": [
    {
      "matches": ["<all_urls>"],
      "js": ["content.js"],
      "run_at": "document_idle"
    }
  ]
}
EOF

cat > /tmp/test-extension/content.js <<'EOF'
console.log('[TEST EXTENSION] Injected!');
window.TEST_EXTENSION_LOADED = true;
document.body.style.border = "5px solid red";
EOF

# 2. Launch with extension
agent-browser close
agent-browser --headed --extension /tmp/test-extension open https://example.com

# 3. Check if content script ran
agent-browser eval 'window.TEST_EXTENSION_LOADED'  # Returns undefined, should be true
agent-browser console  # No "[TEST EXTENSION] Injected!" message
```

## Expected Behavior

Extension content scripts should be injected and execute. When using Playwright directly with the same configuration, it works:

```javascript
import { chromium } from 'playwright';
const context = await chromium.launchPersistentContext(userDataDir, {
  channel: 'chromium',
  args: [`--load-extension=${extPath}`, `--disable-extensions-except=${extPath}`]
});
// Result: content scripts ARE injected ✅
```

## Workaround

Use Playwright directly instead of agent-browser. The following configuration works:

```javascript
import { chromium } from 'playwright-core';

const extensionPath = '/path/to/extension';
const context = await chromium.launchPersistentContext(userDataDir, {
  headless: false,
  channel: 'chromium',  // This is required!
  args: [
    `--disable-extensions-except=${extensionPath}`,
    `--load-extension=${extensionPath}`
  ]
});
```

## Investigation

1. **Fix attempted**: Added `channel: 'chromium'` to `dist/browser.js` line 1147 in the `launchPersistentContext` call (where `hasExtensions` is true)
2. **Result**: Fix works with Playwright directly, but NOT through agent-browser
3. **Both Playwright and Native modes tested** - neither works
4. **Manual Chrome launch with `--load-extension`** also doesn't inject content scripts on macOS Chrome 147

This suggests the issue may be deeper than just the `channel` parameter - possibly related to:
- How agent-browser's daemon caches code
- Chrome 147 + macOS specific behavior
- User data directory state

## Environment

- agent-browser: 0.16.3
- Playwright: 1.58.2
- OS: macOS (Darwin 25.2.0, arm64)
- Chrome: 147.0.7632.160 (system Chrome)

## Related

- Playwright Chrome extensions docs: https://playwright.dev/docs/chrome-extensions
- The docs specify using `channel: 'chromium'` for extension support

jtstothard:
Fixes #640 - Chrome extensions were not loading content scripts.

## Root Cause

Playwright requires `channel: 'chromium'` (or 'chrome') for Chrome extension
content script injection to work properly.

## Changes

- Add `channel` option to `LaunchCommand` type for Chrome channel selection
- Set `channel: 'chromium'` (default) when launching with extensions

## Testing

Tested locally with a test extension - content scripts now inject correctly in both
headed and headless modes (as documented in README).

### DIFF
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1352,9 +1352,13 @@ export class BrowserManager {
       // Combine extension args with custom args and file access args
       const extArgs = [`--disable-extensions-except=${extPaths}`, `--load-extension=${extPaths}`];
       const allArgs = baseArgs ? [...extArgs, ...baseArgs] : extArgs;
+      // Chrome extensions require channel to be set for content script injection
+      // See: https://playwright.dev/docs/chrome-extensions
+      const extensionChannel = options.channel || 'chromium';
       context = await launcher.launchPersistentContext(
         path.join(os.tmpdir(), `agent-browser-ext-${session}`),
         {
+          channel: extensionChannel,
           headless: options.headless ?? true,
           executablePath: options.executablePath,
           args: allArgs,
--- a/src/types.ts
+++ b/src/types.ts
@@ -12,6 +12,7 @@ export interface LaunchCommand extends BaseCommand {
   headless?: boolean;
   viewport?: { width: number; height: number } | null;
   browser?: 'chromium' | 'firefox' | 'webkit';
+  channel?: 'chrome' | 'chrome-beta' | 'chrome-dev' | 'chrome-canary' | 'chromium'; // Chrome channel for extensions support
   headers?: Record<string, string>;
   executablePath?: string;
   cdpPort?: number;


jtstothard:
Didn't actually fix the issue, investigating further what is passing --disable-extensions to chrome

PR #656, opened 2026-03-06
feat: Add React DevTools profiling support with component render tracking

### LINKED ISSUE #643: [Feature] Allow React DevTools usage
I'm using agent-browser practically everywhere in my workflow. Love the project!

One thing on my wishlist is the ability to run React DevTools so that I can have my an agent help me pinpoint performance rendering bottlenecks. I see there's already support for loading Chrome extensions but I was not able to find a direct way to control the extension directly.

My current manual workflow is to run the profile with "Record why each component rendered while profiling", then export a JSON for the agent to analyze. But this requires me to be in the loop as a button presser.

If helpful, I could investigate what's needed for implementing this?

ctate:
Fixes #643

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1037,6 +1037,31 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
             }
         }
 
+        // === React Profiling (component-level render data) ===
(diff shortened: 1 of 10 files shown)


tian000:
Excited for this!

PR #679, opened 2026-03-07
Add regression test for Didomi-style quoted ARIA snapshot buttons

simonellefsen:
## Summary

Add a regression test for the snapshot formatting behavior that originally broke on a Didomi cookie modal on `https://borsen.dk/aktiespil/forside`.

At the time this PR was opened, `agent-browser snapshot -i` could see the consent modal but dropped the actionable buttons from snapshot output. The visible failure was that the modal rendered buttons like:

- `Agree and close`
- `Disagree and close`
- `Learn More`

but the snapshot only surfaced unrelated content such as `our 89 partners`.

The underlying issue was that these controls were represented with a longer accessible name plus a visible value text, for example:

- `button "Agree and close: Agree to our data processing and close" ... : Agree and close`

During review and subsequent rebases, upstream `main` changed substantially and the original implementation path used for the initial fix was removed. After merging the branch forward, the remaining diff is now just the regression test that asserts this button/value-text formatting is preserved.

So this PR is no longer proposing the original implementation change. It now serves as a regression guard for behavior that appears to already be present in current `main`.

## What this PR adds

- a native snapshot regression test that verifies interactive snapshot rendering keeps visible button value text alongside the accessible name
- coverage for the specific shape that originally reproduced on the Didomi consent modal

## Why this still may be useful

Even though the functional fix is no longer part of the diff, this test documents the original failure mode and helps prevent it from silently reappearing in future snapshot refactors.

If maintainers feel existing coverage is already sufficient, closing this PR would also be reasonable.

## Verification

- `cargo test snapshot --manifest-path cli/Cargo.toml`

### DIFF
--- a/cli/src/native/snapshot.rs
+++ b/cli/src/native/snapshot.rs
@@ -1426,6 +1426,35 @@ mod tests {
         assert!(!dups.contains_key("button:Cancel"));
     }
 
+    /// Regression guard for Didomi-style consent buttons, whose accessible name
+    /// is itself a quoted "key: description" string (e.g. "Agree and close:
+    /// Agree to our data processing and close") while the AX value carries the
+    /// short label. Interactive snapshots must keep BOTH: the quoted name and
+    /// the trailing value text. An earlier parser dropped the value suffix,
+    /// leaving agents unable to tell such buttons apart.
+    #[test]
+    fn test_render_tree_keeps_button_value_text_in_interactive_mode() {
+        let mut node = TreeNode::empty();
+        node.role = "button".to_string();
+        node.name = "Agree and close: Agree to our data processing and close".to_string();
+        node.value_text = Some("Agree and close".to_string());
+        node.has_ref = true;
+        node.ref_id = Some("e1".to_string());
+
+        let options = SnapshotOptions {
+            interactive: true,
+            ..SnapshotOptions::default()
+        };
+        let mut output = String::new();
+
+        render_tree(&[node], 0, 0, &mut output, &options);
+
+        assert_eq!(
+            output.trim(),
+            "- button \"Agree and close: Agree to our data processing and close\" [ref=e1]: Agree and close"
+        );
+    }
+
     // -----------------------------------------------------------------------
     // Cursor-interactive text dedup (Issue #841 regression guard)
     // -----------------------------------------------------------------------


ctate (changes requested):
Hey, thanks for working on this! The quoted-key parsing logic looks great.

Heads up though — the new TS test fails. The `suffix` from quoted lines (like `: Agree and close`) gets dropped here because of the `suffix.includes('[')` check. That condition needs to also allow value text through.

Would also be good to update `processLine()` with the same quoted-key handling for the non-interactive path.

simonellefsen:
Thanks for the review. Both code changes are already on the branch: the `suffix.includes('[')` gate was removed in `9929c0d`, and `processLine()` is already using the same `parseAriaLine()` handling for the non-interactive path.

I pushed a follow-up commit, `e4907c3`, to make that second part explicit in test coverage. The PR now includes a full-snapshot regression test in addition to the `snapshot -i` test, so both the interactive path and the `processLine()` path are exercised by the TS suite.

Current targeted check:
- `pnpm exec vitest run src/snapshot.test.ts`

simonellefsen:
Merged the latest `main` into this branch and re-ran the targeted checks.

Current status on this branch:
- `pnpm build` passes
- `pnpm exec vitest run src/snapshot.test.ts` passes
- `cargo test snapshot --manifest-path cli/Cargo.toml` passes

On the suffix point specifically: that is no longer an issue on the current branch. The old `suffix.includes('[')` gate is gone, and both snapshot formatting paths now keep any suffix text via `if (suffix) enhanced += suffix;`.

That applies in both places that matter here:
- the interactive path in `processAriaTree()`
- the non-interactive path through `processLine()`

The TS coverage now exercises both paths:
- one regression test for `snapshot -i`
- one regression test for the full snapshot path, which covers `processLine()`

simonellefsen:
Updated the PR title/body to reflect the current scope after merging forward from `main`.

At this point the branch is no longer carrying the original implementation change. The remaining diff is the regression test that documents the original Didomi/Børsen snapshot failure mode and asserts the current behavior stays intact.

So this can be treated as a small regression-test PR if that coverage is useful, or closed if maintainers feel current upstream coverage is already sufficient.

simonellefsen:
@ctate — apologies for the long silence on this. Coming back to it, the requested changes turned out not to be applicable any more, so let me lay out what I found rather than just pushing.

Both things you pointed at were in the TypeScript implementation:

> The `suffix` from quoted lines (like `: Agree and close`) gets dropped here because of the `suffix.includes('[')` check.
> Would also be good to update `processLine()` with the same quoted-key handling for the non-interactive path.

That code was removed in #754 ("full native"), which landed later the same day as this review — it deleted `src/*.ts` wholesale. There's no `suffix.includes('[')`, no `processLine()`, and no TS test in the tree today, so there's nothing left to change on that front.

So I checked whether the underlying behaviour still holds in the native renderer, since that's what actually matters. It does: ported to current `main`, the test passes unmodified. `render_tree` keeps both the quoted name and the trailing value text in interactive mode.

I've rebased onto current `main` and rewritten the test to use `TreeNode::empty()`, which didn't exist when this was opened — it replaces the 17-field struct literal and won't need touching every time a field is added.

That leaves this PR as a pure regression guard: it locks in behaviour that already works, so the Didomi case can't silently break again. Still cheap and still worth having, in my view — but it's a smaller claim than the original PR made, and if you'd rather close it as overtaken by the rewrite, that's a completely reasonable call and I won't be precious about it.

`cargo fmt --check` clean; 1233 tests pass.

PR #706, opened 2026-03-10
feat: Support for screenshot annotate on rust

notHide:
- enable the screenshot feature of rust native to support the `--annotate` compared to Node.js. self-test passed in the cdp url scenario.
- fixed an issue where error messages containing special characters could cause incorrect JSON structures to be returned in certain scenarios.


ctate:
@notHide Nice work! Tested both changes and they look good

PR #708, opened 2026-03-10
fix: remove npx from allowed-tools to avoid Bash permission check error

kangjinghang:
## Problem

The `/agent-browser` skill fails with:

```
Bash command permission check failed for pattern "!` characters (history expansion), backticks, and `": This command uses shell operators that require approval for safety
```

## Root Cause

The `npx` command is a shell function created by nvm (Node Version Manager), not an executable file. When the Bash permission checker validates `Bash(npx agent-browser:*)`, it attempts to parse this shell function, which triggers a false positive security check.

## Solution

Remove `Bash(npx agent-browser:*)` from `allowed-tools`, keeping only `Bash(agent-browser:*)`. Since agent-browser is typically installed globally, `npx` is not needed and direct invocation works correctly.

This fix resolves the permission check error while maintaining full functionality.

### DIFF
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -94,6 +94,61 @@ async fn create_storage_state_with_cookie(path: &str, cookie_name: &str, cookie_
     assert_success(&resp);
 }
 
+async fn send_raw_http_request(port: u64, request: &str) -> String {
+    let mut stream = tokio::net::TcpStream::connect(format!("127.0.0.1:{port}"))
+        .await
+        .expect("HTTP client should connect to stream server");
+    stream
+        .write_all(request.as_bytes())
+        .await
+        .expect("HTTP request should be written");
+    stream
+        .shutdown()
+        .await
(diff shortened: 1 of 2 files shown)


ctate (changes requested):
Thanks for the fix — root cause analysis is spot-on! Before merging, could you also remove `Bash(npx agent-browser:*)` from `skills/slack/SKILL.md`, `skills/electron/SKILL.md`, and `skills/dogfood/SKILL.md`? They have the same pattern and will hit the same error. Happy to merge once those are updated.

wanderer-yk:
@ctate Updated as requested — removed `Bash(npx agent-browser:*)` from all SKILL.md files (core, electron, slack, dogfood, agentcore). Ready for re-review, thanks!

wanderer-yk:
Hi @ctate — the commits are now signed (verified signatures) and the Bash(npx agent-browser:*) removal has been applied to all SKILL.md files (core, electron, slack, dogfood, agentcore) as requested. Could you please re-review when you have a moment? Thanks!

PR #718, opened 2026-03-11
fix: inherit current viewport for recordings

Walter0697:
## Summary
- make `startRecording()` reuse the active page viewport when one exists
- keep the existing `1280x720` fallback when there is no current page viewport
- add a regression test covering recording after `setViewport(440, 956)`

## Problem
`agent-browser record start` recreated a fresh Playwright context with a hardcoded viewport and recording size of `1280x720`.

That meant:
- the active browser session could be in a portrait/mobile viewport
- but the recording context would still switch back to landscape
- resulting videos did not match the layout the user was actually testing

## Root Cause
`src/browser.ts` used:

```ts
const viewport = { width: 1280, height: 720 };
this.recordingContext = await this.browser.newContext({
  viewport,
  recordVideo: {
    dir: this.recordingTempDir,
    size: viewport,
  },
  storageState,
});
```

## Fix
Use `currentPage?.viewportSize()` when available and only fall back to `1280x720` when no active page viewport exists.

## Validation
Local validation after patching the installed build showed:
- `set viewport 320 960` followed by `record start` produced a `320x960` `.webm`
- `set viewport 440 956 3` followed by `record start` produced a `440x956` portrait recording

## Notes
I added a regression test in `src/browser.test.ts` for the inherited viewport behavior.

I did not run the upstream test suite in the fork clone because dependencies were not installed in that checkout.

### DIFF
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -2,6 +2,9 @@ import { describe, it, expect, beforeAll, afterAll, beforeEach, afterEach, vi } 
 import { BrowserManager, getDefaultTimeout } from './browser.js';
 import { executeCommand } from './actions.js';
 import { chromium } from 'playwright-core';
+import os from 'node:os';
+import path from 'node:path';
+import { existsSync, rmSync } from 'node:fs';
 
 describe('BrowserManager', () => {
   let browser: BrowserManager;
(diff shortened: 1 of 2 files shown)


ctate:
Great catch - thanks @Walter0697!

PR #719, opened 2026-03-11
fix: escape `!` in markdown code block to avoid bash permission check false positive

Cerdore:
Claude Code's bash permission checker incorrectly flags `!` characters inside markdown inline code blocks as bash history expansion commands, blocking the skill from loading. Replace `!` with HTML entity `&#33;`.

Related: https://github.com/anthropics/claude-code/issues/17267


ctate:
Thanks @Cerdore\! I couldn't reproduce this — the skill loads fine with `!` on CC 2.1.71. Could you share repro steps?

PR #720, opened 2026-03-11
fix: sanitize lone Unicode surrogates using toWellFormed()

jin-2-kakaoent:
## Problem

Commands that return strings containing lone Unicode surrogates fail with:

```
Invalid response: unexpected end of hex escape at line 1 column 56
```

`JSON.stringify` escapes lone surrogates (e.g. `\uD800`) in the output, but Rust's `serde_json` expects valid surrogate pairs — causing a parse error.

## Fix

Use `String.prototype.toWellFormed()` (ES2024) in `JSON.stringify`'s replacer to replace lone surrogates with U+FFFD before serialization.

```ts
export function serializeResponse(response: Response): string {
  return JSON.stringify(response, (_key, value) =>
    typeof value === 'string' && !value.isWellFormed() ? value.toWellFormed() : value
  );
}
```

Also upgrades `tsconfig.json` target/lib from ES2022 → ES2024 (CI tests Node 20+, which fully supports ES2024).

## Reproduction

```bash
agent-browser eval 'String.fromCharCode(0xD800)'
# Before: ✗ Invalid response: unexpected end of hex escape at line 1 column 56
# After:  ✓ outputs � (U+FFFD replacement character)
```

### DIFF
--- a/src/protocol.test.ts
+++ b/src/protocol.test.ts
@@ -1424,15 +1424,6 @@ describe('parseCommand', () => {
       expect(parsed.data.snapshot).not.toMatch(/[\uD800-\uDFFF]/);
     });
 
-    it('should serialize response without surrogates normally', () => {
-      const response = successResponse('1', {
-        snapshot: 'button "Submit"',
-      });
-      const json = serializeResponse(response);
-      const parsed = JSON.parse(json);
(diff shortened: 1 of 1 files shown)


ctate:
@jin-2-kakaoent Thanks for this excellent contribution!

PR #722, opened 2026-03-11
feat: add idle timeout to daemon to prevent orphaned Chrome processes

### LINKED ISSUE #721: Daemon has no idle timeout — Chrome processes persist indefinitely after session use
## Bug Description

The daemon process (`daemon.js`) has no idle timeout mechanism. After a browser session is used (e.g., `open`, `snapshot`, `click`), the daemon and all its spawned Chromium processes remain running indefinitely until `close` is explicitly called.

Any workflow that opens a browser but doesn't explicitly close it leaves behind orphaned processes that accumulate over time.

## Reproduction

```bash
# 1. Open a page (spawns daemon + Chromium)
agent-browser open "https://example.com" --session test-leak

# 2. Do some work
agent-browser snapshot --session test-leak

# 3. Don't call close — just move on to other work
# ... hours/days later:

# 4. Daemon + Chrome still running
ps aux | grep agent-browser
# → daemon.js still alive

ps aux | grep chrome | grep -v grep
# → 4+ Chromium processes per session still running
```

## Root Cause

In `daemon.js`, `process.stdin.resume()` keeps the event loop alive indefinitely. Shutdown only occurs via:
- Explicit `close` command
- Process signals: SIGINT, SIGTERM, SIGHUP

There is no idle timeout — if no commands arrive on the socket, the daemon waits forever.

## Impact

Each abandoned session leaves behind:
- 1 Node.js daemon process
- 4+ Chromium processes (main, GPU, network utility, renderer)
- Stale `.pid` and `.sock` files in `~/.agent-browser/`

Over time this leads to dozens of orphaned Chrome processes consuming significant memory and CPU.

## Suggested Fix

Add an idle timeout to the daemon that shuts down the browser and exits after N minutes of inactivity (e.g., 10-15 min):

```js
let idleTimer;
const IDLE_TIMEOUT_MS = 10 * 60 * 1000; // 10 minutes

function resetIdleTimer() {
  clearTimeout(idleTimer);
  idleTimer = setTimeout(() => {
    console.log('Idle timeout reached, shutting down');
    shutdown();
  }, IDLE_TIMEOUT_MS);
}

// Call resetIdleTimer() on each incoming socket command
```

This preserves current behavior for active sessions while preventing indefinite resource leaks.

## Environment

- macOS (Apple Silicon)
- Observed on agent-browser v0.8.4, also present in v0.17.1 (no idle timeout in current codebase)
- Used via `agent-browser-mcp`

CommerceMax:
## Summary

- Adds a configurable idle timeout (default 15 minutes) that automatically shuts down the daemon when no commands arrive on the socket
- Timer resets on every incoming command, so active sessions are completely unaffected
- Set `AGENT_BROWSER_IDLE_TIMEOUT_MS=0` to disable and preserve old behavior

Fixes #721

## Problem

The daemon process (`daemon.js`) persists indefinitely after browser sessions are used. `process.stdin.resume()` keeps the event loop alive with no idle detection. Each abandoned session leaves behind 1 Node.js daemon + 4+ Chromium processes (main, GPU, network utility, renderer) that accumulate over time.

## Implementation

- New env var `AGENT_BROWSER_IDLE_TIMEOUT_MS` (default: 900000 = 15 min, set to 0 to disable)
- `resetIdleTimer()` called on daemon startup and on every incoming socket command
- Timer uses `.unref()` so it doesn't keep the process alive on its own
- On expiry, calls existing `shutdown()` which cleanly closes browser, server, socket, and PID files
- Timer cleared in `shutdown()` to prevent races

## Test plan

- [x] `npm run build` — compiles clean
- [x] `npx vitest run src/daemon.test.ts` — all 18 existing tests pass
- [x] Manual test: `AGENT_BROWSER_IDLE_TIMEOUT_MS=3000` — daemon auto-shuts down after 3s idle
- [x] Manual test: `AGENT_BROWSER_IDLE_TIMEOUT_MS=0` — daemon stays alive indefinitely (old behavior)
- [ ] Verify active sessions are not interrupted by the timer resetting correctly

### DIFF
--- a/src/daemon.ts
+++ b/src/daemon.ts
@@ -696,6 +696,26 @@ export async function startDaemon(options?: {
       idleTimer = null;
     }
 
+    // Auto-save session state before closing (same as the explicit `close` command path)
+    if (manager instanceof BrowserManager && manager.isLaunched()) {
+      const savePath = getSessionSaveStatePath();
+      if (savePath) {
+        try {
+          const { encrypted } = await saveStateToFile(manager, savePath);
+          fs.chmodSync(savePath, 0o600);
+          if (process.env.AGENT_BROWSER_DEBUG === '1') {
+            console.error(
+              `Auto-saved session state before shutdown: ${savePath}${encrypted ? ' (encrypted)' : ''}`
+            );
+          }
+        } catch (err) {
+          if (process.env.AGENT_BROWSER_DEBUG === '1') {
+            console.error(`Failed to auto-save session state before shutdown:`, err);
+          }
+        }
+      }
+    }
+
     // Stop stream server if running
     if (streamServer) {
       await streamServer.stop();


ctate (commented):
Thanks for this — clean implementation and a real problem worth solving!

One thing: could we default to `0` (disabled) instead of 15 minutes? When the idle timeout fires, it calls `shutdown()` which kills the browser without saving state (unlike the explicit `close` path which calls `saveStateToFile`). So if a user comes back after 15 idle minutes, their next command silently spawns a fresh browser — losing all tabs, cookies, and login sessions.

This could bite in headed/interactive mode, CI pipelines with long gaps between browser steps, or just someone stepping away. Defaulting to opt-in (`AGENT_BROWSER_IDLE_TIMEOUT_MS=900000`) avoids a breaking change while keeping the feature available.

ctate:
Perfect. Thanks @CommerceMax!

PR #731, opened 2026-03-12
fix: allow newTab() in persistent context (--extension/--profile) mode

### LINKED ISSUE #411: `tab new` fails with "Browser not launched" when using `--extension` flag
### Description

  When using `--extension` to load a Chrome extension, the `tab new` command fails with "Browser not launched" even though the browser is running and responsive to other commands.

  ### Steps to Reproduce

  ```bash
  # Launch with extension
  agent-browser --session test --extension /path/to/extension open "chrome-extension://extensionid/page.html"

  # This works fine
  agent-browser --session test tab list
  # → [0] My Extension - chrome-extension://extensionid/page.html

  # This fails
  agent-browser --session test tab new
  # ✗ Browser not launched

  # This also fails
  agent-browser --session test tab new https://example.com
  # ✗ Browser not launched
```

##  Expected Behavior

  `tab new` should create a new tab in the existing browser context.

## Actual Behavior

 `tab new` throws "Browser not launched" despite the browser being open and other commands (tab list, eval, screenshot, click, etc.) all working correctly.

 ##Root Cause

  In `src/browser.ts`, the `newTab()` method checks:

```
  async newTab(): Promise<{ index: number; total: number }> {
    if (!this.browser || this.contexts.length === 0) {
      throw new Error('Browser not launched');
    }
    // ...
  }
```
  When `--extension` is used, the launch path calls `launcher.launchPersistentContext()` which returns a BrowserContext directly. It does not call `launcher.launch()`, so `this.browser` is never assigned. The context is correctly
  pushed to `this.contexts`, but the `!this.browser` check fails.

  The same issue likely affects `newWindow()` which has the same guard:

```
  async newWindow(): Promise<{ index: number; total: number }> {
    if (!this.browser) {
      throw new Error('Browser not launched');
    }
    // ...
  }
```

##  Suggested Fix

  Change the guard in `newTab()` to only check for contexts:

```
  if (this.contexts.length === 0) {
    throw new Error('Browser not launched');
  }
```

This is safe because `this.contexts` is only populated during launch, so an empty array genuinely means the browser isn't running. The `this.browser` check is redundant here since persistent contexts don't use it.

 
## Workaround

  Use eval with `window.open()` to create new tabs:
```
  agent-browser --session test eval "void(window.open('https://example.com', '_blank'))"
```

  The new tab is tracked correctly by `setupContextTracking` and appears in tab list.

### PR DESCRIPTION
**Pull request by mikewong23571**
## Problem

When launched with `--extension` or `--profile`, Playwright uses `launchPersistentContext()` which:
- Sets `this.isPersistentContext = true`
- Leaves `this.browser` as `null`

The guard in `newTab()` checked `!this.browser`, causing a false `Browser not launched` error even though the browser was running and other commands (e.g. `tab list`, `eval`, `screenshot`) worked fine.

## Root Cause

```typescript
// src/browser.ts — before
async newTab() {
  if (!this.browser || this.contexts.length === 0) { // this.browser is null in persistent context mode
    throw new Error('Browser not launched');
  }
```

`isLaunched()` (line 167) already handles both launch paths correctly:

```typescript
isLaunched(): boolean {
  return this.browser !== null || this.isPersistentContext;
}
```

`newTab()` just wasn't using it.

## Fix

- `newTab()`: replace `!this.browser` with `!this.isLaunched()`
- `newWindow()`: same guard, but `newWindow` genuinely requires a `Browser` object to create a new context and cannot work in persistent context mode — improve the error message to make this clear

## Tests

Added integration tests that launch with a temporary profile directory (triggering the persistent context path) and verify `newTab()` succeeds:

```
✓ should report as launched in persistent context mode
✓ should create new tab in persistent context mode without throwing
```

Fixes #411

### DIFF
--- a/src/browser.test.ts
+++ b/src/browser.test.ts
@@ -1243,6 +1243,32 @@ describe('BrowserManager', () => {
   });
 });
 
+describe('BrowserManager (persistent context / --profile mode)', () => {
+  let profileBrowser: BrowserManager;
+  let tmpProfileDir: string;
+
+  beforeAll(async () => {
+    tmpProfileDir = path.join(os.tmpdir(), `agent-browser-test-profile-${Date.now()}`);
+    profileBrowser = new BrowserManager();
+    await profileBrowser.launch({ headless: true, profile: tmpProfileDir });
+  });
+
+  afterAll(async () => {
+    await profileBrowser.close();
+    rmSync(tmpProfileDir, { recursive: true, force: true });
+  });
+
+  it('should report as launched in persistent context mode', () => {
+    expect(profileBrowser.isLaunched()).toBe(true);
+  });
+
+  it('should create new tab in persistent context mode without throwing', async () => {
+    const result = await profileBrowser.newTab();
+    expect(result.index).toBe(1);
+    expect(result.total).toBe(2);
+  });
+});
+
 describe('getDefaultTimeout', () => {
   const originalEnv = { ...process.env };
 
--- a/src/browser.ts
+++ b/src/browser.ts
@@ -1784,7 +1784,7 @@ export class BrowserManager {
    * Create a new tab in the current context
    */
   async newTab(): Promise<{ index: number; total: number }> {
-    if (!this.browser || this.contexts.length === 0) {
+    if (!this.isLaunched() || this.contexts.length === 0) {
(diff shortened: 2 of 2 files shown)


ctate:
@mikewong23571 Thanks for the clean fix

PR #734, opened 2026-03-12
fix: align native daemon port hash with client on Windows

### LINKED ISSUE #705: [Windows] Native mode daemon fails to start — port hash mismatch between client and daemon
The native daemon (`export AGENT_BROWSER_NATIVE=1`) fails on Windows with:

```
Daemon failed to start (port: 127.0.0.1:50838)
```

Image

## Root cause (suspected)

The port hash function `get_port_for_session` has different implementations in the client and the native daemon:

| | Client (`cli/src/connection.rs:142`) | Native Daemon (`cli/src/native/daemon.rs:260`) |
|---|---|---|
| Type | `i32` | `i64` |
| Iteration | `.chars()` (Unicode code point) | `.bytes()` (raw bytes) |
| Algorithm | `(hash << 5) - hash + c` (djb2) | `hash * 31 + b` (Java hashCode) |

This causes them to compute different ports for the same session name. The daemon binds on port X, but the client polls port Y, times out after 5 seconds, and reports the failure.

The Node.js daemon (`src/daemon.ts:188`) uses the same djb2 algorithm as the client, so default (non-native) mode works fine. Unix platforms are unaffected because they use Unix domain sockets instead of TCP ports.

## To reproduce

Run `./agent-browser.exe open https://news.ycombinator.com && agent-browser.exe wait --load networkidle && ./agent-browser.exe snapshot -i
` on Windows.

mikewong23571:
## Problem

On Windows with `AGENT_BROWSER_NATIVE=1`, the native daemon fails to start:

```
✗ Daemon failed to start (port: 127.0.0.1:50838)
```

The root cause is that `get_port_for_session()` has two incompatible implementations. The client (`connection.rs`) and native daemon (`native/daemon.rs`) compute different ports for the same session name, so the daemon binds port X while the client polls port Y.

| | `connection.rs` (client) | `native/daemon.rs` (daemon) |
|---|---|---|
| Type | `i32` | `i64` |
| Iteration | `.chars()` (Unicode codepoints) | `.bytes()` (raw bytes) |
| Algorithm | djb2: `(hash << 5) - hash + c` | Java hashCode: `hash * 31 + b` |

For `session = "default"`:
- Client computes **50838** (the port shown in the error message)
- Daemon binds **51174**

The Node.js daemon (`daemon.ts`) already uses the same djb2 as the client, so non-native mode works fine. Unix is unaffected because it uses Unix domain sockets.

## Fix

Align `native/daemon.rs` to use the identical djb2 algorithm from `connection.rs`.

## Test plan

- [x] On Windows with `AGENT_BROWSER_NATIVE=1`: `agent-browser open https://example.com` no longer reports daemon startup failure
- [x] Verified mathematically: both functions now return the same port for `"default"` and any other ASCII session name

Fixes #705

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -792,4 +792,13 @@ mod tests {
         assert!(!is_transient_error("Permission denied"));
         assert!(!is_transient_error("Daemon not found"));
     }
+
+    #[test]
+    #[cfg(windows)]
+    fn test_get_port_for_session() {
+        assert_eq!(get_port_for_session("default"), 50838);
+        assert_eq!(get_port_for_session("my-session"), 63105);
+        assert_eq!(get_port_for_session("work"), 51184);
+        assert_eq!(get_port_for_session(""), 49152);
+    }
 }
--- a/cli/src/native/daemon.rs
+++ b/cli/src/native/daemon.rs
@@ -258,9 +258,26 @@ fn get_daemon_socket_dir() -> PathBuf {
 
 #[cfg(windows)]
 fn get_port_for_session(session: &str) -> u16 {
-    let mut hash: i64 = 0;
-    for b in session.bytes() {
-        hash = hash.wrapping_mul(31).wrapping_add(b as i64);
+    let mut hash: i32 = 0;
+    for c in session.chars() {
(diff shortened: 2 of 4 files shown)


ctate:
Thanks for the fix @mikewong23571

Issue #479, opened 2026-02-16
Add `--wait-until` flag to `open` command and change default to `domcontentloaded`


go-to-the-future:
## Problem

The `open` command currently uses `waitUntil: 'load'` by default when navigating. This frequently causes timeouts on modern SPAs that rely on external scripts (analytics, auth providers, etc.) which delay the `load` event indefinitely.

This is especially problematic for AI agent workflows where the agent navigates to a URL and expects to interact with the page promptly. A timeout here wastes an entire agent turn and can cascade into repeated failures.

## Observation: Internal Inconsistency

The daemon already has an inconsistency in `waitUntil` defaults:

- **`handleNavigate`** (`src/actions.ts`): uses `command.waitUntil ?? 'load'`
- **`handleTabNew`** (`src/actions.ts`): hardcodes `waitUntil: 'domcontentloaded'`

```typescript
// handleNavigate - defaults to 'load'
await page.goto(command.url, { waitUntil: command.waitUntil ?? 'load' });

// handleTabNew - uses 'domcontentloaded'
await page.goto(command.url, { waitUntil: 'domcontentloaded' });
```

The `tab new` command already chose the more practical default.

### 1. Add `--wait-until` CLI flag to `open` command

The daemon already supports the `waitUntil` field on `NavigateCommand` (`src/types.ts`), but the Rust CLI (`cli/src/commands.rs`) never passes it through:

```rust
// Current (line 102)
let mut nav_cmd = json\!({ "id": id, "action": "navigate", "url": url });
// waitUntil is never set
```

Adding a `--wait-until` flag (accepting `load`, `domcontentloaded`, `networkidle`) would expose this existing capability:

```bash
agent-browser open https://example.com --wait-until domcontentloaded
```

### 2. Change default from `load` to `domcontentloaded`

The Playwright community consensus is that `domcontentloaded` is the better default for most use cases:

- `load` waits for **all** resources (images, stylesheets, iframes, third-party scripts) — often unnecessary and unreliable
- `domcontentloaded` fires when the HTML is parsed and DOM is ready — sufficient for most interactions
- Agent workflows already follow a `navigate → snapshot → interact` pattern, so waiting for full `load` provides no additional value

This would also align `handleNavigate` with `handleTabNew`.

## Changes Required

| File | Change |
|---|---|
| `cli/src/flags.rs` | Add `wait_until: Option<String>` field + `--wait-until` parsing |
[...]

PR #739, opened 2026-03-12
docs: add guidance for apps with persistent connections (SSE/WebSocket)

mecampbellsoup:
## Summary

- Adds a "Apps with Persistent Connections (SSE, WebSockets)" subsection under "Timeouts and Slow Pages" in `skills/agent-browser/SKILL.md`
- Explains why `wait --load networkidle` always times out on apps with SSE, WebSocket, or long-polling connections
- Provides alternative wait strategies: `domcontentloaded`, element waits, text waits, JS condition waits

## Context

We hit this while using agent-browser to QA a Django app that maintains an SSE connection on every authenticated page (for real-time chat updates). `wait --load networkidle` timed out every time because Playwright's networkidle requires zero network activity for 500ms, which never happens with a persistent connection.

The current SKILL.md recommends `networkidle` in multiple code examples without noting this limitation. This affects any app using SSE, WebSockets, long-polling, or similar persistent connection patterns — which includes most real-time web apps (chat, dashboards, collaborative editors, notification systems).

Related to #479 (changing `open` default to `domcontentloaded`).

## Test plan

- [x] Verified the timeout behavior against a real SSE app
- [x] Confirmed `wait --load domcontentloaded` works as a replacement
- [ ] Review: does this section belong here or in `references/commands.md`?

### DIFF
--- a/skills/agent-browser/SKILL.md
+++ b/skills/agent-browser/SKILL.md
@@ -451,6 +451,30 @@ agent-browser wait 5000
 
 When dealing with consistently slow websites, use `wait --load networkidle` after `open` to ensure the page is fully loaded before taking a snapshot. If a specific element is slow to render, wait for it directly with `wait <selector>` or `wait @ref`.
 
+### Apps with Persistent Connections (SSE, WebSockets)
+
+**`wait --load networkidle` will always time out** on apps that maintain persistent server connections — this includes any app using Server-Sent Events (SSE), WebSockets, or long-polling. Playwright's `networkidle` waits for zero network activity for 500ms, which never happens when a connection is held open.
+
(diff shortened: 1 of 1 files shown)


ctate (changes requested):
Nice catch — I hit the same timeout on a test SSE server and confirmed all four alternatives work. Two small asks: update the `networkidle` code comment to something like `# Wait for zero network activity for 500ms (times out if page uses SSE/WebSocket)`, and add a "see below" note on the paragraph right above your new section so readers get the heads-up before copy-pasting `networkidle`. Otherwise this is good to go!

ctate:
Thanks for the pr @mecampbellsoup. Addressed in #1834

PR #748, opened 2026-03-13
fix: restore BrowserManager.navigate() and package entry point

### LINKED ISSUE #307: Property 'navigate' does not exist on type 'BrowserManager'
import { BrowserManager } from 'agent-browser';

const browser = new BrowserManager();
await browser.launch({ headless: true });
await browser.navigate('https://example.com');

Image

mikewong23571:
## Problem

`BrowserManager` was not usable as a programmatic API. Three things were broken simultaneously:

```ts
import { BrowserManager } from 'agent-browser'; // ❌ main pointed at dist/daemon.js
const browser = new BrowserManager();
await browser.launch({ headless: true });         // ❌ required id + action fields
await browser.navigate('https://example.com');    // ❌ method did not exist
```

Reported in #307 with a TypeScript screenshot showing `Property 'navigate' does not exist on type 'BrowserManager'`.

## Root Cause

- `package.json` `main` pointed at `dist/daemon.js` (the internal daemon process), so any `import`/`require` of the package received the daemon module instead of a usable class
- `BrowserManager.launch()` accepted `LaunchCommand` which required `id` and `action` — IPC-protocol fields that make no sense for programmatic callers
- `navigate()` existed only as a private function inside `src/actions.ts`, never exposed on `BrowserManager`

## Fix

**`src/index.ts`** — new package entry point that re-exports `BrowserManager` and all public symbols.

**`src/browser.ts`**
- Add `BrowserLaunchOptions` (`Pick<LaunchCommand>` minus the IPC-only `id`, `action`, `engine` fields) — decouples the programmatic API from the wire protocol
- Change `launch()` signature from `LaunchCommand` → `BrowserLaunchOptions`
- Add `BrowserManager.navigate(url, options?)` — consolidates domain-allow-list check, scoped-header setup, and `page.goto()` into one reusable method. If the browser is launched but all pages have been closed, a new page is created automatically before navigating.
- Add `BrowserManager.getUrl()` and `getTitle()` convenience methods

**`src/actions.ts`** — `handleNavigate()` now delegates to `browser.navigate()` instead of duplicating the logic.

**`package.json`** — `main` → `./dist/index.js`, add `types` → `./dist/index.d.ts`, add `exports["."]` with `import`/`types`/`default` conditions.

## Tests Added

| Test | What it covers |
|------|---------------|
| `navigate()` returns `{ url, title }` | basic happy path |
| `navigate()` with `headers` option | header path exercised end-to-end |
| `navigate()` with `waitUntil` option | option forwarding to Playwright |
| `navigate()` blocks domain outside `allowedDomains` | security path throws `Navigation blocked` |
| `navigate()` allows domain inside `allowedDomains` | domain filter + navigate coexist |
| `navigate()` blocks non-http(s) scheme | scheme guard is enforced |
| `getUrl()` / `getTitle()` | convenience method return values |
| Type assertion: `PublicBrowserManager` has `navigate` | compile-time proof the public entry exports the right type (direct repro of #307) |
| Type assertion: `BrowserLaunchOptions` excludes `id`/`action`/`engine` | IPC fields do not leak into the API |

## Validation

All checks pass locally against Node 20 and 22:

```
pnpm typecheck   → no errors
pnpm format:check → All matched files use Prettier code style!
pnpm test        → 506 passed | 17 skipped (serverless/e2e, expected)
pnpm build       → clean tsc output
node scripts/check-version-sync.js → Versions are in sync: 0.18.0
```

Fixes #307

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,16 @@
   "version": "0.18.0",
   "description": "Headless browser automation CLI for AI agents",
   "type": "module",
-  "main": "dist/daemon.js",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./package.json": "./package.json"
+  },
   "files": [
     "dist",
     "bin",
--- a/src/actions.ts
+++ b/src/actions.ts
@@ -621,23 +621,12 @@ async function handleNavigate(
   command: NavigateCommand,
   browser: BrowserManager
 ): Promise<Response<NavigateData>> {
-  browser.checkDomainAllowed(command.url);
-
-  const page = browser.getPage();
-
-  // If headers are provided, set up scoped headers for this origin
-  if (command.headers && Object.keys(command.headers).length > 0) {
-    await browser.setScopedHeaders(command.url, command.headers);
-  }
-
-  await page.goto(command.url, {
-    waitUntil: command.waitUntil ?? 'load',
+  const result = await browser.navigate(command.url, {
+    headers: command.headers,
+    waitUntil: command.waitUntil,
   });
 
-  return successResponse(command.id, {
-    url: page.url(),
-    title: await page.title(),
-  });
+  return successResponse(command.id, result);
 }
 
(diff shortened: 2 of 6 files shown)


ctate:
Thanks for this excellent contribution!

PR #757, opened 2026-03-13
fix: respect --headed false flag in CLI

### LINKED ISSUE #743: Playwright 1.58.x on Linux servers without X Server
Please add the following section to `SKILL.md`, right before the `## Quick start` heading:

---

**Purpose**: Resolve compatibility issues between agent-browser 0.17.1 and Playwright 1.58.x on Linux servers without X Server.

**Rationale**:
- Playwright 1.58+ introduced a separate chromium-headless-shell executable.
- The `find_playwright_chromium()` function in agent-browser only searches for `chromium-*` directories, ignoring `chromium_headless_shell-*`.
- This causes it to incorrectly launch regular Chrome in headless mode, which fails due to the absence of X Server.

**Solution**:
- agent-browser natively supports the `--executable-path` argument and the environment variable `AGENT_BROWSER_EXECUTABLE_PATH`.
- By explicitly specifying the headless shell path, you can bypass the above issue.

## Headless Server Configuration

**Required for Linux servers without X Server (Playwright 1.58+)**

Playwright 1.58+ uses a dedicated headless shell binary. On headless Linux servers, you must specify the headless shell path.

### Option 1: Environment Variable (Recommended)

```bash
export AGENT_BROWSER_EXECUTABLE_PATH="$HOME/.cache/ms-playwright/chromium_headless_shell-1208/chrome-headless-shell-linux64/chrome-headless-shell"
```

Add to `~/.bashrc` or `~/.zshrc` for persistence.

### Option 2: Command-line Flag

```bash
agent-browser open <url> --executable-path "/root/.cache/ms-playwright/chromium_headless_shell-1208/chrome-headless-shell-linux64/chrome-headless-shell"
```

### Verify Headless Shell Exists

```bash
ls -la $HOME/.cache/ms-playwright/chromium_headless_shell-*/
```

If missing, run: `npx playwright install chromium`

---

This section explains the problem and provides clear configuration steps for users running on headless Linux servers with Playwright 1.58+.

QuietyAwe:
Fixes #743

## Problem
When user explicitly sets `--headed false`, the CLI was ignoring this flag. The launch condition only checked if `flags.headed` was true, so `--headed false` would not trigger a launch command. Subsequent commands would auto-launch with default `headless=true`, overriding user's explicit setting.

## Solution
- Added `cli_headed` flag to track when user explicitly sets `--headed` (regardless of value)
- Include this in the launch condition check
- Pass `--headed` to daemon when explicitly set by user

## Changes
- `cli/src/flags.rs`: Add `cli_headed` tracking flag
- `cli/src/main.rs`: Check `cli_headed` in launch condition, pass to daemon
- `cli/src/commands.rs`: Update test fixtures

## Testing
```bash
# Before fix: headless=true (ignored --headed false)
agent-browser --headed false navigate https://example.com

# After fix: headless=false (respects user setting)
agent-browser --headed false navigate https://example.com
```

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -2164,6 +2164,7 @@ mod tests {
             cli_annotate: false,
             cli_download_path: false,
             cli_native: false,
+            cli_headed: false,
             annotate: false,
             color_scheme: None,
             download_path: None,
--- a/cli/src/flags.rs
+++ b/cli/src/flags.rs
@@ -267,6 +267,7 @@ pub struct Flags {
     pub cli_annotate: bool,
     pub cli_download_path: bool,
     pub cli_native: bool,
+    pub cli_headed: bool,
 }
 
 pub fn parse_flags(args: &[String]) -> Flags {
@@ -382,6 +383,7 @@ pub fn parse_flags(args: &[String]) -> Flags {
         cli_annotate: false,
         cli_download_path: false,
         cli_native: false,
+        cli_headed: false,
     };
 
     let mut i = 0;
@@ -404,6 +406,7 @@ pub fn parse_flags(args: &[String]) -> Flags {
             "--headed" => {
                 let (val, consumed) = parse_bool_arg(args, i);
                 flags.headed = val;
+                flags.cli_headed = true;
                 if consumed {
                     i += 1;
                 }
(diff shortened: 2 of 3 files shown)


ctate:
Thanks for the fix @QuietyAwe!

PR #759, opened 2026-03-13
fix: narrow "not found" pattern in to_ai_friendly_error to avoid catching   non-element errors

### LINKED ISSUE #758: Cannot connect to or open browser on Windows (Rust native)
## Environment

I'm on Windows 11, with `agent-browser` version `v0.19.0`. This is my `config.json` (User-level):

```json
{
  "headed": true,
  "native": true
}
```

## Bug

I have opened remote debugging port on Edge at 9222:

Image

And when I try to connect with `agent-browser`:

```shell
$ agent-browser connect 9222
✗ Element not found. Verify the selector is correct and the element exists in the DOM.
$ agent-browser open example.com
✗ Element not found. Verify the selector is correct and the element exists in the DOM.
$ agent-browser --auto-connect open example.com
✗ No running Chrome instance found. Launch Chrome with --remote-debugging-port or use --cdp.
$ agent-browser open example.com --cdp 9222
✗ Invalid /json/version response: EOF while parsing a value at line 1 column 0
$ agent-browser close
✗ Element not found. Verify the selector is correct and the element exists in the DOM.
```

The "Element not found" error messages seem out of place here. Maybe the error originates from the JSON parsing part.

hyunjinee:
## Summary

Narrow `contains("not found")` to `contains("element not found")` in `to_ai_friendly_error()` so that non-element errors like "Browser not found" pass through unchanged instead of being incorrectly mapped to the "Element not found" message

  Closes #758

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -1241,7 +1241,7 @@ mod tests {
     /// Errors containing "not found" but NOT "element" should pass through unchanged.
     #[test]
     fn test_to_ai_friendly_error_ignores_non_element_not_found() {
-        let err = "Browser not found";
+        let err = "Chrome not found. Install Chrome or use --executable-path.";
         assert_eq!(to_ai_friendly_error(err), err);
     }
 


ctate:
Thanks so much @hyunjinee

PR #760, opened 2026-03-13
fix: Lightpanda engine launch with release binaries

seruman:
`--engine lightpanda` has been broken since it was added — it hangs on launch with official Lightpanda release binaries. It only ever worked with debug builds (which have different logging defaults).

Lightpanda release builds default to `log_level=warn`, but we parse stderr looking for the `info`-level "server running" message to detect the listen address. No output → `wait_for_address()` blocks forever.

On top of that, `--timeout 0` was intended as "no timeout" but Lightpanda treats it as "timeout after 0ms" (instant disconnect). And the address parser only matched the pretty/debug log format (`address = 127.0.0.1:9222` with spaces) — release builds use logfmt (`address=127.0.0.1:9222`, no spaces).

Fix:
- Pass `--log_level info` so the startup message always shows up
- `--timeout 604800` (1 week, documented max) instead of `0`
- `extract_address()` handles both logfmt and pretty formats

Tested with the current Lightpanda release binary — open, screenshot, close all work now.

### DIFF
--- a/cli/src/native/cdp/lightpanda.rs
+++ b/cli/src/native/cdp/lightpanda.rs
@@ -110,9 +110,16 @@ pub fn launch_lightpanda(options: &LightpandaLaunchOptions) -> Result<Lightpanda
         args.push(proxy.clone());
     }
 
-    // Disable inactivity timeout so the connection stays alive during long sessions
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the fix @seruman!

PR #761, opened 2026-03-13
fix: correct storage local key lookup parsing and text output

### LINKED ISSUE #554: [Bug] `storage local` silent in text mode; `storage local <key> --json` returns full store instead of requested key
## Summary

Two related bugs in the `storage local` command:

1. **`storage local` in text mode prints `✓ Done` with no data** — localStorage contents are invisible unless `--json` is added.
2. **`storage local <key> --json` returns the entire localStorage object, not the requested key** — per-key reads are broken.

## Environment

- **agent-browser version:** 0.15.1
- **OS:** macOS darwin 25.3.0
- **Install method:** `npm install -g agent-browser`

### Bug 1 — Silent output in text mode

```bash
# Set up some localStorage values
agent-browser --session s1 open https://the-internet.herokuapp.com
agent-browser --session s1 eval "localStorage.setItem('token', 'abc123'); localStorage.setItem('user', 'alice')"

# Read localStorage (text mode, default)
agent-browser --session s1 storage local
```

**Actual output:**
```
✓ Done
```

**Expected output:**
```
token: abc123
user: alice
```

---

### Bug 2 — Per-key read returns full object

```bash
agent-browser --session s1 storage local token --json
```

**Actual output:**
```json
{"token":"abc123","user":"alice"}
```

**Expected output:**
```json
"abc123"
```

## Workaround

```bash
# Bug 1: add --json to see any data at all
agent-browser --session s1 storage local --json
# → returns full object; agent must filter manually

# Bug 2: no clean workaround — must parse full JSON and extract key client-side
```

## Impact

AI agents that issue `storage local <key>` to read a specific value will silently get an empty `✓ Done` response in text mode, or an unexpected full object in JSON mode. Both cases cause silent misreads that are hard to debug in automated workflows.

mikewong23571:
## Problem

`storage local` has two CLI-facing regressions:

1. `agent-browser storage local` falls through to the generic success path in text mode and prints `✓ Done` instead of the storage contents.
2. `agent-browser storage local <key> --json` is parsed incorrectly, so it performs a full-store read instead of a keyed lookup.

The underlying storage handlers already support keyed reads in both the Node and native implementations. The bug is in the Rust CLI parser/output layer, so this change stays scoped to the CLI.

## Fix

- treat an unrecognized second token after `storage <local|session>` as an implicit key lookup
- add dedicated text rendering for `storage_get` responses so full-store and keyed reads print meaningful output
- add regression tests for implicit key parsing and storage text formatting

## Test plan

- [x] `cargo fmt --manifest-path cli/Cargo.toml -- --check`
- [x] `cargo clippy --manifest-path cli/Cargo.toml -- -D warnings`
- [x] `cargo test --profile ci --manifest-path cli/Cargo.toml storage -- --nocapture`

## Notes

- No docs changes needed here: the existing docs already document `agent-browser storage local <key>`, and this PR brings the implementation back in line with that documented behavior.

Fixes #554

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -2084,10 +2084,14 @@ fn parse_storage(rest: &[&str], id: &str) -> Result<Value, ParseError> {
     match rest.first().copied() {
         Some("local") | Some("session") => {
             let storage_type = rest.first().unwrap();
-            let op = rest.get(1).unwrap_or(&"get");
-            let key = rest.get(2);
-            let value = rest.get(3);
-            match *op {
+            let (op, key, value) = match rest.get(1) {
+                Some(&"get") => ("get", rest.get(2), rest.get(3)),
+                Some(&"set") => ("set", rest.get(2), rest.get(3)),
+                Some(&"clear") => ("clear", rest.get(2), rest.get(3)),
+                Some(_) => ("get", rest.get(1), rest.get(2)),
+                None => ("get", None, None),
+            };
+            match op {
                 "set" => {
                     let k = key.ok_or_else(|| ParseError::MissingArguments {
                         context: format!("storage {} set", storage_type),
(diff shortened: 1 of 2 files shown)


ctate:
Clean fix @mikewong23571 - thank you!

PR #762, opened 2026-03-13
fix: harden Lightpanda startup timeouts

### LINKED ISSUE #725: macOS arm64: --engine lightpanda hangs silently even with clean HOME and latest Lightpanda
### Summary
On macOS arm64, `agent-browser --engine lightpanda` hangs silently when opening a page.

The issue reproduces even with a clean `HOME`, after updating Lightpanda to the latest nightly, and even when passing `--native false`.

What works on the same machine:
- default `agent-browser` engine
- standalone `lightpanda`
- `agent-browser` native mode with the default Chrome engine

So this appears to be specific to the `agent-browser` Lightpanda engine path on macOS Apple Silicon.

### Repro
Machine:
- macOS 26.4 (`25E5233c`)
- arm64 / Apple Silicon

Versions:
- `agent-browser 0.17.1`
- `lightpanda version` before update: `6a421a1d`
- `lightpanda version` after update: `487ee183`

Commands that hang:
```bash
agent-browser --engine lightpanda open https://example.com
```

Also hangs:
```bash
agent-browser --engine lightpanda --session lp-test open https://example.com
agent-browser --engine lightpanda --native false --session lp-test open https://example.com
```

Also reproduces with a clean home directory:
```bash
rm -rf /tmp/ab-clean && mkdir -p /tmp/ab-clean
HOME=/tmp/ab-clean agent-browser --engine lightpanda --session cleanlp open https://example.com
```

### Actual
The command produces no output and does not complete.

Using a bounded timeout:
```bash
timeout 20s agent-browser --engine lightpanda --session lp-test open https://example.com
```
returns exit code `124` with no stdout/stderr.

During the hang, `agent-browser` does launch Lightpanda:
```text
/Users/.../agent-browser-darwin-arm64
/Users/.../lightpanda serve --host 127.0.0.1 --port <port> --timeout 0
```

So the native daemon and `lightpanda serve` process come up, but the command still stalls.

### Expected
`agent-browser --engine lightpanda open https://example.com` should complete and return like the default engine does.

### Control Tests
These all worked on the same machine:

Default engine:
```bash
agent-browser --session chrome-control open https://example.com
agent-browser --session chrome-control get title
agent-browser --session chrome-control close
```
Output included:
```text
✓ Example Domain
Example Domain
✓ Browser closed
```

Standalone Lightpanda:
```bash
lightpanda fetch --dump html https://example.com
```
This returned the expected HTML immediately.

Native mode with default engine:
```bash
agent-browser --native --session nativechrome open https://example.com
agent-browser --native --session nativechrome get title
agent-browser --native --session nativechrome close
```
This also worked.

### Notes
A few things that did **not** change the behavior:
- updating Lightpanda to the latest nightly
- clearing `~/.agent-browser/config.json` from `{"native": true}` to `{}`
- using a clean `HOME`
- passing `--native false`

One additional observation: although help text says `--engine lightpanda` implies `--native`, it looks like `--native false` is accepted syntactically but still ends up spawning `lightpanda serve` and hanging.

That may be expected, but mentioning it because it made this harder to reason about from the CLI surface.

mikewong23571:
Fixes #725

## Summary

This PR fixes the native Lightpanda auto-launch path so it no longer hangs silently during startup and no longer relies on a broken `--timeout 0` launch configuration.

The fix is intentionally narrow:
- it only touches the native Lightpanda launcher / initialization path
- it does not change unrelated browser behavior
- it does not change the Node.js path, which is not used for `engine=lightpanda`

## Root cause

This issue turned out to be narrower than "Lightpanda is generally unusable".

After tracing the native startup path and re-testing with both stable and nightly Lightpanda builds on macOS arm64, I confirmed that the problematic behavior was primarily caused by the way `agent-browser` launched and initialized Lightpanda:

1. The native Lightpanda launcher hardcoded `lightpanda serve ... --timeout 0`.
2. In local testing, `--timeout 0` consistently broke early CDP client initialization.
3. The previous startup flow also depended on Lightpanda-specific stderr output as a readiness signal, which is not a reliable contract.

### Why not keep `--timeout 0`

I re-ran the startup matrix locally against a stable Lightpanda release and confirmed:

- `lightpanda serve --timeout 0` fails with `agent-browser`
- `lightpanda serve --timeout 0` also fails with `puppeteer-core`
- the same server works when the timeout is omitted or set to a positive value such as `100` or `1000`

So `0` is not behaving like a safe "disable timeout" value here. In practice, it causes the CDP connection to become unusable.

### Why not simply omit `--timeout`

Omitting `--timeout` falls back to Lightpanda's current upstream default of `10` seconds.

That default is too short for the way `agent-browser` keeps a browser process alive across commands in an interactive session. A short idle period between commands could leave the user talking to a browser process that has already been disconnected by the server.

For that reason, this PR now sets an explicit long positive timeout instead of:
- forcing `0`, which is broken in practice
- or relying on the upstream default `10s`, which is too fragile for a long-lived session

### Why switch to explicit probing instead of log-based detection

Log output is not a stable readiness API.

A launcher should treat the actual CDP discovery endpoint as the source of truth, not whether a particular startup message appeared on stderr. Explicit probing is more robust because it:
- works even when log output changes or is absent
- verifies the endpoint that the client will actually use
- produces deterministic timeout errors with useful context
- avoids coupling startup behavior to Lightpanda log formatting

This PR therefore keeps the explicit readiness probing approach and uses the discovery endpoint as the readiness signal.

## Changes

- stop relying on the broken `--timeout 0` launch behavior
- launch Lightpanda with an explicit long positive session timeout
- keep explicit CDP readiness probing instead of stderr-log-based startup detection
- enforce the Lightpanda Target-domain initialization deadline as a real hard timeout
- preserve actionable timeout / probe error reporting
- add unit coverage for the timeout configuration and deadline enforcement
- keep Lightpanda e2e coverage for both explicit launch and env-driven auto-launch

## Scope

This PR is intentionally limited to the native Lightpanda path:

- `cli/src/native/cdp/lightpanda.rs`
- `cli/src/native/browser.rs`
- `cli/src/native/e2e_tests.rs`

The Node.js path is not changed on purpose. In this codebase, `engine=lightpanda` is a native-only path; the non-native path already rejects Lightpanda usage rather than launching it. This bug is therefore isolated to the native implementation.

## Local verification

I validated this in several layers.

### Reproduction matrix against Lightpanda itself

On macOS arm64, I tested both a stable Lightpanda release and nightly builds.

Observed behavior:

- no explicit `--timeout`: works
- `--timeout 0`: fails
- `--timeout 100`: works
- `--timeout 1000`: works

I confirmed the same pattern with:
- `agent-browser --native`
- `puppeteer-core`

That was important because it showed the `timeout=0` behavior was not specific to `agent-browser`.

### Manual CLI verification

I also verified the native auto-launch path directly with the rebuilt binary:

```bash
cli/target/debug/agent-browser \
  --session lp-verify-newbin \
  --engine lightpanda \
  --native \
  --executable-path /private/tmp/ab-issue-725/lightpanda \
  open https://example.com
```

This now succeeds and opens the page correctly.

### Automated validation

Ran successfully:

```bash
cargo fmt --manifest-path cli/Cargo.toml -- --check
cargo clippy --manifest-path cli/Cargo.toml -- -D warnings
cargo test --manifest-path cli/Cargo.toml lightpanda -- --nocapture
LIGHTPANDA_BIN=/private/tmp/ab-issue-725/lightpanda cargo test --manifest-path cli/Cargo.toml e2e_lightpanda_launch_can_open_page -- --ignored --nocapture
LIGHTPANDA_BIN=/private/tmp/ab-issue-725/lightpanda cargo test --manifest-path cli/Cargo.toml e2e_lightpanda_auto_launch_can_open_page -- --ignored --nocapture
```

## Notes

This PR is meant to solve the concrete issue in #725 only:
- avoid the broken `--timeout 0` launch path
- avoid fragile log-based startup detection
- fail fast and clearly if the Target domain still does not become usable in time

It does not try to redesign broader Lightpanda support or change unrelated startup behavior.

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -1,12 +1,13 @@
 use serde_json::{json, Value};
 use std::collections::HashSet;
+use std::future::Future;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 use tokio::sync::{broadcast, Mutex};
 
-use super::cdp::chrome::{
-    auto_connect_cdp, discover_cdp_url, launch_chrome, ChromeProcess, LaunchOptions,
-};
+use super::cdp::chrome::{auto_connect_cdp, launch_chrome, ChromeProcess, LaunchOptions};
 use super::cdp::client::CdpClient;
+use super::cdp::discovery::discover_cdp_url;
 use super::cdp::lightpanda::{launch_lightpanda, LightpandaLaunchOptions, LightpandaProcess};
 use super::cdp::types::*;
 
@@ -167,6 +168,10 @@ pub struct BrowserManager {
     default_timeout_ms: u64,
 }
 
+const LIGHTPANDA_CDP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const LIGHTPANDA_CDP_CONNECT_POLL_INTERVAL: Duration = Duration::from_millis(100);
+const LIGHTPANDA_TARGET_INIT_TIMEOUT: Duration = Duration::from_secs(10);
+
 impl BrowserManager {
     pub async fn launch(options: LaunchOptions, engine: Option<&str>) -> Result<Self, String> {
         let engine = engine.unwrap_or("chrome");
@@ -205,9 +210,7 @@ impl BrowserManager {
                     proxy: options.proxy.clone(),
                     port: None,
                 };
-                let lp = tokio::task::spawn_blocking(move || launch_lightpanda(&lp_options))
-                    .await
-                    .map_err(|e| format!("Lightpanda launch task failed: {}", e))??;
+                let lp = launch_lightpanda(&lp_options).await?;
                 let url = lp.ws_url.clone();
                 (url, BrowserProcess::Lightpanda(lp))
             }
@@ -220,17 +223,21 @@ impl BrowserManager {
             }
         };
 
-        let client = CdpClient::connect(&ws_url).await?;
-        let mut manager = Self {
-            client,
-            browser_process: Some(process),
-            ws_url,
-            pages: Vec::new(),
-            active_page_index: 0,
-            default_timeout_ms: 25_000,
+        let manager = if engine == "lightpanda" {
+            initialize_lightpanda_manager(ws_url, process).await?
+        } else {
+            let client = CdpClient::connect(&ws_url).await?;
+            let mut manager = Self {
+                client,
+                browser_process: Some(process),
+                ws_url,
+                pages: Vec::new(),
+                active_page_index: 0,
+                default_timeout_ms: 25_000,
+            };
+            manager.discover_and_attach_targets().await?;
+            manager
         };
-
-        manager.discover_and_attach_targets().await?;
 
         let session_id = manager.active_session_id()?.to_string();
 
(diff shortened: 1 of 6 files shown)


mikewong23571:
This PR has been rebased on top of #760.

#760's fixes (adding `--log_level info`, changing `--timeout 0` to `604800`, and extending `extract_address` to handle logfmt format) are fully superseded by this PR. Since we switch to explicit CDP HTTP probing instead of stderr log parsing, we no longer depend on log output at all — so neither `--log_level info` nor the `extract_address` logfmt handling are needed.

mikewong23571:
Noting this in relation to #760, which was just merged:

#760's fix is solid for the immediate issue, but it still relies on parsing Lightpanda's stderr output to detect when the server is ready. That approach is inherently fragile — if Lightpanda changes its log format or log level defaults again in a future release, the startup detection will silently break again.

This PR takes a different angle: instead of parsing logs, it polls the CDP `/json/version` HTTP endpoint directly as the readiness signal. That endpoint is part of the stable DevTools Protocol surface, so it won't be affected by log format changes, log level defaults, or build type (debug vs release).

Happy to close this if the maintainers prefer keeping the simpler patch. Just wanted to flag the distinction in case it's useful.

ctate:
Works perfectly. Thanks @mikewong23571!

PR #775, opened 2026-03-14
fix: accept integer nodeId/childIds in AX tree for Lightpanda compatibility

STRML:
## Summary

- Accept both string and integer values for `nodeId` and `childIds` in `AXNode` deserialization
- Lightpanda returns these as integers while Chrome returns strings; the CDP spec is ambiguous but strict typing breaks Lightpanda compatibility

## Problem

`agent-browser --engine lightpanda` fails on `snapshot` with:
```
Failed to deserialize CDP response for Accessibility.getFullAXTree: invalid type: integer `2`, expected a string
```

Lightpanda serializes `AXNode.nodeId` and `childIds` as JSON integers, while Chrome uses strings. The current serde `Deserialize` on `AXNode` expects `String` for both fields.

## Fix

Add custom deserializers (`string_or_int` and `opt_vec_string_or_int`) that accept both types, converting integers to strings. No behavioral change for Chrome — strings pass through unchanged.

## Note

Filed upstream as lightpanda-io/browser#1822. If Lightpanda fixes their CDP serialization, this patch becomes a no-op (still correct, just unused). But this is currently the only blocker preventing `--engine lightpanda` from working with `snapshot`, which is the core workflow.

## Test

```bash
agent-browser --engine lightpanda open https://example.com
agent-browser snapshot -i
# Before: ✗ Failed to deserialize CDP response...
# After:  - heading "Example Domain" [level=1, ref=e1]
#          - link "Learn more" [ref=e2]
```

### DIFF
--- a/cli/src/native/cdp/types.rs
+++ b/cli/src/native/cdp/types.rs
@@ -1,5 +1,50 @@
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value;
+
+/// Deserialize a value that may be either a string or an integer into a String.
+/// Lightpanda sends numeric nodeIds/childIds in AX tree responses, while Chrome
+/// sends strings. This accepts both.
+fn string_or_int<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: Deserializer<'de>,
+{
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the clean fix @STRML

PR #783, opened 2026-03-14
fix: correct e2e test assertions for diff_snapshot and domain_filter

jin-2-kakaoent:
## Problem

Two e2e tests were failing on CI:

1. `e2e_diff_snapshot` accessed `data["diff"]["identical"]`, but `diff` is a unified diff string, not an object. Indexing a string with `"identical"` returned `Null`.
2. `e2e_domain_filter` set `domain_filter` before launch, which triggered `Fetch.enable` on all requests. During `navigate`, `Fetch.requestPaused` events were never resolved because the event handler only runs at the start of the next `execute_command`, causing a deadlock.

## Fix

1. Access `data["changed"]` directly instead of `data["diff"]["identical"]`
2. Move `domain_filter` setup after launch so `Fetch.enable` is not installed. Domain filtering still works via `check_url` in `handle_navigate`.

## Test plan

- [x] `e2e_diff_snapshot` passes locally
- [x] `e2e_domain_filter` passes locally
- [ ] CI e2e tests pass

### DIFF
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -1226,7 +1226,6 @@ async fn e2e_state_management() {
 #[ignore]
 async fn e2e_domain_filter() {
     let mut state = DaemonState::new();
-    state.domain_filter = Some(super::network::DomainFilter::new("example.com"));
 
     let resp = execute_command(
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the fix!

PR #784, opened 2026-03-14
feat: add linux-musl (Alpine) builds for x64 and arm64

mtzanidakis:
## Summary
- Add `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` targets to the release workflow, built via `cargo-zigbuild`
- Update the JS wrapper (`bin/agent-browser.js`) and `scripts/postinstall.js` to detect musl libc at runtime and select the correct binary
- Update binary verification steps to expect 7 platform binaries (was 5)

## Details

Musl detection uses `ldd --version` output (which includes "musl" on Alpine/musl systems), with a fallback to checking for `/lib/ld-musl-*.so.1`. This allows `npm install agent-browser` to work natively on Alpine Linux and other musl-based distributions.

Binary naming: `agent-browser-linux-musl-x64` / `agent-browser-linux-musl-arm64`

## Test plan
- [ ] Verify `Linux musl x64` and `Linux musl ARM64` build jobs succeed in CI
- [ ] Verify musl binaries are statically linked and run on Alpine
- [ ] Verify existing glibc/macOS/Windows builds are unaffected
- [ ] Test `npm install` on Alpine picks up the musl binary

### DIFF
--- a/bin/agent-browser.js
+++ b/bin/agent-browser.js
@@ -20,17 +20,10 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 function isMusl() {
   if (platform() !== 'linux') return false;
   try {
-    const lddOutput = execSync('ldd --version 2>&1', { encoding: 'utf8' });
-    return lddOutput.toLowerCase().includes('musl');
+    const result = execSync('ldd --version 2>&1 || true', { encoding: 'utf8' });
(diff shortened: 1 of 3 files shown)


ctate:
Thanks for the update - looks great

PR #821, opened 2026-03-15
fix: remove obsolete BrowserManager TypeScript API from README

jin-2-kakaoent:
## Summary
- Replace Lambda example's `BrowserManager` TypeScript API with CLI-based `execSync` handler (the TS `src/` was removed in #754)
- Remove `Programmatic API` section that referenced the non-existent `BrowserManager` class

## Test plan
- [x] Verify `BrowserManager` string no longer appears in README
- [x] Verify `import.*from 'agent-browser'` pattern no longer appears in README
- [x] Confirm WebSocket Protocol section is preserved

### DIFF
--- a/README.md
+++ b/README.md
@@ -792,7 +792,7 @@ See the [environments example](examples/environments/) for a working demo with a
 
 ### Serverless (AWS Lambda)
 
-```javascript
+```typescript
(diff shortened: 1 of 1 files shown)


ctate:
Good catch @jin-2-kakaoent

PR #827, opened 2026-03-15
fix: filter chrome:// internal targets from auto-connect discovery

### LINKED ISSUE #813: auto-connect open falls back to temp browser instead of reusing live Chrome
Version: `agent-browser 0.20.6`
OS: macOS

## Repro
1. Start Chrome with remote debugging enabled so `~/Library/Application Support/Google/Chrome/DevToolsActivePort` exists.
2. Confirm Chrome is listening on `127.0.0.1:9222`.
3. Run `agent-browser --auto-connect open http://localhost:3000/settings`.

## Observed
- The command reports successful navigation.
- Process inspection shows a new `agent-browser-chrome-*` temp browser was spawned instead of reusing the live Chrome instance.
- `AGENT_BROWSER_AUTO_CONNECT=1 agent-browser open http://localhost:3000/settings` behaves the same way.

Also observed:
- `agent-browser connect 9222` fails with `Invalid /json/version response: EOF while parsing a value at line 1 column 0`.
- The Chrome instance still exposes a valid `DevToolsActivePort` file and listens on `127.0.0.1:9222`.
- `curl -I http://127.0.0.1:9222/json/version` returns `404 Not Found` on this setup.

## Workaround
Reading the raw websocket path from `DevToolsActivePort` and passing it directly via `--cdp` works and does not spawn `agent-browser-chrome-*`:

```bash
WS=$(awk 'NR==1{port=$1} NR==2{path=$1} END{print "ws://127.0.0.1:" port path}' "$HOME/Library/Application Support/Google/Chrome/DevToolsActivePort")
agent-browser --cdp "$WS" open http://localhost:3000/settings
```

## Expected
- `--auto-connect` should reuse the live Chrome instance when `DevToolsActivePort` is present, or
- fail loudly if attach/discovery is not possible.

It should not silently fall back to a temp managed browser while presenting success, because that makes auth/session reuse results misleading.

mvanhorn:
## Summary

Filter out Chrome internal targets (`chrome://`, `chrome-extension://`, `devtools://`) during `--auto-connect` target discovery so follow-up commands operate on user-facing pages instead of internal Chrome pages.

## Problem

When using `--auto-connect`, `discover_and_attach_targets()` enumerates all browser targets and selects the first one with `target_type == "page"` and a non-empty URL. Chrome internal pages like `chrome://omnibox-popup.top-chrome/` satisfy both conditions and get selected as the active target.

This causes follow-up commands to return wrong data:
- `agent-browser --auto-connect get url` returns `chrome://omnibox-popup.top-chrome/`
- `agent-browser --auto-connect snapshot -i` returns `origin=about:blank` with no interactive elements

Reported by [@zbeyens](https://github.com/vercel-labs/agent-browser/issues/813#issuecomment-3041889247): "follow-up commands attach to the wrong target... It looks like `--auto-connect` is latching onto Chrome internal targets like `chrome://omnibox-popup.top-chrome/` instead of the actual app tab."

## Changes

- Add `is_internal_chrome_target()` helper that matches `chrome://`, `chrome-extension://`, and `devtools://` URL prefixes
- Add the filter to the target discovery pipeline in `discover_and_attach_targets()`
- If all targets are internal, the existing "create a new tab" fallback handles it correctly
- Add unit test covering both positive and negative cases

## Testing

- `cargo fmt --check` passes
- `cargo clippy -- -D warnings` passes
- All 436 existing tests pass, plus the new `test_is_internal_chrome_target` test

Fixes #813

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -87,6 +87,14 @@ fn validate_lightpanda_options(options: &LaunchOptions) -> Result<(), String> {
     Ok(())
 }
 
+/// Returns true for Chrome internal targets that should not be selected
+/// during auto-connect (e.g. chrome://, chrome-extension://, devtools://).
+fn is_internal_chrome_target(url: &str) -> bool {
+    url.starts_with("chrome://")
+        || url.starts_with("chrome-extension://")
+        || url.starts_with("devtools://")
+}
+
 /// Converts common error messages into AI-friendly, actionable descriptions.
 pub fn to_ai_friendly_error(error: &str) -> String {
     let lower = error.to_lowercase();
@@ -327,7 +335,9 @@ impl BrowserManager {
             .target_infos
             .into_iter()
             .filter(|t| {
-                (t.target_type == "page" || t.target_type == "webview") && !t.url.is_empty()
+                (t.target_type == "page" || t.target_type == "webview")
+                    && !t.url.is_empty()
+                    && !is_internal_chrome_target(&t.url)
             })
             .collect();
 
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the clean fix @mvanhorn!

PR #834, opened 2026-03-15
feat: implement frameLocator support for cross-origin iframe interaction

### LINKED ISSUE #279: Feature Request: Add frameLocator support for cross-origin iframe interaction

giulio-leone:
## Summary

Implements the `frameLocator` feature request from #279, enabling cross-origin iframe interaction (Apple auth, OAuth providers, embedded forms, etc.) via CDP execution context scoping.

Supersedes #460 (which targeted the old TypeScript codebase).

### Core: Frame-scoped evaluation
- Added `context_id: Option<i64>` to `EvaluateParams` for frame-scoped `Runtime.evaluate`
- Added `BrowserManager::get_frame_context_id()` using `Page.createIsolatedWorld` with `grantUniversalAccess: true`
- Added `BrowserManager::evaluate_in_context()` for optional frame-scoped JS evaluation

### New command: `frame locator`
- `frame locator <css-selector>` — scope all subsequent commands to the matched iframe
- `frame locator clear` — return to main frame context
- Equivalent to Playwright's `page.frameLocator(selector)`

### Frame-aware command execution
- `evaluate` — JS evaluation scoped to active frame
- `getByRole`, `getByText`, `getByLabel`, `getByPlaceholder`, `getByTestId`, `getByTitle`, `getByAltText` — semantic locators resolve elements within iframe DOM
- `click`, `fill`, `check`, `hover`, `text` subactions — coordinate resolution accounts for iframe viewport offset
- Attribute cleanup (`data-agent-browser-located`) runs in the correct frame context

### Improved frame resolution
- Multi-strategy matching: name → id → src URL → nth child frame
- Handles anonymous iframes without `name` or `id` attributes
- Extracted `find_frame_in_tree()` and `find_nth_child_frame()` as reusable helpers

## How it works

When `frame locator <selector>` is invoked:
1. The CSS selector resolves the iframe element in the main page
2. The iframe's `name`, `id`, and `src` are extracted
3. The frame tree (`Page.getFrameTree`) is searched using multiple strategies
4. `Page.createIsolatedWorld` creates an execution context in the target frame
5. All subsequent `getBy*`, `evaluate`, and interaction commands use this context
6. `frame locator clear` resets to the main frame

This uses CDP's built-in cross-origin access (`grantUniversalAccess`), bypassing same-origin restrictions at the browser level — no proxy or content injection needed.

## Testing

```bash
# Navigate to a page with an iframe
> navigate https://example.com/with-oauth-iframe

# Scope to the iframe
> frame locator "#auth-iframe"

# Interact with elements inside the iframe
> getByRole button --name "Sign in"
> click

# Return to main frame
> frame locator clear
```

Closes #279

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -965,12 +965,8 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
                 Ok(json!({ "id": id, "action": "mainframe" }))
             } else if rest.first().copied() == Some("locator") {
                 match rest.get(1).copied() {
-                    None | Some("clear") => {
-                        Ok(json!({ "id": id, "action": "framelocator" }))
-                    }
-                    Some(sel) => {
-                        Ok(json!({ "id": id, "action": "framelocator", "selector": sel }))
-                    }
+                    None | Some("clear") => Ok(json!({ "id": id, "action": "framelocator" })),
+                    Some(sel) => Ok(json!({ "id": id, "action": "framelocator", "selector": sel })),
                 }
             } else {
                 let sel = rest.first().ok_or_else(|| ParseError::MissingArguments {
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -3588,13 +3588,9 @@ async fn handle_frame(cmd: &Value, state: &mut DaemonState) -> Result<Value, Str
         let label = name.or(url).unwrap_or("frame");
         state.active_frame_id = Some(frame_id);
         // Build a CSS selector that targets the specific iframe element
-        state.active_frame_selector = if let Some(n) = name {
-            Some(format!("iframe[name=\"{}\"], frame[name=\"{}\"]", n, n))
-        } else if let Some(u) = url {
-            Some(format!("iframe[src=\"{}\"], frame[src=\"{}\"]", u, u))
-        } else {
-            None
-        };
+        state.active_frame_selector = name
+            .map(|n| format!("iframe[name=\"{}\"], frame[name=\"{}\"]", n, n))
+            .or_else(|| url.map(|u| format!("iframe[src=\"{}\"], frame[src=\"{}\"]", u, u)));
         return Ok(json!({ "frame": label }));
     }
 
--- a/cli/src/native/element.rs
+++ b/cli/src/native/element.rs
@@ -246,8 +246,8 @@ pub async fn resolve_element_object_id(
                 expression: js,
                 return_by_value: Some(false),
                 await_promise: Some(false),
-                    context_id: None,
-                },
+                context_id: None,
+            },
             Some(session_id),
         )
         .await?;
@@ -339,8 +339,8 @@ async fn resolve_by_selector(
                 expression: js,
                 return_by_value: Some(true),
                 await_promise: Some(false),
-                    context_id: None,
-                },
+                context_id: None,
+            },
             Some(session_id),
         )
         .await?;
@@ -689,8 +689,8 @@ pub async fn get_element_count(
                 expression: js,
                 return_by_value: Some(true),
                 await_promise: Some(false),
-                    context_id: None,
-                },
+                context_id: None,
+            },
             Some(session_id),
         )
         .await?;
--- a/cli/src/native/snapshot.rs
+++ b/cli/src/native/snapshot.rs
@@ -152,8 +152,8 @@ pub async fn take_snapshot(
                         expression: js,
                         return_by_value: Some(false),
                         await_promise: Some(false),
-                    context_id: None,
-                },
+                        context_id: None,
+                    },
                     Some(session_id),
                 )
                 .await?;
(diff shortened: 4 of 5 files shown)


giulio-leone:
Thanks for the thorough review @ctate! All three issues addressed in `39e2781`:

1. **`cargo fmt`** — Applied across the codebase ✅
2. **`resolve_element_in_frame` iframe targeting** — Added `active_frame_selector` field to `DaemonState` that stores the CSS selector (or equivalent attribute selector for name/url lookups) used when entering a frame. `resolve_element_in_frame` now uses this stored selector to target the specific iframe instead of grabbing the first visible one. Falls back to first-visible only when no selector is available. ✅
3. **`find_frame_in_tree` id vs name** — Reordered matching priority in `handle_frame` to `name → src → id-fallback → nth-child`. Since CDP's `name` field maps to `iframe.name` (not `iframe.id`), iframes with only an `id` attribute now correctly match via their `src` URL first, with id-as-name as a last-resort fallback. ✅

giulio-leone:
@ctate Both fixes applied in `fe4b79d`:

1. **`cargo fmt`** — ran across all 5 files (commands.rs, element.rs, snapshot.rs, storage.rs, actions.rs)
2. **`clippy::manual_map`** — refactored the `active_frame_selector` assignment at actions.rs:3590 from manual `if let / else if let / else` to idiomatic `Option::map().or_else()`

Clean build + zero clippy warnings confirmed locally. Ready for re-review!

giulio-leone:
Hi @ctate — all requested changes have been addressed:
- ✅ `active_frame_selector` uses CSS attribute selector
- ✅ Frame matching prioritizes name/id before URL
- ✅ `cargo fmt` applied
- ✅ Clippy `manual_map` lint fixed

Would you be able to take another look when you get a chance? Thanks! 🙏

ctate:
Hey, thanks for the work on this — I appreciate you tackling #279 and the CDP approach with `Page.createIsolatedWorld` was a reasonable direction.

Unfortunately, after reviewing more closely I found there were enough issues that I ended up reworking frame support from scratch in #869. The main things I ran into:

- `snapshot` and `screenshot` weren't frame-aware, so an agent couldn't actually *see* inside the iframe — which breaks the observe-then-act loop that agents depend on
- `Page.createIsolatedWorld` was being called on every evaluation instead of caching the context
- `fill` used `.value =` assignment instead of CDP keyboard dispatch, which doesn't work with React/Vue controlled inputs
- Frame state wasn't cleared on browser close or navigation
- The existing `frame` command was setting `active_frame_id` but nothing was reading it

The approach in #869 threads frame context through the entire pipeline — `RefEntry` now tracks which frame each element belongs to, `snapshot` is frame-aware, and element resolution uses the frame context from ref entries. This means the standard observe-then-act flow just works:

```bash
> frame @e2          # scope into an iframe ref from snapshot
> snapshot -i        # now shows iframe contents
> click @e3          # interacts within the iframe
> frame main         # back to main frame
```

Sorry I didn't flag these gaps earlier. The core idea was right, I just needed to take it further than review feedback could cover. Thanks again for pushing this forward.

giulio-leone:
Thanks for the detailed follow-up and for carrying iframe support through to a complete implementation in `#869`.

I reviewed your supersession note on this PR plus the merged `#869`, and I agree this branch is no longer the right merge path. The merged rewrite covers the gaps you called out more completely: frame-aware observe-then-act flow, frame-aware snapshots, ref-threaded frame context, and the broader end-to-end handling that this earlier approach did not fully cover.

To keep the queue clean on my side, I'm closing `#834` as superseded by `#869`.

Thanks again for the review and for taking the feature the rest of the way.

PR #840, opened 2026-03-15
fix: Windows auto-connect profiling (#835)

### LINKED ISSUE #835: Bug[Windows]: --auto-connect fails to attach to existing Chrome instance profile and opens a new one
Observed issues: 
1. When using I'm using `--auto-connect`, agent-browser does not connect to an existing Chrome instance that has user profile data, but instead opens a brand new Chrome window.
2. I had to manually write `agent-browser close` each times to get a clean session. otherwise it is throwing warnings `daemon already running. Use 'agent-browser close' first to restart with new options.`

```
C:\Users\ayush>agent-browser --version
agent-browser 0.20.10

C:\Users\ayush>agent-browser --auto-connect open google.com --headed
✗ CDP WebSocket connect failed: IO error: No connection could be made because the target machine actively refused it. (os error 10061)

C:\Users\ayush>agent-browser --auto-connect open google.com --headed
⚠ --headed ignored: daemon already running. Use 'agent-browser close' first to restart with new options.
✓ Google
  https://www.google.com/

C:\Users\ayush>agent-browser close
✓ Browser closed

C:\Users\ayush>agent-browser --auto-connect open google.com --headed
✓ Google
  https://www.google.com/

C:\Users\ayush>
```
---
Environment:

Windows 11
agent-browser: v0.20.10
Chrome: 146.0.7680.80 (Official Build) (64-bit)

---
https://github.com/user-attachments/assets/e321fed8-5d4a-45fc-b24f-e668166d80e2

AyushRajgor:
Closes #835

## Problem

On Windows, `--auto-connect` fails to attach to an existing Chrome instance and instead opens a brand new Chrome window with a temporary profile — losing the user's cookies, history, and extensions. The user also hits `os error 10061` (WSAECONNREFUSED) on first run and has to repeatedly run `agent-browser close` to get a clean session.

**Root cause:** Four interrelated bugs combine to produce this behavior.

### Bug 1: Stale `DevToolsActivePort` file — no port validation

`auto_connect_cdp()` in `chrome.rs` reads the `DevToolsActivePort` file from Chrome user data directories. On Windows, Chrome can leave a stale file after exit or crash. The function returned a WebSocket URL pointing to a dead port without checking if it was actually listening, causing `CdpClient::connect()` to fail with `os error 10061` (WSAECONNREFUSED).

### Bug 2: Missing Windows error codes in `is_transient_error()`

`is_transient_error()` in `connection.rs` handled macOS (`os error 61`) and Linux (`os error 111`) connection-refused codes but **not Windows** (`os error 10061`). Similarly, `os error 10054` (WSAECONNRESET) was missing. This meant CLI-to-daemon transient connection errors on Windows were never retried, causing commands to fail immediately during daemon startup races.

### Bug 3: `--auto-connect` not propagated to daemon environment

`DaemonOptions` had no `auto_connect` field, and `apply_daemon_env()` never set `AGENT_BROWSER_AUTO_CONNECT`. When a follow-up command triggered `auto_launch()` in the daemon, the env var was missing, so it fell through to `BrowserManager::launch()` — opening a brand new Chrome with a temp profile instead of connecting to the user's existing browser.

### Bug 4: `--auto-connect --headed` sent two conflicting launch commands

When both `--auto-connect` and `--headed` were used, the CLI sent **two** launch commands to the daemon:
1. `{"action": "launch", "autoConnect": true}` — correctly connected to the user's Chrome
2. `{"action": "launch", "headless": false}` — fired because the headed launch block only guarded against `--cdp` and `--provider`, not `--auto-connect`

The daemon saw a type mismatch (external CDP connection vs. local launch), **disconnected** from the user's Chrome, and launched a brand new one.

### Bug 5: Auto-connect hijacks existing tabs

After connecting via CDP, `discover_and_attach_targets()` attached to the first existing tab. Subsequent `navigate` commands overwrote that tab's content instead of opening a new tab — destroying whatever the user had open.

### `cli/src/native/cdp/chrome.rs`
- Add `is_port_reachable()` — TCP connect probe with 500ms timeout
- Before returning the M144+ WebSocket URL, verify the port is actually listening
- If the port is dead, delete the stale `DevToolsActivePort` file and fall through

### `cli/src/connection.rs`
- Add `os error 10061` (WSAECONNREFUSED) and `os error 10054` (WSAECONNRESET) to `is_transient_error()`
- Add `auto_connect: bool` field to `DaemonOptions`
- Set `AGENT_BROWSER_AUTO_CONNECT=1` in `apply_daemon_env()` when the field is true

### `cli/src/main.rs`
- Pass `flags.auto_connect` into `DaemonOptions`
- Add `&& !flags.auto_connect` guard to the headed launch block so it doesn't send a second conflicting launch command

### `cli/src/native/actions.rs`
- In both `handle_launch()` (explicit) and `auto_launch()` (implicit), after auto-connecting create a new `about:blank` tab and bring it to front so navigations don't hijack the user's existing tabs

## Test plan

- [x] `cargo check` — compiles without errors
- [x] `agent-browser --auto-connect open google.com --headed` with Chrome remote debugging enabled — connects to existing Chrome profile, opens a new tab, and navigates there
- [x] Second run without `--auto-connect` reuses the same Chrome session (Bug 3 env propagation)
- [x] First run with no Chrome debugging returns a clean "No running Chrome instance found" error instead of `os error 10061` (Bug 1 stale port)
- [x] `agent-browser open google.com --headed` retries transient errors during daemon startup on Windows (Bug 2)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # agent-browser
 
+## 0.20.13
+
+### Patch Changes
+
+- eda956b: ### Bug Fixes
+
+  - **Network idle detection for cached pages** - Fixed an issue where `poll_network_idle` could return immediately when no network events were observed (e.g. pages served from cache). The idle timer is now only satisfied after a consistent **500 ms idle period** has elapsed, preventing false-positive idle detection. The core polling logic has also been extracted into a standalone `poll_network_idle` function to improve testability (#847)
+
+## 0.20.12
+
+### Patch Changes
+
+- 5fa2396: ### Bug Fixes
+
+  - Fixed **`snapshot -C`** and **`screenshot --annotate`** hanging when connected over WSS (WebSocket Secure) due to sequential CDP round-trips per interactive element (#842)
+
+  ### Performance
+
+  - **`snapshot -C` (cursor-interactive mode)** now batches CDP calls instead of issuing N×2 sequential round-trips per cursor-interactive element, preventing timeouts on high-latency WSS connections (#842)
+  - **`screenshot --annotate`** now batches element queries, reducing completion time from potentially 20–40s (e.g. 50+ buttons over WSS) to within expected bounds (#842)
+
+## 0.20.11
+
+### Patch Changes
+
+- 4b5fc78: ### Bug Fixes
+
+  - **Material Design checkbox/radio parity** - Restored Playwright-parity behavior for `check`/`uncheck` actions on Material Design controls. These components hide the native `<input>` off-screen and use overlay elements that intercept coordinate-based clicks; the actions now detect this pattern and fall back to a JS `.click()` to correctly toggle state. Also improves `ischecked` to handle nested hidden inputs and ARIA-only checkboxes (#837)
+  - **Punctuation handling in `type` command** - Fixed incorrect virtual key (VK) codes being used for punctuation characters (e.g. `.`, `@`) in the `type` action, which previously caused those characters to be dropped or mistyped (#836)
+
 ## 0.20.10
 
 ### Patch Changes
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -45,7 +45,7 @@ dependencies = [
 
 [[package]]
 name = "agent-browser"
-version = "0.20.10"
+version = "0.20.13"
 dependencies = [
  "aes-gcm",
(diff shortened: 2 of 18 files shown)


ctate:
Thank you for the clean fix @AyushRajgor

PR #852, opened 2026-03-16
Enhance Chrome launch process with user-data-dir and timeout

jianjianai:
image

# Pull Request: Fix flaky Chrome auto-launch on Windows (DevTools URL discovery)

## Summary
This PR fixes a flaky Chrome auto-launch issue where `agent-browser open ...` could fail with:

> `Auto-launch failed: Chrome exited before providing DevTools URL (no stderr output from Chrome)`

On Windows, Chrome does not reliably emit `DevTools listening on ...` to **stderr**. The previous launch flow depended on scraping stderr for that line, which caused intermittent failures and a user-visible workaround: running `close` before `open`.

### ✅ Prefer `DevToolsActivePort` over stderr scraping
We now treat `DevToolsActivePort` in the selected `--user-data-dir` as the primary source of truth for the DevTools WebSocket URL.

- Chrome is launched with `--remote-debugging-port=0` and a known `--user-data-dir`.
- Chrome writes `DevToolsActivePort` into that directory once the debug endpoint is ready.
- We poll that file (up to 30s) and build the WebSocket URL as:

`ws://127.0.0.1:<port><ws_path>`

### 🔁 Keep stderr parsing as a fallback
If `DevToolsActivePort` is not observed within the timeout, we fall back to the legacy behavior of parsing stderr for:

`DevTools listening on ...`

This preserves compatibility and still provides helpful diagnostics.

## Files changed
- `cli/src/native/cdp/chrome.rs`
  - Added `wait_for_devtools_active_port(...)`
  - Updated `try_launch_chrome(...)` to:
    1. Poll `DevToolsActivePort`
    2. Fallback to stderr scraping

## Why this fixes the “must close first” behavior
The failure mode was not that Chrome *must* be closed first, but that the tool’s launch handshake relied on stderr output that is sometimes missing. Using `DevToolsActivePort` makes readiness detection deterministic and independent of Chrome’s stderr behavior.

## Testing
- Built successfully: `cargo build` (Windows)
- Manual verification:
  - Repeated runs of `agent-browser open <url>` should no longer require an explicit `agent-browser close` beforehand.

## Notes / Follow-ups
- If desired, we can further improve error reporting by including the chosen `--user-data-dir` path and whether `DevToolsActivePort` ever appeared (helpful for diagnosing profile permission/lock issues).

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -346,31 +346,6 @@ fn wait_for_ws_url_until(
     ))
 }
 
-fn wait_for_ws_url(reader: BufReader<std::process::ChildStderr>) -> Result<String, String> {
-    let deadline = std::time::Instant::now() + Duration::from_secs(30);
-    let prefix = "DevTools listening on ";
-    let mut stderr_lines: Vec<String> = Vec::new();
-
-    for line in reader.lines() {
-        if std::time::Instant::now() > deadline {
-            return Err(chrome_launch_error(
-                "Timeout waiting for Chrome DevTools URL",
-                &stderr_lines,
-            ));
-        }
-        let line = line.map_err(|e| format!("Failed to read Chrome stderr: {}", e))?;
-        if let Some(url) = line.strip_prefix(prefix) {
-            return Ok(url.trim().to_string());
-        }
-        stderr_lines.push(line);
-    }
-
-    Err(chrome_launch_error(
-        "Chrome exited before providing DevTools URL",
-        &stderr_lines,
-    ))
-}
-
 fn chrome_launch_error(message: &str, stderr_lines: &[String]) -> String {
     let relevant: Vec<&String> = stderr_lines
         .iter()


ctate:
Thanks for the contribution! The approach of preferring `DevToolsActivePort` over stderr scraping with a fallback is solid and well-motivated. A few items to address:

- **`cargo fmt` needs to be run** — the `wait_for_devtools_active_port` function signature at line 295 exceeds the line width limit and will fail CI
- **Unused `_user_data_dir` in `build_chrome_args`** — the return tuple was changed to include `_user_data_dir` but it's never used; `try_launch_chrome` re-extracts the dir by scanning the args vec. Either plumb it through `ChromeArgs` and use it directly, or revert the `build_chrome_args` changes
- **Stale `DevToolsActivePort` risk** — if Chrome previously crashed without cleanup, a stale file could cause false-positive detection. Consider deleting it before spawning Chrome (as Puppeteer does)
- **Double timeout** — if `DevToolsActivePort` polling times out (30s), the stderr fallback starts its own 30s deadline, making worst-case failure ~60s. Consider sharing a single deadline or shortening the fallback

Happy to re-review once those are addressed!

jianjianai:
## Update (follow-up to review feedback)

- **Formatting:** Ran `cargo fmt` to ensure CI passes (including wrapping long function signatures).
- **No unused `user_data_dir`:** Plumbed `user_data_dir` through `ChromeArgs` and used it directly in `try_launch_chrome` (no longer re-extracting it by scanning the args vector).
- **Stale `DevToolsActivePort` mitigation:** Removed `<user-data-dir>/DevToolsActivePort` before spawning Chrome to avoid false-positive readiness detection after prior crashes.
- **Single timeout budget:** Switched polling and stderr fallback to share a single overall deadline, avoiding worst-case ~60s failures.

ctate:
Thanks @jianjianai - One change requested: the original wait_for_ws_url() function (lines 349–372) is now dead code — try_launch_chrome exclusively calls the new wait_for_ws_url_until(). Could you plz remove it?

jianjianai:
I have removed the unused wait_for_ws_url() function.

ctate:
Great - thanks @jianjianai!

PR #857, opened 2026-03-16
fix: propagate --cdp flag to daemon for reliable CDP reconnection

BusyBee3333:
## Summary

This PR fixes a bug where the `--cdp` flag value was not propagated to the daemon process, causing auto-reconnection to fail after CDP connection drops.

## The Bug

When using `--cdp PORT` to connect to external browsers (Lightpanda, Electron, remote Chrome, etc.), the first connection works correctly. However, if the CDP connection drops and the daemon attempts to auto-reconnect via `auto_launch()`, it fails because:

1. The daemon's `auto_launch()` function checks for `AGENT_BROWSER_CDP" environment variable
2. But `apply_daemon_env()` never sets this env var when spawning the daemon
3. The `--cdp` value was missing from `DaemonOptions` struct

This meant auto-reconnection would try to launch a local Chrome browser instead of reconnecting to the CDP endpoint.

## The Fix

This PR adds three small changes:

1. **cli/src/connection.rs**: Add `cdp: Option<&'a str>` field to `DaemonOptions` struct
2. **cli/src/connection.rs**: Set `AGENT_BROWSER_CDP` env var in `apply_daemon_env()`  
3. **cli/src/main.rs**: Pass `flags.cdp.as_deref()` to daemon options

## Testing

- ✅ All existing tests pass (`cargo test` - 60 relevant tests)
- ✅ Code compiles cleanly (`cargo check`)
- ✅ Verified `--cdp` flag works with Lightpanda

## Impact

This ensures reliable CDP connection recovery when using `--cdp` with:
- Lightpanda (Zig-based headless browser)
- Electron applications  
- Remote Chrome instances
- Any browser exposing a CDP endpoint

Previously, connection drops would cause `--cdp` sessions to fail. Now they recover automatically.

## Changeset

Added patch-level changeset in `.changeset/fix-cdp-env-propagation.md`

### DIFF
--- a/.changeset/fix-cdp-env-propagation.md
+++ b/.changeset/fix-cdp-env-propagation.md
@@ -1,9 +1,0 @@
----
-"agent-browser": patch
----
-
-fix: propagate --cdp flag to daemon for reliable CDP reconnection
-
-When using the --cdp flag, the CDP endpoint was not being passed to the daemon process via environment variables. This caused auto-reconnection to fail after connection drops, as auto_launch() checks AGENT_BROWSER_CDP but it was never set.
-
-This fix ensures the --cdp value is propagated to the daemon, enabling reliable CDP connection recovery.


ctate:
Thank you so much @BusyBee3333

PR #862, opened 2026-03-17
feat: add --ignore-default-args to exclude Chrome launch flags

### LINKED ISSUE #682: Feature request: expose ignoreDefaultArgs option to allow Brave Shields to work
## Problem

When using agent-browser with Brave as the custom browser (`executablePath`), Brave's built-in ad/tracker blocker (Shields) does not work. The Shields icon shows as active but blocks 0 requests.

The root cause is a single Playwright default launch flag: `--disable-component-update`. This flag prevents Brave from downloading its ad-block filter lists, which Shields needs to function. On a fresh or incognito profile, there are no cached filter lists, so Shields has nothing to block with.

This cannot be worked around by adding `--enable-component-update` via the `args` config -- Chromium does not treat these as toggles, and the disable flag always wins when both are present.

## Proposed Solution

Expose Playwright's `ignoreDefaultArgs` option in agent-browser's launch config. This option accepts an array of specific default arguments to exclude.

Example config for Brave with working Shields:
```json
{
  "executablePath": "/path/to/brave",
  "headed": true,
  "ignoreDefaultArgs": ["--disable-component-update"]
}
```

mvanhorn:
## Summary

Adds `--ignore-default-args` to let users exclude specific default Chrome launch flags. Accepts a comma-separated list of flags to remove.

## Why this matters

Chrome's `--disable-component-update` flag prevents Brave from downloading its Shields filter lists ([#682](https://github.com/vercel-labs/agent-browser/issues/682)). Users with custom Chromium-based browsers need a way to opt out of specific defaults without replacing the entire args list.

## Changes

- `flags.rs`: Add `--ignore-default-args` CLI flag and `AGENT_BROWSER_IGNORE_DEFAULT_ARGS` env var
- `chrome.rs`: Filter default args through the exclusion set before launch, matching on the flag name before any `=` value
- `actions.rs`: Plumb `ignore_default_args` through both `LaunchOptions` construction sites
- `commands.rs`: Add field to test helper

## Usage

```bash
# CLI
agent-browser --ignore-default-args=--disable-component-update open example.com

# Environment variable
AGENT_BROWSER_IGNORE_DEFAULT_ARGS="--disable-component-update" agent-browser open example.com

# Config (agent-browser.json)
{ "ignoreDefaultArgs": "--disable-component-update" }
```

## Testing

- `cargo test -- chrome` passes all 23 tests
- `cargo fmt --check` and `cargo clippy` clean

Fixes #682

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -3235,6 +3235,7 @@ mod tests {
             plugins: Vec::new(),
             verbose: false,
             quiet: false,
+            ignore_default_args: None,
         }
     }
 
--- a/cli/src/flags.rs
+++ b/cli/src/flags.rs
@@ -102,6 +102,7 @@ pub struct Config {
     pub no_auto_dialog: Option<bool>,
     pub model: Option<String>,
     pub plugins: Option<Vec<PluginConfig>>,
+    pub ignore_default_args: Option<String>,
 }
 
 impl Config {
@@ -179,6 +180,7 @@ impl Config {
                 }
                 (a, b) => b.or(a),
             },
+            ignore_default_args: other.ignore_default_args.or(self.ignore_default_args),
         }
     }
 }
@@ -278,6 +280,7 @@ fn extract_config_path(args: &[String]) -> Option<Option<String>> {
         "--screenshot-format",
         "--idle-timeout",
         "--model",
+        "--ignore-default-args",
     ];
     let mut i = 0;
     while i < args.len() {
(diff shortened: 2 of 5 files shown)


ctate:
Thanks for the PR! The filtering logic in `build_chrome_args()` looks good and the env var path works. I did find two issues that prevent the CLI `--ignore-default-args` flag from actually working: it's missing from `GLOBAL_FLAGS_WITH_VALUE` in `clean_args()` (causes "Unknown command" errors), and it's never added to the launch JSON in `main.rs`. Would also be great to add a unit test for the new filtering behavior.

mvanhorn:
Fixed in c16913f: added `--ignore-default-args` to `GLOBAL_FLAGS_WITH_VALUE` so `clean_args()` strips it before `parse_command`, and wired the value into the launch JSON so the daemon's `handle_launch` receives it. Also added 5 unit tests covering comma-separated parsing, single values, absent flag, and `clean_args` stripping behavior.

PR #872, opened 2026-03-17
Fix raw native mouse drag state across down/move/up

danielvershinin-pixieset:
## Summary

  Fix native raw mouse events so `mouse down` / `mouse move` / `mouse up` preserve the current pointer coordinates and
  button state across a drag sequence.

  ## What changed

  - add persistent mouse state to the native daemon state
  - build typed CDP `Input.dispatchMouseEvent` params from tracked position and button state
  - update raw native `mousemove`, `mousedown`, `mouseup`, and `input_mouse` to reuse that state instead of falling back
  to `(0, 0)` / empty button state
  - reset tracked mouse state on close, relaunch, and new browser session setup
  - add unit coverage for mouse event param construction and input-state reset
  - add native e2e coverage for:
    - drag-state preservation across raw mouse down/move/up
    - pointer-capture drag behavior
  - move the probe HTML into checked-in native test fixtures

  ## Why

  Before this change, low-level native mouse sequences could dispatch `mousedown` / `mouseup` at stale default
  coordinates and lose button state between events. That breaks realistic drag flows and pointer-capture interactions
  driven through the raw mouse commands.

  ## Verification

  ```bash
  cargo test --manifest-path cli/Cargo.toml native::actions::tests::
  cargo test --manifest-path cli/Cargo.toml e2e_mouse_ -- --ignored --test-threads=1

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -5202,6 +5202,7 @@ fn primary_button_from_mask(buttons: i32) -> &'static str {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 fn build_mouse_event_params(
     mouse_state: &mut MouseState,
     event_type: &str,


danielvershinin-pixieset:
> Thanks for the contribution — the fix is solid and the test coverage is thorough! One thing to address: CI runs `cargo clippy -- -D warnings`, and `build_mouse_event_params` will fail with `too_many_arguments` (10 params, limit 7). Adding `#[allow(clippy::too_many_arguments)]` above the function should be a quick fix.

Thanks! Updated per your instructions just now.

ctate:
Looks great - thanks @danielvershinin-pixieset

PR #873, opened 2026-03-17
feat: fall back to WebSocket when HTTP discovery fails (fixes #870)

### LINKED ISSUE #870: `--cdp` / `--auto-connect` should fall back to ws://host/devtools/browser when /json/version returns 404
## Summary

Chrome 136+ silently ignores `--remote-debugging-port` on real user profiles ([blog post](https://developer.chrome.com/blog/remote-debugging-port)). The new way to debug a real Chrome profile is via `chrome://inspect/#remote-debugging`, which opens a CDP server on port 9222.

This server serves **standard CDP over WebSocket** but does **not** expose the HTTP discovery endpoints (`/json/version`, `/json/list`). agent-browser's `--cdp` and `--auto-connect` both require `/json/version` and fail immediately.

**The fix is a one-line fallback:** when `/json/version` returns 404, try connecting directly to `ws://{host}:{port}/devtools/browser`. The WebSocket endpoint works perfectly and returns standard CDP responses.

## Proof that it works today (with manual WebSocket URL)

```bash
# Enable debugging: chrome://inspect/#remote-debugging -> Enable
# Chrome reports: "Server running at: 127.0.0.1:9222"

# This fails (HTTP discovery endpoints return 404):
agent-browser --cdp 9222 state save ./auth.json
# ✗ Invalid /json/version response: EOF while parsing a value

# This works perfectly (direct WebSocket):
agent-browser --cdp "ws://127.0.0.1:9222/devtools/browser" snapshot -i
# ✓ Shows full page with authenticated content

agent-browser --cdp "ws://127.0.0.1:9222/devtools/browser" state save ./auth.json
# ✓ State saved (13 cookies exported from real Chrome profile)
```

## What the server looks like

```bash
# HTTP endpoints all return 404:
curl http://127.0.0.1:9222/json/version   # 404
curl http://127.0.0.1:9222/json/list      # 404
curl http://127.0.0.1:9222/               # 404

# WebSocket on / is 403 (origin check):
# "Rejected an incoming WebSocket connection... Use --remote-allow-origins=* to allow"

# WebSocket on /devtools/browser works - standard CDP:
python3 -c "
import asyncio, json, websockets
async def test():
    async with websockets.connect('ws://127.0.0.1:9222/devtools/browser') as ws:
        await ws.send(json.dumps({'id': 1, 'method': 'Browser.getVersion'}))
        print(await ws.recv())
asyncio.run(test())
"
# {"id":1,"result":{"protocolVersion":"1.3","product":"Chrome/146.0.7680.80",...}}
```

## Suggested change

In the CDP connection logic, when `/json/version` returns a non-200 status or empty body:

1. Try `ws://{host}:{port}/devtools/browser` directly
2. If that WebSocket connects and responds to `Browser.getVersion`, use it
3. This also fixes `--auto-connect` if it detects port 9222 is open but `/json/version` fails

## Why this matters

Since Chrome 136 (April 2025), the "Import auth from your browser" workflow in the README is broken for everyone. `--remote-debugging-port` on a real profile is silently ignored, and `chrome://inspect` debugging (the replacement) uses the pattern above. This fallback would restore the primary auth import use case.

## Environment

- agent-browser: 0.20.14
- Chrome: 146.0.7680.80
- macOS: Darwin 25.1.0 (Sequoia)

## Related

- #42 (Support --user-data-dir to use existing Chrome profiles)
- #77 (CDP Support for --user-data-dir parameter)

jin-2-kakaoent:
## Summary

- Chrome 136+ with `chrome://inspect/#remote-debugging` exposes CDP over WebSocket but returns 404 for `/json/version` and `/json/list` HTTP endpoints
- Add a third fallback in `discover_cdp_url_with_timeout()` that connects directly to `ws://host:port/devtools/browser` and verifies the endpoint with `Browser.getVersion`
- This fixes both `--cdp <port>` and `--auto-connect` for Chrome's inspect-based debugging

## Test plan

- [x] Reproduced bug on main: `--cdp 9222 eval "1+1"` → `✗ Invalid /json/version response: EOF while parsing a value`
- [x] Verified fix: same command on fix branch → `2` (success)
- [x] All tests pass (`cargo test cdp::discovery` — 6/6)
- [x] `cargo clippy` clean, `cargo fmt` clean
- [x] Manual test with `chrome://inspect/#remote-debugging` active on Chrome 145

Fixes #870

### DIFF
--- a/cli/src/native/cdp/discovery.rs
+++ b/cli/src/native/cdp/discovery.rs
@@ -132,7 +132,6 @@ async fn fetch_cdp_list(host: &str, port: u16, timeout: Duration) -> Result<Stri
 async fn discover_cdp_ws(host: &str, port: u16, timeout: Duration) -> Result<String, String> {
     let ws_url = format!("ws://{}:{}/devtools/browser", bracket_ipv6(host), port);
 
-    // Single timeout budget covers connect + send + receive + close.
     tokio::time::timeout(timeout, async {
         let (mut ws_stream, _) = tokio_tungstenite::connect_async(&ws_url)
             .await


ctate:
Thanks for this excellent contribution, @jin-2-kakaoent!

PR #879, opened 2026-03-17
fix: use socket connectivity alone instead combining it with PID check for daemon liveness

Andarist:


ctate:
Thanks for this @Andarist!

PR #886, opened 2026-03-17
prepare v0.21

ctate:


.changeset/prepare-v0.21.md
Andarist, line 1: @ctate this caught my eye (I'm a Changesets maintainer) - it would be interesting to me why exactly you are going through this 2-step process here and if there is anything I could do in Changesets itself to improve the tool to support your workflow better

PR #891, opened 2026-03-17
fix: load extensions in recorded tab by using default browser context

### LINKED ISSUE #890: [Bug] Extensions not loaded in recorded tab (record start --extension)
## Description

When using `agent-browser record start video.webm --extension ./my-extension`, the extension is **not active** in the recorded tab. The extension loads correctly in normal tabs, but the recording tab does not have extension content scripts injected.

## Reproduction

```bash
# 1. Create a minimal test extension
mkdir -p /tmp/test-extension
cat > /tmp/test-extension/manifest.json <<'EXTEOF'
{
  "manifest_version": 3,
  "name": "Test Extension",
  "version": "1.0",
  "content_scripts": [
    {
      "matches": ["<all_urls>"],
      "js": ["content.js"],
      "run_at": "document_idle"
    }
  ]
}
EXTEOF

cat > /tmp/test-extension/content.js <<'EXTEOF'
console.log('[TEST EXTENSION] Injected!');
window.TEST_EXTENSION_LOADED = true;
document.body.style.border = "5px solid red";
EXTEOF

# 2. Launch and record with extension
agent-browser --headed --extension /tmp/test-extension record start test.webm https://example.com

# 3. Check if content script ran in the recorded tab
agent-browser eval 'window.TEST_EXTENSION_LOADED'  # Returns undefined, should be true
```

The same extension works correctly in a normal (non-recording) tab:

```bash
agent-browser --headed --extension /tmp/test-extension open https://example.com
agent-browser eval 'window.TEST_EXTENSION_LOADED'  # Returns true ✅
```

## Expected Behavior

Extensions should be active in the recorded tab, just like they are in normal tabs.

## Root Cause

`handle_recording_start` in `cli/src/native/actions.rs` creates a **new browser context** via CDP's `Target.createBrowserContext` before creating the recording page. In Chrome, contexts created this way are incognito-like — extensions loaded via `--load-extension` only run in the **default browser context**, not in these isolated contexts.

The same issue affects `handle_window_new` (the `window new` command).

Normal page creation (e.g., `tab new`, `open`) uses `Target.createTarget` without a `browserContextId`, which places the page in the default context where extensions work.

## Proposed Fix

When extensions are loaded, skip `Target.createBrowserContext` and create the recording page in the default browser context instead. This preserves the isolated-context behavior when no extensions are present.

Reference implementation: https://github.com/v1s10n-4/agent-browser/pull/1

## Environment

- agent-browser: 0.21.0
- OS: macOS (Darwin 25.3.0, arm64)

## Related

- #640 — Chrome extensions not loading (content scripts not injected)

v1s10n-4:
Fixes #890 

When `--extension` is used, `record start` and `window_new` now create pages in the default browser context instead of a new isolated context. Chrome extensions only run in the default context; CDP-created contexts are incognito-like and do not load extensions.


v1s10n-4:
Hi @ctate, thanks again for the detailed review, I'm happy to contribute to this great project!
I've pushed updates addressing both items, ran `cargo fmt` and changed the test extension to use DOM mutations instead of `window` properties so assertions are visible from `Runtime.evaluate`.
Would you mind taking another look when you get a chance?
Thanks for your time and for all the work you do on this project!

PR #894, opened 2026-03-18
fix: restore origin-scoped --headers persistence across commands

mykmelez:
In the v0.20 Rust rewrite, headers passed via --headers on open were only applied to that single navigation via Network.setExtraHTTPHeaders, which did not persist them for subsequent commands. In v0.19 (Playwright-based), these headers persisted for all subsequent same-origin requests.

This restores the v0.19 behavior using CDP Fetch interception:

- A background task processes Fetch.requestPaused events in real-time, injecting origin-scoped headers into matching requests and continuing non-matching requests unmodified. This avoids the deadlock that occurs when Fetch interception pauses requests during Page.navigate or Runtime.evaluate (which block waiting for completion).

- The same background task also handles domain filtering and route interception, replacing the previous drain-between-commands approach that couldn't process events during navigation or script evaluation.

Fixes:
- --headers persist for same-origin navigations without re-passing flag
- --headers persist for in-page fetch/XHR to the same origin
- --headers do not leak to cross-origin navigations or sub-resources
- `set headers` (global) is unaffected and stacks with --headers
- Domain filter Fetch interception no longer deadlocks during navigation

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -14,7 +14,8 @@ use super::cdp::chrome::LaunchOptions;
 use super::cdp::client::CdpClient;
 use super::cdp::types::{
     AttachToTargetParams, AttachToTargetResult, CdpEvent, ConsoleApiCalledEvent,
-    CreateTargetResult, ExceptionThrownEvent, TargetCreatedEvent, TargetDestroyedEvent,
+    CreateTargetResult, DispatchMouseEventParams, ExceptionThrownEvent, TargetCreatedEvent,
+    TargetDestroyedEvent,
 };
 use super::cookies;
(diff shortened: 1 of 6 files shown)


ctate:
Thanks for this excellent contribution! One minor nit: cargo fmt --check flags a few formatting diffs in actions.rs — a quick cargo fmt should clear those.

mykmelez:
> One minor nit: cargo fmt --check flags a few formatting diffs in actions.rs — a quick cargo fmt should clear those.

Whoops! I've run `cargo fmt` and committed/pushed the fixes.

ctate:
This is amazing. Thanks @mykmelez!

Issue #896, opened 2026-03-18
[Bug] Multiple --session daemons sharing same --profile connect to single Chrome, causing tab accumulation and navigation timeout


tetsuzawa:
## Description

When multiple `--session` daemons share the same `--profile` path (either via `--profile` flag, `AGENT_BROWSER_PROFILE` env var, or `agent-browser.json`), the second daemon seems to connect to the existing Chrome instance instead of launching a new one. This causes:

1. **Extra `chrome://newtab` tabs** — each new session adds ~3 newtab tabs
2. **Navigation timeout** for the second session (target URL never loads)
3. **Tab accumulation** that persists across daemon restarts via Chrome session restore

I believe this happens because Chrome enforces a singleton constraint per `--user-data-dir` — when a second Chrome process is launched with the same directory, it delegates to the existing process rather than starting a new instance.

**This does NOT occur without `--profile`**, because each session gets a unique UUID temp directory by default.

## Steps to Reproduce

```bash
# Setup: agent-browser.json with shared profile
echo '{"profile": "/tmp/test-profile", "headed": true}' > agent-browser.json

# 1. Clean state
pkill -9 -f 'agent-browser-daemon' 2>/dev/null
pkill -9 -f 'Chrome for Testing' 2>/dev/null
rm -rf /tmp/test-profile

# 2. First session (works fine)
agent-browser --session first open "https://example.com"
# → Success, 2 tabs (example.com + initial chrome://newtab)

# 3. Second session (bug)
agent-browser --session second open "https://httpbin.org"
# → Timeout error, +3 newtab tabs added to same Chrome instance
# → httpbin.org never loads
```

## Expected Behavior

Each `--session` should operate an independent Chrome instance, even when `--profile` is set. Either:
- Automatically derive a unique `user-data-dir` per session (e.g., `<profile>/<session>/`)
- Detect the conflict and emit a clear error message
- Document that `--profile` must be unique per session

## Actual Behavior

- 2nd daemon silently connects to the 1st daemon's Chrome via the shared `DevToolsActivePort` file
- The internal target filter excludes `chrome://newtab`, so the target discovery sees no usable page targets and creates new tabs
- Navigation targets the wrong session's page context
- Tab count grows from 2 → 5 with a single additional session

## Verified Reproduction

| Test | Profile | 1st Session | 2nd Session | Tabs | Bug? |
|------|---------|-------------|-------------|------|------|
[...]

PR #897, opened 2026-03-18
fix: pass about:blank as Chrome initial URL to prevent newtab proliferation

tetsuzawa:
## Summary

- Pass `about:blank` as a positional argument when launching Chrome, preventing the default `chrome://newtab` from being opened
- This allows `discover_and_attach_targets()` to find and reuse the initial tab instead of creating a new one
- Follows the same approach as [Puppeteer's ChromeLauncher.ts](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/node/ChromeLauncher.ts)

Relates to #896

## Changes

- `cli/src/native/cdp/chrome.rs`: Add `about:blank` to Chrome launch args in `build_chrome_args()` when no explicit URL is provided by the user via `AGENT_BROWSER_ARGS` or `--args`
- `cli/src/native/cdp/chrome.rs`: Add 3 tests for the new behavior

## Testing

- `cargo test --bin agent-browser -- chrome::tests` — all 22 tests pass (19 existing + 3 new)
- `cargo fmt -- --check` — clean
- Manual verification:
  - **With shared profile**: initial tab count reduced from 2 → 1 (no more `chrome://newtab`)
  - **Without profile**: no regression, sessions remain fully isolated

## Context

When Chrome launches without an explicit URL, it opens `chrome://newtab` by default. The internal target filter excludes this tab, so `discover_and_attach_targets()` sees zero page targets and creates a new `about:blank` tab. This is especially noticeable when multiple `--session` daemons share the same `--profile` (see #896).

This PR addresses the tab creation side of the problem. The broader session isolation question (how `--session` and `--profile` should interact when the same path is shared) is discussed in #896.

This is a small, low-risk change — happy to adjust if you'd prefer a different approach!

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -189,9 +189,7 @@ fn build_chrome_args(options: &LaunchOptions) -> Result<ChromeArgs, String> {
         args.push("--disable-dev-shm-usage".to_string());
     }
 
-    // Prevent chrome://newtab (filtered by is_internal_chrome_target) from being
-    // the only page target — forces discover_and_attach_targets to create extra tabs.
-    // Matches Puppeteer's ChromeLauncher.ts approach.
+    // Avoid chrome://newtab as the initial page — Puppeteer does the same.
     let has_url_arg = options.args.iter().any(|a| {
         !a.starts_with("--")
             && (a.starts_with("http")


tetsuzawa:
Closing this PR — the about:blank change only addresses a minor symptom (reducing initial tab count from 2 to 1) but does not fix the core issue described in #896. I'll leave the issue open for discussion on the broader session isolation problem. Thank you\!

PR #905, opened 2026-03-18
fix: escape text of cursor-interactive elements

Lppy:
# Summary
This PR addresses an issue where special characters in cursor-interactive elements were not properly escaped, which could cause parsing errors or display issues in the snapshot output.

# Changes
   - Added proper escaping for backslashes (\) by replacing them with double backslashes (\\)
   - Added proper escaping for double quotes (") by replacing them with escaped quotes (\")
   - Replaced newline (\n) and carriage return (\r) characters with spaces to ensure clean single-line text representation

### DIFF
--- a/cli/src/native/snapshot.rs
+++ b/cli/src/native/snapshot.rs
@@ -843,16 +843,17 @@ fn render_tree(
     let mut line = format!("{}- {}", prefix, role);
 
     // Use ARIA name if available, otherwise fall back to cursor-interactive textContent
-    let display_name = if !node.name.is_empty() {
-        node.name.clone()
+    let unescaped_display_name = if !node.name.is_empty() {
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the fix! The escaping logic for `ci.text` looks correct and the build/clippy pass clean.

One issue: `node.name` should get the same escaping treatment — it could contain the same special characters. Right now only the `ci.text` path is escaped while the two `node.name` paths are passed through raw. Could you apply consistent escaping to all three branches (or extract a helper)?

Lppy:
Of course, this is a more robust approach. Now the entire display_name has been escaped.

Lppy:
merge to PR[#968](https://github.com/vercel-labs/agent-browser/pull/968)

PR #909, opened 2026-03-18
fix: dedup text content in snapshot

Lppy:
# Summary

  This PR addresses an issue where InlineTextBox accessibility nodes were being included in the snapshot output. InlineTextBox nodes are purely layout-related and should not be presented to the agent as they don't represent meaningful interactive or semantic elements.

# Changes

  The change modifies the node filtering logic in build_tree() to exclude InlineTextBox nodes in addition to the existing ignored node filtering, while preserving all actual text content from parent elements.

   1. Core Logic Fix (cli/src/native/snapshot.rs):
      - Modified the condition in build_tree() function to filter out InlineTextBox nodes
      - Updated from: `if node.ignored.unwrap_or(false) && role != "RootWebArea"`
      - Updated to: `if (node.ignored.unwrap_or(false) && role != "RootWebArea") || role == "InlineTextBox"`

   2. End-to-End Test (cli/src/native/e2e_tests.rs):
      - Added comprehensive E2E test e2e_snapshot_inline_text_box_filtered
      - Verifies that InlineTextBox nodes are filtered from snapshot output
      - Confirms that actual text content ("Hello" and "World") is preserved

# Testing
   - ✅ E2E test `cargo test e2e_snapshot_inline_text_box_filtered -- --ignored` passes successfully
   - ✅ Confirms no InlineTextBox appears in output while preserving text content

# Impact
   - Example:
```
<div>
  <span>Hello</span> <span>World</span>
</div>
```
   - Before:
```
  - StaticText "Hello"
    - InlineTextBox "Hello"
  - StaticText " "
    - InlineTextBox " "
  - StaticText "World"
    - InlineTextBox "World"
```
   - After: 
 ```
  - StaticText "Hello"
  - StaticText " "
  - StaticText "World"
```

### DIFF
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -2226,8 +2226,9 @@ async fn e2e_snapshot_inline_text_box_filtered() {
     assert_success(&resp);
 
     // Simple HTML with text content that would generate InlineTextBox nodes
-    let html = "data:text/html,<html><body><div><span>Hello</span> <span>World</span></div></body></html>";
-    
+    let html =
+        "data:text/html,<html><body><div><span>Hello</span> <span>World</span></div></body></html>";
+
     let resp = execute_command(
         &json!({ "id": "2", "action": "navigate", "url": html }),
         &mut state,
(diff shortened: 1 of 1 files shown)


ctate:
Thank you for this fix @Lppy

PR #916, opened 2026-03-18
perf: early-exit in diff_screenshot to skip encoding clearly-different images

light-merlin-dark:
## Summary

- Add early-exit to diff_screenshot: when >5% diff rate is detected after sampling 10,000 pixels, return immediately with diff_image: None
- Avoids expensive full RGBA buffer construction and PNG encoding for clearly-different images
- Primary win: 2-3x speedup on the common "very different" case

## Benchmark Results (M3 MacBook Air, 5 iterations)

| Scenario | Before | After | Speedup |
|----------|--------|-------|---------|
| Identical 1080p | 5ms | 6ms | 0.83x (minor overhead) |
| 1% changed 1080p | 4ms | 11ms | 0.36x (regression) |
| 50% changed 1080p | 4ms | 2ms | 2x faster |
| Totally different 1080p | 3ms | 1ms | 3x faster |
| Gradient identical 1080p | 20ms | 21ms | 0.95x (minor overhead) |
| Totally different 4K | 14ms | 6ms | 2.3x faster |

## Tradeoff Analysis

The 1% case regresses (4ms -> 11ms) because early-exit overhead is paid but 5% threshold isn't reached. Acceptable tradeoff: agent workloads produce clearly-different or identical results, not marginal 1% changes. Phase 3 (rayon parallelization) will reduce absolute overhead further.

## Changes

- `cli/src/native/diff.rs`: +125 lines -- early-exit logic, 4 unit tests

## Verification

- `cargo test` -- 480 passed, 0 failed
- `cargo clippy` -- clean (1 pre-existing unrelated warning)
- `cargo fmt -- --check` -- clean

---

[Merlin](https://rbeckner.com) ([@EnchantedRobot](https://x.com/EnchantedRobot) on X) & Hermes (minimax-m2.7)

### DIFF
--- a/cli/src/native/diff.rs
+++ b/cli/src/native/diff.rs
@@ -87,13 +87,23 @@ pub fn diff_screenshot(
                 let current_diff_rate = different as f64 / checked as f64;
                 if current_diff_rate >= EARLY_EXIT_THRESHOLD {
                     // Early exit -- images are clearly different.
-                    // The final mismatch will be at least current_diff_rate percent.
-                    // Return early without encoding the partial diff image.
-                    let mismatch = (different as f64 / total as f64) * 100.0;
+                    // IMPORTANT: `different` here is only the count within the sampled pixels
+                    // (up to `checked`), but our output fields are expected to be expressed
+                    // relative to the *full* image pixel count.
+                    //
+                    // So we estimate the overall diff rate using the sample rate and scale it
+                    // up to the full image size.
+                    let sample_diff_rate = current_diff_rate; // fraction in [0, 1]
+                    let mut estimated_different = (sample_diff_rate * total as f64).round();
+                    if estimated_different > total as f64 {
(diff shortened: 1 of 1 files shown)


light-merlin-dark:
Addressing Vercel review on #916: the early-exit path reported inconsistent diff metrics because `different_pixels` was computed from the sampled/checked pixels while `total_pixels` remained full-image pixels.

I updated `diff_screenshot` early-exit to scale the sampled diff rate up to the full-image pixel count so that `mismatch_percentage == different_pixels/total_pixels * 100` (within rounding), and added tests to lock this invariant.

Pushed commit 3d99908 (fix(diff_screenshot): scale early-exit mismatch metrics).

[Merlin](https://rbeckner.com) ([@EnchantedRobot](https://x.com/EnchantedRobot) on X) & Hermes (gpt-5.4-nano)

PR #924, opened 2026-03-19
feat: add shell completion for bash and zsh

mifisignal:
Add `agent-browser completion <shell>` command that prints a shell completion script for bash or zsh. Introduce named module-level constants for every subcommand list (replacing local const blocks) so the new completion test suite can assert that each script covers all top-level commands, subcommands, and global flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,18 @@ agent-browser upgrade
 
 Detects your installation method (npm, Homebrew, or Cargo) and runs the appropriate update command automatically.
 
+### Shell Completion
+
+Enable tab completion for commands and flags:
+
+```bash
+# bash (~/.bashrc or ~/.bash_profile)
+eval "$(agent-browser completion bash)"
+
+# zsh (~/.zshrc)
+eval "$(agent-browser completion zsh)"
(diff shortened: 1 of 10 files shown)


mifisignal:
> Great work — the completion scripts are thorough and the test macro that cross-checks against Rust source is a smart way to prevent drift. Three things needed before merging:
> 
> 1. **Run `cargo fmt`** — several long const arrays in `commands.rs` exceed the line width limit, so CI will fail without this.

✅ done

> 2. **Add docs** to `skills/agent-browser/SKILL.md` and `docs/src/app/` per project convention (AGENTS.md requires all 4 doc locations updated for new user-facing commands; `output.rs` and `README.md` are already done).

✅ done, with one exception: The autocomplete functionality is to assist users who are using the tool interactively in a shell, so I can't imagine a situation in which a coding agent would bother with it. In light of that, do you agree that we can skip updating the Skill?

> 3. **Typo** — `completion.rs:117`: "autoompletion" → "autocompletion".

✏️   fixed

Let me know if there's anything else @ctate !

Issue #774, opened 2026-03-14
docs: content drift between README and docs site — proposed phased fix


mikewong23571:
While exploring the docs, I noticed a few things that seem worth fixing.

**1. Factual error in `ios/page.mdx`**

The "Differences from desktop" table lists `Chromium/Firefox/WebKit` as the desktop browser. The codebase only has launchers for Chrome and Lightpanda (both via CDP — `cdp/chrome.rs`, `cdp/lightpanda.rs`). Firefox and WebKit have no launcher and are not supported. The cell should read `Chrome, Lightpanda`.

**2. Provider docs are incomplete and in the wrong place**

`cdp-mode/page.mdx` has a "Cloud providers" section, but it only shows an API key export and one command per provider. The README has significantly more: env var tables, provider-specific options, behavioral notes.

The placement is also a mismatch. CDP Mode is about connecting to an *already-running* browser. Providers *provision* a managed browser on demand — the CDP connection is an internal detail the user never touches. These are different user journeys.

One more thing: the README documents incorrect defaults for Kernel vs. the source code (`providers.rs`):

| Variable | README | Code |
|---|---|---|
| `KERNEL_HEADLESS` | `false` | `true` |
| `KERNEL_STEALTH` | `true` | `false` |

**3. A few features have no dedicated docs coverage**

`--headed` and `--allow-file-access` each appear as a one-line table row in cdp-mode's options table. Both have fuller explanations in the README that aren't reflected anywhere in the docs site.

**4. MDX metadata renders as JS on GitHub**

Most `page.mdx` files open with:

```mdx
import { pageMetadata } from "@/lib/page-metadata"
export const metadata = pageMetadata("slug")
```

GitHub renders this as plain text. `installation/page.mdx` already uses YAML frontmatter — that pattern avoids the problem entirely.

mikewong23571:
@ctate I've made a start on these in a branch if you'd like to see the approach: https://github.com/mikewong23571/agent-browser/tree/feat/mdx-frontmatter-metadata. Happy to open a PR — or adjust direction entirely based on your feedback.

ctate:
@mikewong23571 I really like this direction. The main call-out I have would be to call "integrations" -> "providers" to keep it consistent, including the URL

mikewong23571:
## Incremental PR Plan

Breaking this into small, focused PRs — each based on latest `main`, merged before the next starts.

[...]

PR #929, opened 2026-03-19
docs: update cdp-mode — remove stale fallback line, add --full flag

mikewong23571:
## Summary
- Remove outdated auto-connect step 3 ("If HTTP-based discovery fails, falling back to a direct WebSocket connection") — this behavior no longer exists
- Add `--full, -f` to the global options table (captures full page screenshot, not just viewport)

Part of #774 (item 4 in the [incremental PR plan](https://github.com/vercel-labs/agent-browser/issues/774#issuecomment-2988053657))

## Test plan
- [ ] `cd docs && npm run build` passes
- [ ] cdp-mode page renders correctly

### DIFF
--- a/docs/src/app/cdp-mode/page.mdx
+++ b/docs/src/app/cdp-mode/page.mdx
@@ -49,7 +49,7 @@ Auto-connect discovers Chrome by:
 
 1. Reading Chrome's `DevToolsActivePort` file from the default user data directory
 2. Falling back to probing common debugging ports (9222, 9229)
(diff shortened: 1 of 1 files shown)


mikewong23571:
Closing — both changes are inaccurate after further source code verification: the WebSocket fallback was actually added in #870 (so removing it from docs is wrong), and `--full` was refactored to a command-level flag in #877 (so adding it to the global options table is misleading).

Issue #555, opened 2026-02-27
[Bug] `network requests` does not capture page navigations — only XHR/fetch (distinct from #130)


digitall-it:
## Summary

`network requests` only records XHR and fetch requests. Full page navigations (document loads, redirects, form submissions that trigger a page load) are **never captured**, making it impossible to inspect them via this command.

> **Note:** This is distinct from issue #130, which reports that `--filter` is required even when not filtering. This bug is about the **scope** of what gets captured, not about the command's argument handling.

## Environment

- **agent-browser version:** 0.15.1
- **OS:** macOS darwin 25.3.0
- **Install method:** `npm install -g agent-browser`

## Steps to Reproduce

```bash
# 1. Open a page
agent-browser --session s1 open https://the-internet.herokuapp.com/redirector

# 2. Take snapshot to find the redirect button
agent-browser --session s1 snapshot -i
# Output: link "here" [ref=e1] (triggers a redirect to /status_codes)

# 3. Click the link (triggers a full page navigation / redirect)
agent-browser --session s1 click @e1

# 4. Wait for page load and inspect captured network requests
agent-browser --session s1 wait --load networkidle
agent-browser --session s1 network requests --filter ""
```

**Actual output:**
```
(empty — no requests listed)
```

**Expected output:**
The navigation request (e.g. `GET /status_codes`) and any redirect chain should appear in the request log.

---

**Contrast with XHR (works correctly):**
```bash
agent-browser --session s1 open https://reqres.in
# Trigger an XHR by interacting with the page
agent-browser --session s1 network requests --filter "api"
# ✅ XHR requests to /api/... are captured
```

## Expected Behavior

`network requests` should capture all network activity including:
- Full page navigations (document requests)
- Redirects (3xx chains)
- Form submissions that trigger page loads
- Resource loads (scripts, stylesheets, images) — optionally filterable

## Actual Behavior

Only `XMLHttpRequest` and `fetch()` calls are captured. Document-level requests are invisible.

## Impact

Workflows that need to inspect what URL was navigated to, detect redirects, or audit page-level network traffic (e.g. verifying a login redirects to the correct endpoint) cannot rely on `network requests` and must use workarounds like `get url` polling.

ChunHao-dev:
Hey @ctate , I dug into this while working on #935 and found the root cause.

Root cause

[...]

PR #935, opened 2026-03-19
feat: add network request detail and filtering for request tracking

### LINKED ISSUE #932: Request and Response Body is not being captured in network request
This is a very common use case for agents where they require HTML body content to make sense of the request/reponse parameters and the type of values they contain. But when agent runs "network requests", everything is visible except the request and response body! Is there a fix coming up for this?

ChunHao-dev:
## Summary

Add `network request <requestId>` command and filtering flags (`--type`, `--method`, `--status`) for `network requests`.

fixed #932

## Changes

- `network request <requestId>` — view full request/response detail including response body via CDP `Network.getResponseBody`
- `--type xhr,fetch` — filter by resource type (comma-separated)
- `--method POST` — filter by HTTP method
- `--status 2xx` — filter by status code (exact `200`, class `2xx`, range `400-499`)
- Extend `TrackedRequest` with `request_id`, `post_data`, `status`, `response_headers`, `mime_type`
- `Network.responseReceived` now also updates `tracked_requests` (previously only HAR entries)
- Updated README, SKILL.md, docs/commands MDX, and help text

## Test plan

- `cargo fmt --check` ✅
- `cargo clippy -- -D warnings` ✅
- `cargo test` — 490 passed, 0 failed ✅
- Added 7 parse tests for filter flags and `network request` command
- Added `matches_status_filter` unit tests (exact, class, range, None)

### DIFF
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -374,7 +374,15 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
                         .get("resourceType")
                         .and_then(|v| v.as_str())
                         .unwrap_or("");
-                    println!("{} {} ({})", method, url, resource_type);
+                    let request_id = req.get("requestId").and_then(|v| v.as_str()).unwrap_or("");
+                    let status = req.get("status").and_then(|v| v.as_i64());
+                    match status {
+                        Some(s) => println!(
+                            "[{}] {} {} ({}) {}",
+                            request_id, method, url, resource_type, s
+                        ),
+                        None => println!("[{}] {} {} ({})", request_id, method, url, resource_type),
+                    }
                 }
             }
             return;


ChunHao-dev:
Done! Added request ID and status to the formatted output in `output.rs`. 

Example:
[19234.3] POST https://httpbin.org/post (Fetch) 200

Requests without a response yet omit the status code.

ChunHao-dev:
I found some existing behaviors during testing:

1. Lazy activation — Network.enable is sent only on the first network requests call. So if the user runs open first, requests from page load are missed. This is the root cause of #555.

2. Single-tab only — Network.enable is sent to the current active tab. If the user opens a new tab, it won't be tracked.

3. No limit — tracked_requests keeps growing. On sites like x.com that poll frequently, the list gets very long.

These are not from this PR — they already existed before.

I thought about adding a tip in the docs to tell users: run network requests before open to capture page load requests. But I want to ask you first — should I add it? And where?

Also, I can open a follow-up issue and PR to improve these (e.g. re-send Network.enable on tab switch, add a max size for tracked_requests). Let me know if you'd like that!

ChunHao-dev:
@ctate Changes addressed — ready for re-review when you get a chance.

ctate:
Looks great. Thank you @ChunHao-dev

PR #945, opened 2026-03-20
fix: make auth login selector targeting more reliable

third774:
# Description

- **`auth login` navigation strategy**: Changed from `NetworkIdle` to `Load` + explicit selector polling. Added `wait_for_any_selector` helper that polls for visible, enabled form elements via `Runtime.evaluate`.
- **Staged username detection**: Username auto-detection now uses a two-phase approach -- targeted selectors (email/username by `type`, `name`, `id`, `autocomplete`) are tried first for a configurable window, then broad fallback selectors (`input[type=text]`, `input:not([type])`) are tried with remaining timeout. This replaces the old single-pass `querySelector` check.
- **Password and submit selectors also wait**: Password field and submit button detection now use `wait_for_selector` before interaction, rather than assuming elements exist immediately after navigation.
- **Expanded selector list**: Added selectors for `autocomplete=email`, `autocomplete=username`, case-insensitive `name`/`id` attribute matching (`i` flag), and a new `input:not([type])` fallback.
- **E2E test**: New `e2e_auth_login_waits_for_delayed_spa_form_render` test with a purpose-built delayed-render login server that keeps `/ping` requests active (preventing `networkidle`) while rendering the login form after a 1.2s delay. Verifies the email field is filled (not the decoy search input), password is filled, and form is submitted.
- **Documentation**: Updated `--help` output, README, SKILL.md, and docs site (commands + security pages) to describe the new wait behavior.

# Context

The previous `auth login` behavior navigated with `networkidle` and immediately checked for form elements with a single `querySelector` pass. This was problematic in two ways:

1. **SPA timing failures** -- Single page applications that render login forms after initial page load (via client-side JS) would not have form elements present when `auth login` tried to fill them. The `networkidle` wait strategy compounded this by hanging indefinitely on pages with continuous background requests (polling, analytics, websockets), meaning navigation itself could time out before the form ever rendered.

2. **Incorrect field matching** -- The old selector list was broad enough that unrelated text inputs (search bars, filters) could match before the actual login form rendered, causing credentials to be typed into the wrong field.

The new approach navigates with `load` (fires reliably once the page's initial HTML is parsed) and then actively polls for specific login form selectors. The staged detection gives targeted email/username selectors priority, only falling back to broad `input[type=text]` if specific selectors aren't found within the first 5 seconds. This should work reliably across a broader range of login pages.


ctate:
@third774 Thanks for the fix!

PR #955, opened 2026-03-21
fix: prevent find flags from leaking into fill value

daidi:
## Summary
- fix `find` parsing so `--name` and `--exact` are treated as flags, not part of fill text
- keep non-flag tokens after action as fill value
- add regression test for `find role ... fill ... --name ... --exact`

## Repro
```bash
agent-browser --session repro open about:blank
agent-browser --session repro eval "document.body.innerHTML = '<label for=\"u\">User Name</label><input id=\"u\" type=\"text\" />'"
agent-browser --session repro find label "User Name" fill hello --name Foo --exact
agent-browser --session repro get value "#u"
```
Before: `hello --name Foo --exact`  
After: `hello`

## Validation
- `cd cli && cargo fmt`
- `cd cli && cargo test test_find_role_fill_does_not_include_flags_in_value -- --nocapture`

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1742,10 +1742,6 @@ fn parse_find(rest: &[&str], id: &str) -> Result<Value, ParseError> {
         usage: "find <locator> <value> [action] [text]",
     })?;
 
-    let name_idx = rest.iter().position(|&s| s == "--name");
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the fix @daidi

PR #969, opened 2026-03-23
fix: enhance target tracking and update page information handling

zzzgydi:
When I try to use `agent-browser tab` to get the tab list, I found that the tab which open by the `window.open('xxx')` will be ignored.


ctate:
Thanks for the contribution, @zzzgydi!

PR #984, opened 2026-03-23
Fix modifier chord parsing for press

### LINKED ISSUE #983: press does not correctly send modifier chords like Meta+,
## Summary

`agent-browser press <key>` is documented as supporting modifier chords like `Control+a` and `Control+Shift+s`, but the current implementation does not correctly send modifier combos in practice.

I reproduced this against a live Electron app (`Squad`) on macOS with `agent-browser 0.22.0`:

```bash
agent-browser --session squad connect 9222
agent-browser --session squad press 'Meta+,'
```

Expected: the app receives `metaKey=true` with `event.key === ','` and opens Settings.

Actual: the route stays unchanged and the app does not receive the expected modifier combo.

## Repro details

In the target app, the shortcut handler listens for:

- `event.metaKey || event.ctrlKey`
- `event.key === ','`

Manual `Cmd+,` works.

`agent-browser press ','` also works as a plain comma key, so punctuation synthesis itself is fine.

The failure is specifically in modifier-chord handling.

## Root cause

`press` routes through `handle_press()` in `cli/src/native/actions.rs`, which forwards the raw string to `interaction::press_key(...)`.

That helper immediately calls `press_key_with_modifiers(..., None)`, so the documented chord syntax never gets parsed into CDP modifier bits.

## Proposed fix

Parse leading modifier segments in the CLI `press` action (for example `Control+`, `Shift+`, `Meta+`, `Alt+`) and pass the remaining key plus the combined modifier bitmask into the existing `press_key_with_modifiers(...)` helper.

That keeps the fix at the CLI seam and reuses the existing lower-level CDP primitive.

## Verification

I validated a local patch with:

- targeted unit tests for chord parsing
- full `cargo test` in `cli/`
- live repro against the Electron app, where the patched binary successfully opened `#/settings` with `press 'Meta+,'`

I’m opening a PR with the fix and receipts.

danialhasan:
## Summary

Fix `press <key>` so documented modifier chords like `Control+a`, `Control+Shift+s`, and `Meta+,` are parsed into CDP modifier bits before dispatch.

This keeps the change at the CLI action seam and reuses the existing `press_key_with_modifiers(...)` interaction primitive rather than adding a second input abstraction.

Closes #983.

## What changed

- parse leading modifier segments in `handle_press()`
- pass the remaining key plus combined modifier bitmask into `press_key_with_modifiers(...)`
- keep plain keys and unknown names on their existing concrete path
- add unit coverage for plain keys, modifier chords, punctuation chords, and `Control++`
- update help/docs/skill examples to include a punctuation combo example (`'Meta+,'`)

## Why this shape

The bug was in CLI chord parsing, not in the lower CDP dispatch primitive.

So the fix stays where the user-facing syntax is interpreted:

- no new input subsystem
- no change to existing lower-level key dispatch behavior beyond using the already-present modifier-aware helper
- no duplicate source of truth for key semantics

## Verification

- `cd cli && cargo fmt`
- `cd cli && cargo test`
- live repro against a running Electron app over CDP:
  - before the fix, `press 'Meta+,'` did not open Settings
  - after the fix, the patched binary opened `#/settings`
  - plain `press ','` still behaved as a plain comma key

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ agent-browser dblclick <sel>          # Double-click element
 agent-browser focus <sel>             # Focus element
 agent-browser type <sel> <text>       # Type into element
 agent-browser fill <sel> <text>       # Clear and fill
-agent-browser press <key>             # Press key (Enter, Tab, Control+a) (alias: key)
+agent-browser press <key>             # Press key (Enter, Tab, Control+a, Meta+,) (alias: key)
 agent-browser keyboard type <text>    # Type with real keystrokes (no selector, current focus)
 agent-browser keyboard inserttext <text>  # Insert text without key events (no selector)
 agent-browser keydown <key>           # Hold key down
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -2093,8 +2093,49 @@ async fn handle_press(cmd: &Value, state: &mut DaemonState) -> Result<Value, Str
         .and_then(|v| v.as_str())
         .ok_or("Missing 'key' parameter")?;
 
-    interaction::press_key(&mgr.client, &session_id, key).await?;
+    let (base_key, modifiers) = split_key_chord(key);
+    match modifiers {
+        Some(bits) => {
+            interaction::press_key_with_modifiers(&mgr.client, &session_id, base_key, Some(bits))
+                .await?
+        }
+        None => interaction::press_key(&mgr.client, &session_id, base_key).await?,
+    }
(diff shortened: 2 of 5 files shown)


ctate:
Thanks for this PR. This issue was fixed earlier today by #980 and will be published this evening

danialhasan:
> Thanks for this PR. This issue was fixed earlier today by #980 and will be published this evening

thankyou! closing.

PR #994, opened 2026-03-24
fix: `state show` always fails with "Missing 'path' parameter"

wanghanzhen:
`agent-browser state show <filename>` unconditionally errored with `✗ Missing 'path' parameter` regardless of input due to a key name mismatch between the CLI parser and the command dispatcher.

## Root Cause

- `commands.rs` serialized the filename argument under the key `"filename"`
- `state.rs` dispatcher looked it up as `"path"` → always `None` → always errored

## Fix

- **`cli/src/commands.rs`**: change the JSON key from `"filename"` to `"path"` in the `state_show` command object

```rust
// Before
Ok(json!({ "id": id, "action": "state_show", "filename": filename }))

// After
Ok(json!({ "id": id, "action": "state_show", "path": filename }))
```

Note: existing tests didn't catch this because they bypass `commands.rs` and construct the dispatch JSON directly with the correct `"path"` key.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1223,7 +1223,7 @@ pub fn parse_command(args: &[String], flags: &Flags) -> Result<Value, ParseError
                         context: "state show".to_string(),
                         usage: "state show <filename>",
                     })?;
-                    Ok(json!({ "id": id, "action": "state_show", "filename": filename }))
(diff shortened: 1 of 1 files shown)


ctate:
Great catch @wanghanzhen - thank you for fixing

PR #1004, opened 2026-03-24
docs: add troubleshooting note for Chrome remote debugging checkbox

ImBIOS:
## Summary

- Added troubleshooting note to README explaining that `--auto-connect` requires the "Allow remote debugging for this browser instance" checkbox in `chrome://inspect` to be enabled
- This addresses confusion where users launching Chrome with `--remote-debugging-port` still get "No running Chrome instance found" errors

## Test plan

- [x] Verify the note appears in the Authentication section of README

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,6 +398,9 @@ agent-browser --session-name myapp state load ./my-auth.json
 > - `--remote-debugging-port` exposes full browser control on localhost. Any local process can connect. Only use on trusted machines and close Chrome when done.
 > - State files contain session tokens in plaintext. Add them to `.gitignore` and delete when no longer needed. For encryption at rest, set `AGENT_BROWSER_ENCRYPTION_KEY` (see [State Encryption](#state-encryption)).
 
+> **Troubleshooting:**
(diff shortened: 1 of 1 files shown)


ctate:
Thanks @ImBIOS! Mind updating the skill too?

thesobercoder:
This is important beyond just troubleshooting — as Chrome moves toward `chrome://inspect/#remote-debugging` as the default (with `--remote-debugging-port` deprecated on real user profiles since Chrome 136), the SKILL.md that ships with agent-browser should also document the direct WebSocket connection workaround.

Currently, `--cdp 9222` fails because Chrome's new native CDP server doesn't expose `/json/version`. But `--cdp "ws://127.0.0.1:9222/devtools/browser"` works perfectly. This isn't obvious and took significant debugging to discover.

For AI agents using the skill, this matters because the skill docs are the only context they have — if the workaround isn't documented there, agents will repeatedly fail to connect to a user's running Chrome session and fall back to launching a separate headless instance, which defeats the purpose of reusing an authenticated browser session.

---
*Disclaimer: This comment was written by Claude (Opus 4.6) on behalf of @thesobercoder, after investigating and debugging the issue together in a Claude Code session.*

PR #1013, opened 2026-03-25
fix: inline VS Code webviews in snapshots

zhanba:
## Summary
- track VS Code webview iframe targets so their sessions are available to snapshotting
- resolve the nested active-frame inside VS Code webviews when taking snapshots
- reuse the same frame lookup helper in frame selection and add focused VS Code regression coverage

## Testing
- cargo test vscode --manifest-path cli/Cargo.toml


zhanba:
@ctate would you consider this pr? it is important as VSCode extension api forces all extensions to use this two level nested iframe structure, and agent-browser can't work with it properly.

PR #1014, opened 2026-03-25
fix: route keyboard type through text input

zhanba:
## Summary
- make keyboard type use the high-level text typing path instead of a single raw dispatchKeyEvent payload
- make keyboard inserttext explicitly call Input.insertText
- type printable characters via Input.insertText while keeping key events for control keys like Enter and Tab

## Testing
- cargo test native::interaction --manifest-path cli/Cargo.toml


ctate:
Great catch and fix. Thanks @zhanba

PR #1015, opened 2026-03-25
fix: handle --clear flag in console command

xu-yongliang:
## Problem

`console --clear` was parsed correctly into `{ "action": "console", "clear": true }` by the CLI, but the action handler silently ignored the flag.

**Root cause:** `handle_console` did not accept a `cmd` parameter, so it had no way to read the `clear` field — the flag was a no-op.

## Changes

- **`cli/src/native/network.rs`**: Added `clear_console()` method to `EventTracker` that clears the console entry buffer.
- **`cli/src/native/actions.rs`**: Updated `handle_console` to accept `cmd: &Value` and `state: &mut DaemonState`, read the `clear` field, and clear the buffer when `--clear` is passed (returns `{ "cleared": true }`). Updated the call site in `execute_command` to pass `cmd`.

## Behavior

| Command | Before | After |
|---|---|---|
| `console` | Returns logs | Returns logs |
| `console --clear` | Returns logs, buffer **not** cleared | Clears buffer, returns `{ "cleared": true }` |

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -1074,7 +1074,7 @@ pub async fn execute_command(cmd: &Value, state: &mut DaemonState) -> Value {
         "setcontent" => handle_setcontent(cmd, state).await,
         "headers" => handle_headers(cmd, state).await,
         "offline" => handle_offline(cmd, state).await,
-        "console" => handle_console(state).await,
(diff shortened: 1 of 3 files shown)


ctate:
Thanks for your contribution @xu-yongliang

PR #1042, opened 2026-03-27
fix: detect externally opened tabs in --cdp mode

### LINKED ISSUE #1037: tab list does not show tabs opened after --cdp connection
When connected to an existing Chrome instance via `--cdp <port>`, `tab list` only shows tabs that existed at connection time. Tabs opened manually by the user afterwards are never shown.

This is a **regression from v0.20.0** (the Rust native rewrite). v0.19.0 (Node.js/Playwright) correctly tracks new tabs via `context.on('page', ...)`.

**Versions:**
- agent-browser 0.22.3 (broken)
- agent-browser 0.19.0 (works)
- Chrome 147.0.7727.24
- macOS (arm64)

**Reproduction:**

```bash
# 1. Launch Chrome with remote debugging
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome \
  --remote-debugging-port=9222

# 2. Connect agent-browser (Chrome starts with one blank tab)
agent-browser --cdp 9222 tab list
# → [0] New tab - chrome://newtab/

# 3. Manually open a new tab in Chrome and navigate to any URL

# 4. Check tab list again
agent-browser --cdp 9222 tab list
# → [0] New tab - chrome://newtab/    ← new tab missing

# 5. Verify the tab exists via raw CDP
curl -s http://127.0.0.1:9222/json/list | python3 -c "
import json, sys
for t in json.load(sys.stdin):
    if t['type'] == 'page':
        print(f'{t[\"title\"]} — {t[\"url\"]}')"
# → New tab — chrome://newtab/
# → Example Domain — https://example.com/    ← visible here
```

**Comparison with v0.19.0:**

```bash
npm install --prefix /tmp/ab-old agent-browser@0.19.0
OLD=/tmp/ab-old/node_modules/.bin/agent-browser

# Same steps as above — v0.19.0 sees all tabs:
$OLD --cdp 9222 tab list
# → [0] New tab - chrome://new-tab-page/
# → [1] Example Domain - https://example.com/    ← works
```

**Root cause:**

Two issues in the Rust native implementation:

1. **`chrome://newtab/` filtered as internal target** (`browser.rs`). New tabs start with `url=chrome://newtab/`, which `is_internal_chrome_target()` matches on all `chrome://` URLs. `should_track_target()` rejects it. Chrome only sends `Target.targetCreated` once — the later `Target.targetInfoChanged` with the real URL cannot update a target that was never tracked.

2. **Daemon does not process target events between commands** (`daemon.rs`). The background drain loop only runs when `request_tracking || har_recording` is active. Otherwise, `Target.targetCreated` events sit in the broadcast channel (256 slots) and are silently dropped when it overflows — the `Lagged` error is swallowed with `continue`.

Both issues must be fixed together. Fixing (1) alone still fails because the daemon never processes the event between invocations. Fixing (2) alone still fails because the event is filtered out.

jin-2-kakaoent:
## Summary

Fixes #1037

- Promote untracked targets in `Target.targetInfoChanged` to new targets — tabs that were initially filtered (e.g. `chrome://newtab/`) are now detected when their URL changes
- Run the background drain loop unconditionally (guarded by `browser.is_some()`) so target events between CLI commands are not dropped
- Extract `apply_drained_events` to share target lifecycle processing (attach, domain filter, iframe sessions, screencast ACKs) between `execute_command` and the background drain, eliminating code duplication

## Test plan

- [x] Added `e2e_externally_opened_tab_detected` regression test that simulates an external CDP client opening a tab and verifies it appears in `tab list`
- [x] Existing `e2e_tabs` test passes
- [x] All 542 unit tests pass

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -465,9 +465,8 @@ impl DaemonState {
             if let Some(ref browser) = self.browser {
                 if let Ok(session_id) = browser.active_session_id() {
                     for ack_sid in drained.pending_acks {
-                        let _ =
-                            stream::ack_screencast_frame(&browser.client, session_id, ack_sid)
-                                .await;
+                        let _ = stream::ack_screencast_frame(&browser.client, session_id, ack_sid)
+                            .await;
                     }
                 }
             }
@@ -597,7 +596,9 @@ impl DaemonState {
                                         .browser
                                         .as_ref()
                                         .is_some_and(|b| b.has_target(&te.target_info.target_id));
-                                    if already_tracked || new_target_ids.contains(&te.target_info.target_id) {
+                                    if already_tracked
+                                        || new_target_ids.contains(&te.target_info.target_id)
+                                    {
                                         changed_targets.push(te);
                                     } else {
                                         new_target_ids.insert(te.target_info.target_id.clone());


ctate:
Thanks for the fix @jin-2-kakaoent - works great

PR #1019, opened 2026-03-25
fix: re-apply download behavior on recording context

shtefcs:
## Summary

`record start` creates a new browser context via `Target.createBrowserContext`, but `Browser.setDownloadBehavior` (called at launch time) only applies to the default context. Downloads triggered during recording are **silently dropped** — the file is never saved anywhere.

## Changes

- **browser.rs**: Store `download_path` on `BrowserManager` struct (from `LaunchOptions`)
- **actions.rs**: After creating the recording context in `handle_recording_start`, call `Browser.setDownloadBehavior` with the new `browserContextId`

## Root cause

In `browser.rs` launch():
```rust
// Line 302-308: Only applies to DEFAULT context (no browserContextId)
if let Some(ref path) = download_path {
    let _ = manager.client.send_command(
        "Browser.setDownloadBehavior",
        Some(json!({ "behavior": "allow", "downloadPath": path })),
        None,  // ← No session/context specified
    ).await;
}
```

In `actions.rs` `handle_recording_start()`:
- Creates new context via `Target.createBrowserContext` ✅
- Transfers cookies ✅
- Enables domains ✅
- **Does NOT re-apply download behavior** ❌

## Fix

```rust
// After enable_domains_pub in handle_recording_start:
if let Some(ref dl_path) = mgr.download_path {
    let _ = mgr.client.send_command(
        "Browser.setDownloadBehavior",
        Some(json!({
            "behavior": "allow",
            "downloadPath": dl_path,
            "browserContextId": context_id,
            "eventsEnabled": true
        })),
        None,
    ).await;
}
```

## Reproduction

```bash
export AGENT_BROWSER_DOWNLOAD_PATH=/tmp/downloads
mkdir -p /tmp/downloads

# Without recording — download works ✅
agent-browser open https://automatio-play-font-dir.vercel.app/success
agent-browser snapshot -i
agent-browser click @e3
sleep 10 && ls /tmp/downloads/  # Inter-4.0.zip appears

# With recording — download silently dropped ❌
agent-browser close
agent-browser open https://automatio-play-font-dir.vercel.app/success
agent-browser record start /tmp/test.webm
agent-browser snapshot -i
agent-browser click @e3
sleep 15 && ls /tmp/downloads/  # EMPTY — file is nowhere
```

## Test plan

- [ ] Verify downloads work without recording (regression check)
- [ ] Verify downloads work WITH recording active
- [ ] Verify recording still produces valid video files
- [ ] Verify cookies are still transferred to recording context

[...]

PR #1045, opened 2026-03-27
fix: propagate recording session to screencast stream

shtefcs:
## Problem

After `record start`, the live preview (screencast) stays on the old pre-recording context while CLI commands execute on the new recording context. This causes:

1. **Live preview stuck** — shows the old page, not the recording context
2. **Extensions appear broken** — content scripts inject into the recording context but the screencast shows the old context where nothing happens

## Root Cause

`handle_recording_start` creates a new `BrowserContext` via `Target.createBrowserContext`, adds a new page, and switches `active_page_index` to it. All subsequent CLI commands (`snapshot`, `eval`, `click`) use `mgr.active_session_id()` which returns the new recording session.

However, the `StreamServer`'s `cdp_session_id` (used for `Page.startScreencast`) is never updated. `update_stream_client()` is called by 13+ other commands that switch contexts (open, close, connect, tab switch, etc.) but was missing from both `handle_recording_start` and `handle_recording_stop`.

## Fix

Add `state.update_stream_client().await` in both `handle_recording_start` (after `set_recording`) and `handle_recording_stop` (after `set_recording`). This propagates the active session to the screencast stream, matching the pattern used everywhere else.

## Same class of bug as #1019

PR #1019 fixed downloads not working in the recording context by re-applying `Browser.setDownloadBehavior` to the new context. This PR fixes the screencast stream not following the recording context by calling `update_stream_client()`.

## Testing

Verified that without this fix:
- `record start` → live preview stays on old context
- CLI commands execute on new recording context  
- Extensions (loaded via `--load-extension`) appear to not work because the screencast shows the wrong context

With this fix:
- Live preview follows the recording context
- Extensions visible in the screencast
- `record stop` syncs the screencast back to the current page

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -3608,23 +3608,17 @@ async fn handle_recording_start(cmd: &Value, state: &mut DaemonState) -> Result<
             .await
             .ok();
 
-        // Create new browser context
-        let ctx_result = mgr
-            .client
-            .send_command_no_params("Target.createBrowserContext", None)
-            .await?;
-        let context_id = ctx_result
-            .get("browserContextId")
-            .and_then(|v| v.as_str())
-            .ok_or("Failed to get browserContextId")?
-            .to_string();
-
-        // Create page in new context
+        // Create a new TAB in the DEFAULT browser context (no browserContextId).
+        // Previously this used Target.createBrowserContext which creates an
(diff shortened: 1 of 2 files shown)


ctate:
Thanks for the fix @shtefcs\! One thing: please remove `cli/src/native/actions.rs.patch` — looks like a scratch file that shouldn't be committed.

PR #1059, opened 2026-03-28
fix: skip wait_for_lifecycle on same-document navigation

hechang27-sprt:
## Summary

- Fixes issue where `agent-browser open <url>` followed by another `open` or `wait` command would hang forever on SPAs with hash routing
- Chrome returns `loader_id: None` for same-document navigations (e.g., hash routing), and `Page.loadEventFired` never fires in these cases, causing `wait_for_lifecycle` to hang forever

## Test plan

- [x] Added regression test `e2e_navigate_same_url_twice_should_not_hang` that verifies the fix
- [x] Test fails without fix (hangs 25s), passes with fix (~0.5s)
- [x] All existing tests pass

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -488,8 +488,13 @@ impl BrowserManager {
             return Err(format!("Navigation failed: {}", error_text));
         }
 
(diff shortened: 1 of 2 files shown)


ctate:
Thanks for the fix @hechang27-sprt

PR #1064, opened 2026-03-28
fix: save_state captures cross-domain cookies and localStorage

### LINKED ISSUE #1060: save_state missing cross-domain cookies after Rust rewrite (regression from JS/Playwright version)
## Description

The Rust rewrite of `save_state` does not preserve the same behavior as the JS/Playwright version (v0.19.0), resulting in missing cross-domain cookies and localStorage when saving browser state.

## Root Cause

**JS version (v0.19.0)** used Playwright's `context.storageState()`, which internally calls `Network.getAllCookies` — returning cookies from **all domains** the browser has visited.

**Rust version (current)** uses `Network.getCookies` with no URL parameters (`cookies.rs:38`), which per the CDP spec only returns cookies **for the current page's URL**. This means cookies set by other domains during the session (e.g. SSO/CAS authentication redirects) are silently dropped.

Similarly, the Rust version only captures `localStorage` for the **current origin** via `Runtime.evaluate` (`state.rs:45-60`), whereas Playwright's `storageState()` collects storage across all known origins.

## Reproduction

1. Visit a page that redirects through a CAS/SSO login flow (e.g. `cas.example.com` → target site)
2. Complete login successfully
3. Run `save_state`
4. Compare the saved state between v0.19.0 (JS) and the current Rust version

The Rust version is missing:
- Cross-domain cookies (e.g. `TGC2`, `DT2`, `nonce`, `cookie.lang` on the CAS domain)
- Cross-origin `localStorage` entries (e.g. `_trackid_` under `https://cas.example.com`)

### Cookies
Replace `Network.getCookies` with `Network.getAllCookies` in `state.rs`. Note that `Network.getAllCookies` is already used elsewhere in the codebase (`actions.rs:3607` for context switching), so this should be a straightforward change:

```rust
// state.rs — current (broken)
let cookies = cookies::get_cookies(client, session_id, None).await?;

// suggested fix — use getAllCookies
let result = client
    .send_command_no_params("Network.getAllCookies", Some(session_id))
    .await?;
```

### localStorage
Consider iterating over all visited origins to collect their localStorage, matching Playwright's `storageState()` behavior. This may require leveraging CDP's `Target` or `Storage` domain commands.

jin-2-kakaoent:
## Summary

Fixes #1060 — `save_state` now captures cookies and localStorage from **all visited origins**, not just the current page.

### Root Cause

The Rust rewrite of `save_state` used:
- `Network.getCookies` (returns cookies for current URL only) instead of `Network.getAllCookies` (all domains)
- `Runtime.evaluate` on the current page only, missing localStorage from previously visited origins

This caused SSO/CAS authentication cookies and cross-origin localStorage to be silently dropped.

### How It Works

1. **Cookies**: `Network.getAllCookies` returns cookies from all domains in the browser
2. **localStorage**: Origins are collected from two sources:
   - `BrowserManager.visited_origins` — accumulated during `navigate()` calls
   - `Page.getFrameTree` — current page's frames (iframes, etc.)
3. Current page's localStorage is collected directly via `Runtime.evaluate`
4. Remaining origins are collected via a **temporary CDP target**:
   - `Target.createTarget("about:blank")` — same browser context, shares localStorage
   - `Fetch.enable` with wildcard pattern — intercepts all requests
   - For each origin: navigate → fulfill requests with `<html></html>` → `Page.loadEventFired` → evaluate localStorage
   - `Target.closeTarget` — always runs (even on error)

### Reproduction

**Before (broken)**:
```bash
agent-browser open "https://httpbin.org/cookies/set/sso_cookie/auth_token" --headless
agent-browser eval "localStorage.setItem('sso_key', 'sso_value'); 'ok'"
agent-browser open "https://example.com"
agent-browser eval "document.cookie = 'app_cookie=value; path=/'; localStorage.setItem('app_key', 'app_value'); 'ok'"
agent-browser state save /tmp/state.json
cat /tmp/state.json
# ❌ Only example.com cookies and localStorage — httpbin.org data missing
```

**After (fixed)**:
```bash
# Same steps as above
cat /tmp/state.json
# ✅ Both domains' cookies AND localStorage present
```

**Saved state output after fix**:
```json
{
  "cookies": [
    { "name": "sso_cookie", "domain": "httpbin.org", ... },
    { "name": "app_cookie", "domain": "example.com", ... }
  ],
  "origins": [
    { "origin": "https://example.com", "localStorage": [{"name": "app_key", ...}] },
    { "origin": "https://httpbin.org", "localStorage": [{"name": "sso_key", ...}] }
  ]
}
```

## Test Plan

- [x] `cargo test` — 542 unit tests pass
- [x] `e2e_save_state_cross_domain` — verifies both cookies and localStorage from two different domains
- [x] `e2e_state_management` — existing state test still passes (no regression)
- [x] Manual verification with release binary
- [x] `cargo fmt` and `cargo clippy` clean

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -1960,6 +1960,7 @@ async fn handle_close(state: &mut DaemonState) -> Result<Value, String> {
                     None,
                     Some(session_name.as_str()),
                     &state.session_id,
+                    mgr.visited_origins(),
                 )
                 .await;
             }
@@ -2996,6 +2997,7 @@ async fn handle_state_save(cmd: &Value, state: &DaemonState) -> Result<Value, St
         path,
         state.session_name.as_deref(),
         &state.session_id,
+        mgr.visited_origins(),
     )
     .await?;
 
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -199,6 +199,8 @@ pub struct BrowserManager {
     default_timeout_ms: u64,
     /// Stored download path from launch options, re-applied to new contexts (e.g., recording)
     pub download_path: Option<String>,
+    /// Origins visited during this session, used by save_state to collect cross-origin localStorage.
+    visited_origins: HashSet<String>,
 }
 
 const LIGHTPANDA_CDP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
@@ -268,6 +270,7 @@ impl BrowserManager {
                 active_page_index: 0,
                 default_timeout_ms: 25_000,
                 download_path: download_path.clone(),
+                visited_origins: HashSet::new(),
             };
             manager.discover_and_attach_targets().await?;
             manager
@@ -333,6 +336,7 @@ impl BrowserManager {
             active_page_index: 0,
             default_timeout_ms: 10_000,
             download_path: None, // CDP connections don't have a launch-time download path
+            visited_origins: HashSet::new(),
         };
 
         manager.discover_and_attach_targets().await?;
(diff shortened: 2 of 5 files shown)


ctate:
Thanks for the fix @jin-2-kakaoent!

PR #1067, opened 2026-03-28
feat: add dashboard host support

chethanuk:
## Summary
- add configurable host binding for the dashboard server so it can listen on a non-localhost interface when needed
- pass the selected host through the CLI dashboard startup path instead of always binding to 127.0.0.1
- update the dashboard frontend to derive API URLs from the current host and protocol instead of assuming localhost
- use the exported dashboard port constant when deciding between same-origin session requests and cross-port requests

## Validation
- `cd cli && cargo test`
- `cd cli && cargo clippy`
- `cd packages/dashboard && pnpm build`


ThomasK33:
I think you'll also need to update: https://github.com/vercel-labs/agent-browser/blob/510ba728ede9d852fb738bbe622dc2db4fffeeea/packages/dashboard/src/store/stream.ts#L122

packages/dashboard/src/lib/exec.ts
```diff
@@ -1,4 +1,17 @@
...
+  if (typeof window === "undefined") {
+    return `http://127.0.0.1:${DASHBOARD_PORT}`;
+  }
+
+  const { protocol, hostname, port } = window.location;
+  if (!port || port === String(DASHBOARD_PORT)) {
+    return `${protocol}//${hostname}:${DASHBOARD_PORT}`;
+  }
+
+  return `${protocol}//${hostname}:${DASHBOARD_PORT}`;
```
ThomasK33: Isn't this going to always return `${protocol}//${hostname}:${DASHBOARD_PORT}`?

akarshghale:
Was waiting for someone to make this pull request! Would help a lot over hosted setting.

Hope this is reviewed and merged soon. Was wondering if adding base-path could also be considered.

packages/dashboard/src/lib/exec.ts
chethanuk: Fixed in 95fdecd. I removed the redundant branch so `getDashboardBaseUrl()` now directly derives `${protocol}//${hostname}:4848`, and updated the stream websocket path to use the selected host instead of `localhost`.

chethanuk:
Addressed the remaining review feedback in 95fdecd. The dashboard frontend no longer keeps the redundant `getDashboardBaseUrl()` branch, and `packages/dashboard/src/store/stream.ts` now derives its websocket URL from the active host instead of hardcoding `localhost`.

dchenk:
I need this too! I haven't reviewed the changes thoroughly myself but I can see that this PR is directionally right.

PR #1073, opened 2026-03-29
fix: scroll element into view before click, hover, tap, and drag

### LINKED ISSUE #1044: Clicking an invisible element with agent-browser seems to have no effect, yet the command doesn't throw an error？
Here is my scenario: When interacting with page elements, the snapshot command lists all elements, but some are not visible in the Chrome window until I scroll. When I run agent-browser click @e3 on such an element, it completes without error, but nothing happens on the page.

      I tried using agent-browser set viewport 2400 2000 to enlarge the browser window, but it only scaled up the page content (zoomed in) instead of actually expanding the window size.

Is it possible to make the click command fail if the element is not visible? Alternatively, is there a command to maximize the Chrome window?

jin-2-kakaoent:
Closes #1044

## Summary

- Automatically scroll elements into view before `click`, `hover`, `tap`, and `drag` using CDP `DOM.scrollIntoViewIfNeeded` with JS fallback
- Return an error when an element cannot be scrolled into the viewport (eliminates silent failure)
- Detect detached (stale) nodes lazily via `isConnected` check inside the existing `getBoundingClientRect` JS call — zero extra CDP round-trips on the happy path
- Retry with a fresh accessibility tree lookup when a detached node is detected, preserving the existing stale-ref fallback behavior
- Fix drag coordinate invalidation: mouseDown at source before scrolling to target, preventing viewport-relative coords from going stale

## Root cause

Two issues combined to produce the silent failure:

**1. No auto-scroll before interactions**

`snapshot` uses `Accessibility.getFullAXTree` which returns all elements regardless of viewport position. When interacting with an off-screen element, `Input.dispatchMouseEvent` / `Input.dispatchTouchEvent` was dispatched to coordinates outside the viewport — CDP doesn't error on this, so the interaction silently had no effect.

**2. Coordinate system mismatch in `resolve_element_center`**

`DOM.getBoxModel` returns page-absolute coordinates while `Input.dispatchMouseEvent` expects viewport-relative coordinates. After scrolling, these diverge and interactions miss the target.

## Changes

- `get_center_and_viewport`: uses `getBoundingClientRect` for viewport-relative coordinates, returns `CenterResult::Found` or `CenterResult::Detached`
- `scroll_into_view_if_needed`: CDP `DOM.scrollIntoViewIfNeeded` with JS `scrollIntoView` fallback (errors propagated)
- `assert_in_viewport`: validates element is within viewport after scrolling
- `resolve_element_object_id_fresh`: skips cached `backend_node_id` for retry after detached node detection
- `resolve_scroll_and_center`: shared helper for click/hover/tap/drag — resolve, scroll, get center, retry on detach
- `handle_drag`: reordered to mouseDown at source before scrolling to target, preventing source coordinate invalidation
- Removed dead code: `resolve_element_center`, `box_model_center`

## Test plan

- [x] E2E: `e2e_offscreen_scroll_before_interactions` — click, hover, click-by-ref, tap, and drag on off-screen elements in a single unified test
- [x] E2E: `e2e_click_stale_ref_falls_back_to_role_name` — stale ref fallback still works
- [x] Manual: off-screen button clicked, hovered, tapped, and dragged via CLI — all verified
- [x] `cargo fmt` + `cargo clippy` clean
- [x] All E2E tests pass


ctate:
@jin-2-kakaoent LGTM! One minor follow-up to consider: `tap_touch` and `drag` still use the old `resolve_element_center` path without auto-scroll, so they'd benefit from the same treatment.

jin-2-kakaoent:
@ctate Thanks for the feedback! I've applied `resolve_scroll_and_center` (auto-scroll + detached node retry) to both `tap_touch` and `handle_drag`.

All e2e + manual tests pass.

Clarkkkk:
Really need this! Cost me an hour to debug the issue.

mutewinter:
Agreed on this being essential. I just ran into this issue and made multiple changes trying to work around it, without realizing it was a fundamental difference between agent browser and Playwright behavior.

jin-2-kakaoent:
@ctate Could you take a look when you have a chance?

ctate:
@jin-2-kakaoent Thanks for your patience. Implementation looks great. Can you please sign your commits before merging?

hyunjinee:
@ctate Sorry for the delay! 

All commits are now signed and you can confirm the Verified badges on the Commits tab:
https://github.com/vercel-labs/agent-browser/pull/1073/commits
I also synced the branch with the latest `upstream/main`, so the PR is mergeable and ready whenever you are.
Thanks again for the review.

mutewinter:
@ctate any chance we can get this merged now that the commits are signed? Our agents keep hitting this issue (we've got special guidance to help them avoid it as much as possible, but it'd be so much better encoded into the library).

PR #1077, opened 2026-03-30
fix: implement glob pattern matching for wait --url and route matching

### LINKED ISSUE #1061: wait --url glob pattern matching broken after Rust rewrite
## Bug Description

The `wait --url` command does not support glob patterns, despite the documentation and help text showing glob examples. This is a regression introduced during the Native Rust rewrite (f341c0a).

## Current Behavior

`wait --url` uses JavaScript `String.includes()` for matching, which treats glob characters (`*`, `**`) as literal strings. This means any pattern containing wildcards will never match.

```bash
agent-browser open https://github.com/settings/profile --headed
# Browser redirects to login page, user logs in, page navigates back

agent-browser wait --url "**/settings/profile"
# BUG: never resolves, times out
```

The built-in help text (`agent-browser wait --help`) shows glob examples that don't actually work:

```
  agent-browser wait --url "**/dashboard"
```

## Expected Behavior

Glob patterns should work as they did in the previous TypeScript implementation, which used Playwright's `page.waitForURL()` with native glob support:

```ts
// Old implementation (src/actions.ts)
await page.waitForURL(command.url, { timeout: command.timeout });
```

Playwright's matching semantics:
- No wildcards → exact match
- `*` → matches any characters except `/`
- `**` → matches any characters including `/`
- `{a,b}` → alternation

## Root Cause

In the Rust rewrite, `wait_for_url` (cli/src/native/actions.rs) was implemented as:

```rust
let check_fn = format!(
    "location.href.includes({})",
    serde_json::to_string(pattern).unwrap_or_default()
);
```

This passes the raw glob pattern to JS `String.includes()`, which does literal substring matching only.

The same issue also affects:
- **Route matching** (simplified glob logic that only handles a single `*` split into prefix/suffix)
- **`response-body` URL matching** (uses `String::contains()`)

## Environment

- agent-browser version: 0.23.0
- Introduced in: f341c0a (Native Rust rewrite of agent-browser daemon)
- Previously worked in: v0.15.x and earlier (TypeScript daemon using Playwright)

tmchow:
## Summary

Replaces literal string matching with Playwright-compatible glob semantics in `wait --url`, route matching, and `responsebody` URL matching.

## Changes

Three code paths in `cli/src/native/actions.rs` used string operations (`includes()`, `contains()`, `starts_with/ends_with` split) instead of glob matching. After the Rust rewrite (f341c0a), patterns like `**/dashboard` were treated as literal strings.

This PR adds a `glob_to_regex()` helper that converts glob patterns to regex:
- `*` matches any characters except `/`
- `**` matches any characters including `/`
- `?` matches a single character
- Regex metacharacters are escaped

Fixed locations:
- `wait_for_url` (line ~2913): `location.href.includes()` -> `new RegExp(glob_to_regex(...)).test(location.href)`
- Route matching (line ~6287): simplified `split('*')` logic -> `glob_matches()`
- `handle_responsebody` (line ~5585): `resp_url.contains()` -> `glob_matches()`

No new crate dependencies beyond `regex` (added to `cli/Cargo.toml`). 6 unit tests cover exact match, escaping, `*`, `**`, and `?` patterns.

Fixes #1061

### DIFF
--- a/.changeset/fix-glob-matching.md
+++ b/.changeset/fix-glob-matching.md
@@ -1,0 +1,4 @@
+---
+"@anthropic-ai/agent-browser-cli": patch
+---
+fix: implement glob pattern matching for wait --url and route matching
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -58,6 +58,7 @@ dependencies = [
  "hmac",
  "image",
  "libc",
+ "regex",
  "reqwest",
  "rust-embed",
  "serde",
@@ -74,6 +75,15 @@ dependencies = [
  "uuid",
  "windows-sys 0.52.0",
  "zip",
+]
+
+[[package]]
+name = "aho-corasick"
+version = "1.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+dependencies = [
+ "memchr",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


tmchow:
Rebased onto main in e784d19 to clear conflicts. Two conflicts resolved: `cli/Cargo.toml` was a clean union of dependencies (both branches added new crates), and `cli/src/native/actions.rs` route-matching now uses `glob_matches(...)` to feed the new `url_matches` variable that main's `resource_types` filter expects. Build, clippy, and the 6 glob unit tests pass locally.

PR #1080, opened 2026-03-30
fix: prevent CDP connect hang on browsers with discarded/frozen tabs

### LINKED ISSUE #1036: Bug: CDP connect hangs on browsers with discarded/frozen tabs (Memory Saver)
## Description

`agent-browser --cdp <port>` hangs indefinitely when connecting to a browser that has discarded (sleeping/frozen) tabs. This is extremely common — browsers aggressively discard inactive tabs to save memory via features like Chrome's "Memory Saver" or Brave's equivalent.

## Root Cause

`discover_and_attach_targets()` in `browser.rs` (line ~402) attaches **ALL** page targets at startup, then calls `enable_domains()` on the first session. `enable_domains()` sends `Page.enable`, `Runtime.enable`, and `Network.enable` — all of which require communication with the tab's **renderer process**.

Discarded tabs have no renderer process (it was killed to save memory). When `Page.enable` is sent to a discarded tab's session, **it hangs forever** — no response, no error, just silence. This blocks the daemon permanently.

## Reproduction

```bash
# 1. Open Brave/Chrome with 5+ tabs, wait for some to be discarded (or enable Memory Saver)
/Applications/Brave\ Browser.app/Contents/MacOS/Brave\ Browser --remote-debugging-port=9222

# 2. Try connecting
agent-browser --cdp 9222 get title  # ❌ hangs forever
```

**Workaround**: Disable Memory Saver and close/reload all discarded tabs. With all tabs active, it works perfectly.

## Verification

We tested individual tabs on a fresh Brave instance (126 tabs, 0 prior attachments):

```
# Active tabs → work
about:blank → Page.enable: OK (0.00s)
github.com/servicetitan/... → Page.enable: OK (0.00s)

# Discarded tabs → hang
x.com/jayneildalal/... → Page.enable: TIMEOUT
reactflow.dev/examples/... → Page.enable: TIMEOUT
youtube.com/watch?v=... → Page.enable: TIMEOUT
```

Out of 20 randomly sampled tabs, **14 timed out** (discarded) and only 6 responded. After disabling Memory Saver and closing discarded tabs (5 active tabs remaining): **5/5 OK, 0 timeouts**, and `agent-browser --cdp 9222 get title` works instantly.

## Why this affects almost everyone with CDP

Any user connecting to their daily browser (not a freshly launched automation instance) will have discarded tabs. The more tabs open, the higher the chance the first page target is discarded, causing an immediate hang.

## Environment

- agent-browser: 0.22.3
- Brave Browser: Chrome/146.0.7680.164
- Chrome: Chrome/146.0.7680.165
- macOS (Darwin 25.3.0, arm64)

## Suggested Fixes

1. **Only attach the active tab at startup** (best fix). The `/json` HTTP endpoint lists the most recently active page first. Attach only that one. Attach others lazily on `tab switch`. This eliminates the problem entirely and dramatically speeds up startup.

2. **Add a timeout to `enable_domains()`.** If `Page.enable` doesn't respond within 2-3s, the tab is likely discarded. Skip it and try the next one, or fall back to creating a new `about:blank` tab.

3. **Check tab lifecycle state before enabling domains.** Use heuristics or CDP signals to detect discarded tabs and skip them.

## Secondary issue: orphaned sessions after crash

A related but separate issue: when the daemon is killed ungracefully (`pkill`, crash), attached sessions become orphaned in Brave (Chrome cleans them up). Re-attaching to orphaned targets creates zombie sessions where all commands are silently dropped. This compounds the discarded tab issue. See comments for details.

ChunHao-dev:
## Summary

Fix CDP connect hanging indefinitely on browsers with discarded/frozen tabs (Chrome Memory Saver).

Fixes #1036

## Changes

- Add 3-second timeout to `enable_domains()` call during initial CDP connection in `discover_and_attach_targets()`
- If the first tab fails to respond (likely discarded/frozen with no renderer process), fall back to creating a new `about:blank` tab
- All existing tabs remain visible in `tab list` (attached but not enabled); they get enabled lazily on `tab switch`

## Test plan

- `cargo fmt --check` ✅
- `cargo clippy` ✅
- `cargo test` — 565 passed, 0 failed ✅
- Manual test: `chrome://discards` → Urgent Discard multiple tabs → `agent-browser --cdp 9222 get title`
  - Before: hangs indefinitely
  - After: completes in ~3s with fallback to new tab
 
## Design decision

Fallback to a new tab after the first timeout rather than trying all tabs because:

1. **`Target.getTargets` order is not sorted by activity** — there's no way to know which tab is more likely to be active, so trying them sequentially doesn't improve the odds
2. **Each timeout costs 3 seconds** — with many discarded tabs, sequential retries would add up (e.g. 10 discarded tabs = 30s wait)
3. **Creating a new tab is instant and guaranteed to work** — it always has a live renderer

A future improvement could use the `/json/list` HTTP endpoint (which returns tabs sorted by most recently active) to pick a better first candidate.

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -425,9 +425,54 @@ impl BrowserManager {
                 });
             }
 
+            // Try to enable domains on the first page with a timeout.
+            // Discarded/frozen tabs (Chrome Memory Saver) have no renderer,
+            // so Page.enable hangs forever. Fall back to a new tab if needed.
             self.active_page_index = 0;
             let session_id = self.pages[0].session_id.clone();
-            self.enable_domains(&session_id).await?;
+            match tokio::time::timeout(
+                std::time::Duration::from_secs(3),
+                self.enable_domains(&session_id),
+            )
+            .await
+            {
+                Ok(Ok(())) => {}
+                _ => {
+                    // First tab is likely discarded — create a new tab
+                    let result: CreateTargetResult = self
+                        .client
+                        .send_command_typed(
+                            "Target.createTarget",
+                            &CreateTargetParams {
+                                url: "about:blank".to_string(),
+                            },
+                            None,
+                        )
+                        .await?;
+
+                    let attach_result: AttachToTargetResult = self
+                        .client
+                        .send_command_typed(
+                            "Target.attachToTarget",
+                            &AttachToTargetParams {
+                                target_id: result.target_id.clone(),
+                                flatten: true,
+                            },
+                            None,
+                        )
+                        .await?;
+
+                    self.pages.push(PageInfo {
+                        target_id: result.target_id,
+                        session_id: attach_result.session_id.clone(),
+                        url: "about:blank".to_string(),
+                        title: String::new(),
+                        target_type: "page".to_string(),
+                    });
+                    self.active_page_index = self.pages.len() - 1;
+                    self.enable_domains(&attach_result.session_id).await?;
+                }
+            }
         }
 
         Ok(())


ChunHao-dev:
@ctate Friendly ping — this PR has been open for about a week.

It fixes #1036, where `--cdp` may hang when Chrome Memory Saver
discards tabs (fairly common in real-world usage).

The change is small:
- 3s timeout on `enable_domains()`
- fallback to a new tab

All tests pass. Happy to tweak if you'd prefer a different approach.

yangyaofei:
Hi, I encounter the same problem, and This PR — it successfully solves the initial hang on cold start with 80+ tabs. I tested it on my Chrome profile (79 tabs, many discarded by Memory Saver) and can confirm the 3-second timeout + fallback works.



However, I found several issues during testing:

1. Frozen tabs remain broken after tab switch

When switching to a tab whose enable_domains timed out during startup, the session is in a broken state. tab switch calls enable_domains again, but since the renderer is still discarded, it either times out again or silently fails. The → pointer doesn't update, and snapshot returns (empty page) even though the tab exists. This means frozen tabs are permanently inaccessible through agent-browser.

2. tab switch doesn't work correctly even for non-frozen tabs after fallback

After the initial fallback, switching between regular (non-frozen) tabs doesn't work properly — snapshot returns content from the wrong tab, and get url shows the previous tab's URL. This suggests the fallback path corrupts the active page index or session tracking.

3. Root cause is deeper than just the first tab

We tested with raw CDP (Python WebSocket, no agent-browser involved) and confirmed that after attaching to 76+ page targets, even Page.enable on a single active tab takes 30+ seconds. This is not limited to discarded tabs — the sheer number of attached sessions slows down Chrome's CDP message routing. The real issue is that discover_and_attach_targets attaches ALL targets upfront, which overwhelms Chrome regardless of which tab is active.

super-cache-money:
Another bump, because I also need this! 🙏

PR #1110, opened 2026-04-01
fix: idle timeout not respected due to sleep future reset in select loop

### LINKED ISSUE #1101: AGENT_BROWSER_IDLE_TIMEOUT_MS is not respected on Unix/macOS
## Summary

`AGENT_BROWSER_IDLE_TIMEOUT_MS` is accepted by the daemon, but on Unix/macOS the session can remain alive indefinitely instead of shutting down after the configured idle window.

## Version

- `agent-browser 0.23.4`

## Environment

- OS: macOS
- Engine: Chrome

## Reproduction

1. Run:

```bash
AGENT_BROWSER_IDLE_TIMEOUT_MS=3000 agent-browser --session repro-timeout open about:blank
sleep 5
agent-browser session list
```

2. Observe that the session is still present well past the configured timeout.

I reproduced this with a fresh named session and also checked the daemon process directly. With a 3 second timeout, the daemon was still alive after 5 seconds and still alive after 31 seconds.

## Expected behavior

The session should close automatically shortly after 3 seconds of inactivity.

## Actual behavior

The daemon and browser process stay alive indefinitely until explicitly closed.

## Notes

This looks distinct from #885, which is about the feature request for idle timeout itself. In this case the timeout exists, but the Unix/macOS implementation does not honor it.

The likely cause is in the Unix daemon loop in `cli/src/native/daemon.rs`: the idle sleep is recreated on each loop iteration, while the same loop wakes every 500 ms for maintenance, so the timer never reaches its deadline.

jin-2-kakaoent:
## Summary

Fixes #1101.

The idle `Sleep` future was recreated **inside** the `select!` loop on every iteration. On Unix/macOS, the 500 ms drain interval caused the future to be dropped and replaced before it could reach its deadline. On Windows, incoming connections (`accept`) caused the same timer reset. The daemon never shut down via idle timeout on either platform.

Moved the pinned `Sleep` future **outside** the loop so it survives loop restarts and only resets on actual command receipt (`reset_rx`). Applied the fix to both Unix and Windows `run_socket_server` implementations. Added a cross-platform regression test.

## Reproduction

```bash
AGENT_BROWSER_IDLE_TIMEOUT_MS=3000 agent-browser --session repro-timeout open about:blank
sleep 5
agent-browser session list
# Session still alive after 5s (expected: closed after ~3s)
```

## AS-IS (bug)

```
loop start
  └─ create new Sleep future (deadline = now + timeout)
       └─ tokio::select!
            ├─ drain tick (500ms) ← always wins
            │     └─ loop restart → Sleep future dropped
            │           └─ create new Sleep (deadline reset to now + timeout)
            │                 └─ drain tick wins again → drop → create → repeat forever
            │
            └─ idle timeout ← never reached
```

## TO-BE (fix)

```
create Sleep future outside loop + pin
  └─ loop start
       └─ tokio::select!
            ├─ drain tick (500ms)
            │     └─ loop restart, but Sleep is an outer variable → preserved
            │           └─ next select polls same Sleep (500ms remaining)
            │
            ├─ idle timeout (1000ms) ← fires normally → break → daemon exits
            │
            └─ reset_rx (command received)
                  └─ only here: create new Sleep (timer reset)
```

## Test plan

- [x] `cargo check` passes
- [x] `cargo fmt` / `cargo clippy` clean
- [x] `cargo test` — 569 passed, 0 failed
- [x] Manual repro: `AGENT_BROWSER_IDLE_TIMEOUT_MS=3000` session now closes after ~3 s of inactivity

### DIFF
--- a/cli/src/native/daemon.rs
+++ b/cli/src/native/daemon.rs
@@ -159,8 +159,7 @@ async fn run_socket_server(
     let mut drain_interval = tokio::time::interval(Duration::from_millis(500));
     drain_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
 
-    let idle_sleep = idle_timeout_ms
-        .map(|ms| tokio::time::sleep(Duration::from_millis(ms)));
+    let idle_sleep = idle_timeout_ms.map(|ms| tokio::time::sleep(Duration::from_millis(ms)));
     let mut idle_sleep_pin = idle_sleep.map(Box::pin);
 
     loop {
@@ -264,8 +263,7 @@ async fn run_socket_server(
     let (reset_tx, mut reset_rx) = mpsc::channel::<()>(64);
     let reset_tx = idle_timeout_ms.map(|_| Arc::new(reset_tx));
 
-    let idle_sleep = idle_timeout_ms
-        .map(|ms| tokio::time::sleep(Duration::from_millis(ms)));
+    let idle_sleep = idle_timeout_ms.map(|ms| tokio::time::sleep(Duration::from_millis(ms)));
     let mut idle_sleep_pin = idle_sleep.map(Box::pin);
 
     loop {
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the excellent contribution @jin-2-kakaoent

PR #1131, opened 2026-04-04
feat: reuse Chrome profile login state via --profile <name>

desenmeng:
## Summary

Adds the ability to reuse existing Chrome login state (cookies, sessions, local storage) by passing a Chrome profile **name** to the existing `--profile` flag. This provides the lowest-friction authentication path — zero manual setup, no state files to manage, no re-login required.

```bash
# List available Chrome profiles
agent-browser profiles

# Launch with your existing Chrome login state
agent-browser --profile Default open https://gmail.com
agent-browser --profile "Work" open https://app.example.com
```

## Motivation

The most common friction point for new users is authentication. The existing `--profile <path>`, `--session-name`, and `--state` approaches all require an initial manual login step. Many users already have active Chrome sessions with the sites they want to automate — this feature lets them reuse that state directly.

This came up frequently in my own usage: I wanted to automate tasks on sites where I was already logged in via Chrome, but had to re-authenticate every time because `--profile` only accepted directory paths.

## How it works

The `--profile` flag now accepts **either** a Chrome profile name (e.g., `Default`, `Profile 1`, `Work`) **or** a directory path (existing behavior, unchanged).

**Detection heuristic:** If the value contains no `/`, `\`, or `~` characters, it's treated as a profile name. Otherwise it's treated as a path. This is simple and covers all real-world cases — Chrome profile names are short identifiers like `Default` or `Profile 1`, while paths always contain separators.

**Profile name resolution** uses three-tier matching:
1. Exact directory name match (e.g., `Default`, `Profile 1`)
2. Case-insensitive display name match (e.g., `work` → `Profile 1` if its display name is "Work")
3. Case-insensitive directory name match (e.g., `default` → `Default`)

Ambiguous matches (multiple profiles with the same display name) produce a clear error listing all candidates.

**Profile copying:** The resolved Chrome profile is copied to a temp directory (read-only snapshot). This ensures:
- The user's original Chrome profile is **never modified**
- Chrome doesn't conflict with a running instance (the copy has its own lock files)
- The temp copy is automatically cleaned up when the browser closes (via `ChromeProcess::Drop`)

**What's copied:** All profile data except large non-auth directories (Cache, Code Cache, GPUCache, Service Worker, blob_storage, etc.) — this keeps the copy fast while preserving cookies, local storage, IndexedDB, and extension state.

**Keychain handling:** When launching with a copied Chrome profile, `--password-store=basic` and `--use-mock-keychain` flags are omitted so Chrome uses the real system keychain. This is necessary because copied profiles may reference system keychain entries for saved passwords.

### Core implementation (`cli/src/native/cdp/chrome.rs`)
- `is_chrome_profile_name()` — heuristic to distinguish profile names from paths
- `find_chrome_user_data_dir()` — finds the first valid Chrome user-data directory (supports Chrome, Chrome Canary, Chromium, Brave)
- `ChromeProfile` struct + `list_chrome_profiles()` — parses Chrome's `Local State` JSON to enumerate profiles
- `resolve_chrome_profile()` — three-tier name resolution with clear error messages
- `copy_chrome_profile()` + `copy_dir_recursive()` — best-effort profile copy with excluded directories
- `PROFILE_COPY_EXCLUDE_DIRS` — list of large non-auth directories to skip
- `LaunchOptions.use_real_keychain` — controls keychain-related Chrome flags
- Updated `launch_chrome()` to detect profile names, resolve, copy, and wire up cleanup
- Made `get_chrome_user_data_dirs()` and `LaunchOptions` public/cloneable as needed

### CLI (`cli/src/main.rs`)
- New `profiles` subcommand that lists available Chrome profiles (supports `--json` output)

### Help text (`cli/src/output.rs`)
- Updated `--profile` description to show `<name|path>`
- Added `profiles` command to help text and examples

### Documentation
- **README.md** — new "Chrome Profile Reuse" section, updated comparison table
- **docs/src/app/sessions/page.mdx** — new section with usage examples and details table
- **skills/agent-browser/SKILL.md** — added as authentication Option 2

### Tests (20 new tests)
- `is_chrome_profile_name` — simple names and path detection
- `list_chrome_profiles` — valid data, missing file, malformed JSON, missing info_cache
- `resolve_chrome_profile` — exact match, display name (case-insensitive), directory name (case-insensitive), not found, ambiguous
- `copy_chrome_profile` — full structure verification, missing source, missing Local State
- `copy_dir_recursive` — exclude directory filtering
- `build_chrome_args` — keychain flag behavior with `use_real_keychain` true/false/path

All tests use an RAII `TempDir` guard for automatic cleanup (no leaked temp dirs on panic).

## Platform support

- **macOS**: Chrome, Chrome Canary, Chromium, Brave (via `~/Library/Application Support/`)
- **Linux**: Chrome, Chromium, Brave (via `~/.config/`)
- **Windows**: Chrome, Chrome Canary, Chromium, Brave (via `%LOCALAPPDATA%/`)

> **Note:** On Windows, Chrome must be closed before using `--profile <name>` because Chrome locks some profile files while running.

## Backward compatibility

Fully backward compatible:
- `--profile /path/to/dir` continues to work as before (path detection is based on presence of `/`, `\`, or `~`)
- `AGENT_BROWSER_PROFILE` env var works with both names and paths
- No changes to the daemon protocol or existing commands
- Default `use_real_keychain: false` preserves existing keychain flag behavior for path-based profiles

## Test plan

- [x] `cargo check` passes
- [x] All 20 new tests pass (`cargo test chrome_profile`)
- [x] `cargo clippy` clean
- [ ] Manual test: `agent-browser profiles` lists Chrome profiles
- [ ] Manual test: `agent-browser profiles --json` outputs valid JSON
- [ ] Manual test: `agent-browser --profile Default open https://gmail.com` launches with existing cookies
- [ ] Manual test: `agent-browser --profile /tmp/test open https://example.com` still works (path mode)

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -118,10 +118,16 @@ fn get_pid_path(session: &str) -> PathBuf {
     get_socket_dir().join(format!("{}.pid", session))
 }
 
+fn get_version_path(session: &str) -> PathBuf {
+    get_socket_dir().join(format!("{}.version", session))
+}
+
 /// Clean up stale socket and PID files for a session
-fn cleanup_stale_files(session: &str) {
+pub fn cleanup_stale_files(session: &str) {
     let pid_path = get_pid_path(session);
     let _ = fs::remove_file(&pid_path);
+    let version_path = get_version_path(session);
+    let _ = fs::remove_file(&version_path);
     let stream_path = get_socket_dir().join(format!("{}.stream", session));
     let _ = fs::remove_file(&stream_path);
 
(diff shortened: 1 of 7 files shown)


ctate:
Thanks for this contribution, @desenmeng!

PR #1145, opened 2026-04-05
fix: rewrite getByRole to use CDP accessibility tree with ref-based element resolution

### LINKED ISSUE #1123: find role link generates wrong CSS selector — matches <link> instead of <a>
## Bug

`find role link click --name "X" --exact` fails to find `<a>` anchor elements because the generated CSS selector queries `<link>` (stylesheet) elements instead of `<a>` tags.

## Root cause

In `handle_getbyrole` ([cli/src/native/actions.rs](https://github.com/AhmedElbohworkaround/agent-browser/blob/main/cli/src/native/actions.rs)), the JavaScript query is:

```javascript
const els = document.querySelectorAll('[role="link"], link');
```

The CSS selector `link` matches `<link>` elements (used in `<head>` for stylesheets/resources), **not** `<a>` anchor elements. The ARIA role `link` maps to `<a>` tags.

This means `find role link` only finds elements with an explicit `role="link"` attribute and misses all regular `<a href="...">` links.

For comparison, `button` works correctly because `<button>` is both the ARIA implicit tag and the HTML element name. `link` is the ARIA role name, but the HTML element is `<a>`.

## Reproduction

```bash
# Start agent-browser and navigate to any page with a standard <a> link
agent-browser navigate https://example.com

# This FAILS — finds no element (or matches a <link> stylesheet in <head>)
agent-browser find role link click --name "More information..." --exact

# Workaround using find text works, but is less precise
agent-browser find text "More information..." click --exact
```

### Minimal HTML to reproduce

```html
<html>
<head>
  <link rel="stylesheet" href="style.css">  <!-- this is what the selector matches -->
</head>
<body>
  <table>
    <tr>
      <td>
        <a href="/page">Example Link</a>  <!-- this is what it SHOULD match -->
      </td>
    </tr>
  </table>
</body>
</html>
```

```bash
agent-browser find role link click --name "Example Link" --exact
# Expected: clicks the <a> element
# Actual: "No element found: getByRole('link', { name: 'Example Link', exact: true })"
```

## Expected behavior

The query should use `a` instead of `link`:

```javascript
const els = document.querySelectorAll('[role="link"], a');
```

More generally, `handle_getbyrole` should map ARIA roles to their implicit HTML tag, since several roles differ from their element name (e.g. `link` → `a`, `heading` → `h1`–`h6`, `textbox` → `input[type="text"], textarea`, `combobox` → `select`).

## Affected version

Latest (v0.24.0)

jin-2-kakaoent:
## Summary

Fixes #1123

`find role link` generated `querySelectorAll('[role="link"], link')` which matched `<link>` stylesheet elements instead of `<a>` anchor tags. This happened because ARIA role names were used directly as CSS tag selectors — `link` is the ARIA role, but the HTML element is `<a>`.

This PR replaces the JS-based DOM query with the CDP `Accessibility.getFullAXTree` API, where the browser engine correctly computes implicit ARIA roles per the [WAI-ARIA / HTML-AAM spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings). Matched elements are resolved via a temporary ref entry in the existing ref system, avoiding marker attribute DOM pollution entirely.

### Approach

1. Query the browser's accessibility tree via `Accessibility.getFullAXTree`
2. Find the matching AX node by role/name in Rust (`find_ax_node_by_role`)
3. Register a temporary ref with the `backendDOMNodeId` in the ref_map
4. `execute_subaction` resolves the element via the ref system directly

### Why CDP Accessibility Tree?

| Approach | Role judgment | Maintenance | Accuracy |
|----------|--------------|-------------|----------|
| CSS selector mapping (e.g. `link` → `a[href]`) | Our table | Manual updates | Incomplete — conditional roles hard to express |
| **CDP Accessibility Tree** ✅ | Browser engine | Automatic with Chrome updates | 100% — browser is the authority |

### Reproduction (before → after)

```bash
agent-browser navigate https://example.com

# BEFORE: ✗ "No element found: getByRole('link', { name: 'Learn more', exact: true })"
# AFTER:  ✓ Done (clicks the <a> element correctly)
agent-browser find role link click --name "Learn more" --exact
```

## Test plan

- [x] `cargo check` / `cargo fmt` / `cargo clippy` — all clean
- [x] `cargo test` — 597 passed, 0 failed (+ 4 new regression tests)
- [x] Manual verification on https://example.com — `find role link click --name "Learn more" --exact` succeeds
- [x] `find role link click` (no name) — finds first link
- [x] `find role link click --name "Learn"` (substring) — matches
- [x] `find role link click --name "Learn" --exact` — correctly rejects partial match
- [x] `find role button click --name "Google 검색"` — button role regression check

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -5229,31 +5229,35 @@ async fn handle_getbyrole(cmd: &Value, state: &mut DaemonState) -> Result<Value,
         )
         .await?;
 
-    let backend_node_id = find_ax_node_by_role(&ax_tree.nodes, role, name, exact)?;
+    let (backend_node_id, actual_name) = find_ax_node_by_role(&ax_tree.nodes, role, name, exact)?;
 
     // Register a temporary ref so execute_subaction can resolve the element
     // via backendNodeId directly — no marker attribute needed.
-    let temp_ref = format!("e{}", state.ref_map.next_ref_num());
+    let ref_num = state.ref_map.next_ref_num();
+    let temp_ref = format!("e{}", ref_num);
     state.ref_map.add_with_frame(
         temp_ref.clone(),
         Some(backend_node_id),
         role,
-        name.unwrap_or(""),
+        &actual_name,
         None,
         state.active_frame_id.as_deref(),
     );
+    state.ref_map.set_next_ref_num(ref_num + 1);
 
-    execute_subaction(cmd, state, &format!("@{}", temp_ref)).await
+    let result = execute_subaction(cmd, state, &format!("@{}", temp_ref)).await;
+    state.ref_map.remove(&temp_ref);
+    result
 }
 
 /// Search the accessibility tree for a node matching the given role and
-/// optional name. Returns the `backendDOMNodeId` of the first match.
+/// optional name. Returns `(backendDOMNodeId, actual_name)` of the first match.
 fn find_ax_node_by_role(
     nodes: &[super::cdp::types::AXNode],
     role: &str,
     name: Option<&str>,
     exact: bool,
-) -> Result<i64, String> {
+) -> Result<(i64, String), String> {
     for node in nodes {
         if node.ignored.unwrap_or(false) {
             continue;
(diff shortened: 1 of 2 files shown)


ctate:
Thank you for this important fix @jin-2-kakaoent

alahawash:
This was reverted by #1160

PR #1158, opened 2026-04-06
Add --ignore-default-args flag

### LINKED ISSUE #682: Feature request: expose ignoreDefaultArgs option to allow Brave Shields to work

ShivanKaul:
Fix https://github.com/vercel-labs/agent-browser/issues/682

My use-case is to control a Chromium browser with working component updates. Component updater is currently hard-disabled in agent-browser [code](https://github.com/vercel-labs/agent-browser/blob/fcb6615f5a1f89b7bb012921a8e489d7d1c8f7e8/cli/src/native/cdp/chrome.rs#L148). 

Playwright allows you to ignoreDefaultArgs, allowing users to remove specific hardcoded Chrome flags from build_chrome_args(). There should be a similar option for `agent-browser`.

This new flag is available via CLI flag, AGENT_BROWSER_IGNORE_DEFAULT_ARGS env var, and config file (`ignoreDefaultArgs`). Plumbed through daemon env forwarding and launch hash for relaunch detection.

Note: I used Claude Code to assist with the PR. Happy to work through any feedback. This shouldn't be a large change.

### DIFF
--- a/skill-data/core/references/commands.md
+++ b/skill-data/core/references/commands.md
@@ -297,21 +297,22 @@ agent-browser state load auth.json    # Restore saved state
 ## Global Options
 
 ```bash
-agent-browser --session <name> ...    # Isolated browser session
-agent-browser --json ...              # JSON output for parsing
-agent-browser --headed ...            # Show browser window (not headless)
-agent-browser --full ...              # Full page screenshot (-f)
-agent-browser --cdp <port> ...        # Connect via Chrome DevTools Protocol
-agent-browser -p <provider> ...       # Cloud browser provider (--provider)
-agent-browser --proxy <url> ...       # Use proxy server
-agent-browser --proxy-bypass <hosts>  # Hosts to bypass proxy
-agent-browser --headers <json> ...    # HTTP headers scoped to URL's origin
-agent-browser --executable-path <p>   # Custom browser executable
-agent-browser --extension <path> ...  # Load browser extension (repeatable)
-agent-browser --ignore-https-errors   # Ignore SSL certificate errors
-agent-browser --help                  # Show help (-h)
-agent-browser --version               # Show version (-V)
-agent-browser <command> --help        # Show detailed help for a command
+agent-browser --session <name> ...       # Isolated browser session
+agent-browser --json ...                 # JSON output for parsing
+agent-browser --headed ...               # Show browser window (not headless)
+agent-browser --full ...                 # Full page screenshot (-f)
+agent-browser --cdp <port> ...           # Connect via Chrome DevTools Protocol
+agent-browser -p <provider> ...          # Cloud browser provider (--provider)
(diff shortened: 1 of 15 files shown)


ctate:
Thanks for the contribution, @ShivanKaul!

Two small issues to fix before merge:

1. Run `cargo fmt` to fix the 3 formatting issues
2. Add the new flag to `skills/agent-browser/SKILL.md` per the project's documentation requirements in `AGENTS.md`

ShivanKaul:
@ctate thanks for the review! Addressed your comments.

ShivanKaul:
@ctate gentle ping for this! Any other thoughts?

skill-data/core/references/commands.md
```diff
@@ -248,21 +248,22 @@ agent-browser state load auth.json    # Restore saved state
...
-agent-browser --proxy <url> ...       # Use proxy server
-agent-browser --proxy-bypass <hosts>  # Hosts to bypass proxy
-agent-browser --headers <json> ...    # HTTP headers scoped to URL's origin
-agent-browser --executable-path <p>   # Custom browser executable
-agent-browser --extension <path> ...  # Load browser extension (repeatable)
-agent-browser --ignore-https-errors   # Ignore SSL certificate errors
-agent-browser --help                  # Show help (-h)
-agent-browser --version               # Show version (-V)
-agent-browser <command> --help        # Show detailed help for a command
+agent-browser --session <name> ...       # Isolated browser session
```
ShivanKaul, line 300: @ctate my addition to SKILL.md ran into a conflict after https://github.com/vercel-labs/agent-browser/pull/1225, I've added the `--ignore-default-args` flag to `skills/agent-browser/references/commands.md` instead. Let me know if that isn't the right approach.

ShivanKaul: (ended up requiring some reformatting of these sections, which makes it a bit hard to review, but really only adding a line)

PR #1178, opened 2026-04-07
fix: re-apply ignore_https_errors to recording context

### LINKED ISSUE #1172: Recording context does not inherit --ignore-https-errors / AGENT_BROWSER_IGNORE_HTTPS_ERRORS
## Problem

When `AGENT_BROWSER_IGNORE_HTTPS_ERRORS=true` is set, the main browser context correctly ignores certificate errors. However, the recording context created by `agent-browser record start` does **not** inherit this setting, causing `goto` to fail with `Operation timed out` on pages with self-signed HTTPS certificates.

## Steps to Reproduce

```bash
export AGENT_BROWSER_EXECUTABLE_PATH="/path/to/chrome-for-testing"
export AGENT_BROWSER_ARGS="--headless=new"
export AGENT_BROWSER_IGNORE_HTTPS_ERRORS=true
export AGENT_BROWSER_DEFAULT_TIMEOUT=120000

agent-browser close --all
agent-browser open "about:blank"         # Works
agent-browser record start ./test.webm   # Starts recording
agent-browser goto "https://self-signed.local:3000"  # Times out
```

The `goto` after `record start` fails because the recording's BrowserContext doesn't have `ignoreHTTPSErrors: true`.

Meanwhile, `agent-browser open "https://self-signed.local:3000"` (without recording) works fine.

## Expected Behavior

`AGENT_BROWSER_IGNORE_HTTPS_ERRORS=true` should apply to all browser contexts, including the recording context.

## Environment

- agent-browser 0.24.1
- macOS (Apple Silicon)
- Chrome for Testing 147.0.7727.15

## Context

This blocks headless visual verification of apps that use self-signed HTTPS certs for local development (common with tools like mkcert). ProofShot relies on agent-browser for recording and hits this when verifying local dev servers.

juniper929:
`record start` calls `Target.createBrowserContext` to create an isolated BrowserContext for the recording session.
  `Security.setIgnoreCertificateErrors` is session-scoped — it only applies to the session it was sent on and does
  not propagate to new contexts. As a result, the recording context always starts with default certificate validation enabled, regardless of `--ignore-https-errors` or `AGENT_BROWSER_IGNORE_HTTPS_ERRORS`.

  Symptoms reported in #1172:
  - Navigation to self-signed HTTPS URLs succeeds in the normal context
  - The same URL times out immediately after `record start`
  - Affects all local HTTPS development environments (e.g. mkcert)

  The fix stores `ignore_https_errors` in `BrowserManager` alongside `download_path`, and re-applies `Security.setIgnoreCertificateErrors` to the new session after context creation — matching the existing pattern for `Browser.setDownloadBehavior` re-application.

  Fixes #1172

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -3939,6 +3939,19 @@ async fn handle_recording_start(cmd: &Value, state: &mut DaemonState) -> Result<
                 .await;
         }
 
+        // Re-apply HTTPS error ignore to the recording context.
+        // Security.setIgnoreCertificateErrors at launch only applies to the session it was sent on.
+        if mgr.ignore_https_errors {
+            let _ = mgr
+                .client
+                .send_command(
+                    "Security.setIgnoreCertificateErrors",
+                    Some(json!({ "ignore": true })),
+                    Some(&new_session_id),
+                )
+                .await;
+        }
+
         // Transfer cookies to new context
         if let Some(ref cr) = cookies_result {
             if let Some(cookie_arr) = cr.get("cookies").and_then(|v| v.as_array()) {
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -202,6 +202,8 @@ pub struct BrowserManager {
     default_timeout_ms: u64,
     /// Stored download path from launch options, re-applied to new contexts (e.g., recording)
     pub download_path: Option<String>,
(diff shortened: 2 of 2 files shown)


ctate:
Thanks for fixing this @juniper929!

juniper929:
My pleasure! @ctate

Issue #1148, opened 2026-04-05
fix: PR_SET_PDEATHSIG kills Chrome in tokio multi-threaded runtime (v0.24.1 regression)


shtefcs:
## Bug Description

v0.24.1 introduced `prctl(PR_SET_PDEATHSIG, SIGKILL)` in Chrome's `pre_exec` hook (PR #1137) to prevent orphaned Chrome processes when the daemon is SIGKILL'd. However, this causes Chrome to be killed after ~7 seconds of idle time on **every page, every site**.

## Root Cause

`PR_SET_PDEATHSIG` tracks the **thread** that called `fork()`, not the **process**. This is documented in the `prctl(2)` man page:

> *"the 'parent' in this case is considered to be the **thread** that created this process"*

The agent-browser daemon uses tokio's multi-threaded runtime. The worker thread that spawns Chrome via `Command::spawn()` (which calls `fork()`) gets recycled by tokio after a few seconds of idle time. When the thread exits, the kernel sends SIGKILL to Chrome — even though the daemon process is still alive.

## Reproduction

```bash
# v0.24.0 — works fine
agent-browser open https://example.com
sleep 15
agent-browser get url  # → https://example.com/ ✅

# v0.24.1 — Chrome dies
agent-browser open https://example.com
sleep 10
agent-browser get url  # → about:blank ❌ (Chrome was killed and auto-relaunched)
```

Tested with:
- Zero env vars, zero custom args, pure v0.24.1 binary
- Happens on every site (example.com, wikipedia.org, etc.)
- Happens in both headless and headed mode
- Does NOT happen on v0.24.0 or v0.23.0

## Impact

- All pages navigate to `about:blank` after ~7-9 seconds
- Download workflows are completely broken (first download works, second fails because Chrome died between actions)
- Live preview streaming shows resolution changes (Chrome auto-relaunches with default viewport)
- Any automation that takes >7 seconds between commands fails

## Proposed Fix

Replace `PR_SET_PDEATHSIG` with a **sentinel process + keepalive pipe**:

1. Before spawning Chrome, create a pipe
2. After Chrome starts, fork a tiny sentinel process that:
   - Joins Chrome's process group (`setpgid`)
   - Blocks on reading the pipe's read end
3. Daemon keeps the pipe's write end open (process-scoped fd, shared by all threads)
4. When daemon dies (ANY reason including SIGKILL):
   - Kernel closes all daemon fds → pipe breaks
[...]

Issue #1127, opened 2026-04-03
Stale daemon after upgrade causes silent CDP failures (no version check)


dandaka:
## Bug

After upgrading `agent-browser` (e.g., from v0.10.0 to v0.22.3), the old daemon process keeps running. `ensure_daemon()` in `cli/src/connection.rs:313` only checks socket connectivity — not version. The new CLI reuses the old daemon, causing **silent failures**:

- `--cdp <port> get url` returns `about:blank` instead of the actual page
- `cookies get` returns empty instead of browser cookies
- No error or warning is shown

Users blame the new version ("v0.11 broke CDP connect") when the real problem is the stale daemon from the previous version.

## Reproduction

```bash
# 1. Install old version, use it (daemon starts)
bun install -g agent-browser@0.10.0
agent-browser --cdp 9222 get url  # works, spawns Node.js daemon

# 2. Upgrade without killing daemon
bun install -g agent-browser@latest
agent-browser --cdp 9222 get url  # silently broken — old daemon still running

# 3. Fix: kill daemon, retry
pkill -f agent-browser
agent-browser --cdp 9222 get url  # works again
```

## Root cause

`ensure_daemon()` (`cli/src/connection.rs:313`) calls `daemon_ready()` which only tests socket liveness (lines 169-184). If the socket responds, it returns `already_running: true` regardless of daemon version. This is especially problematic across the Node.js → Rust daemon transition (v0.15 → v0.16+).

## Suggested fix

**Minimal:** Write CLI version to a sidecar file (e.g., `/tmp/agent-browser-daemon-{session}-version`) when spawning daemon. In `ensure_daemon()`, compare sidecar version against `env!("CARGO_PKG_VERSION")`. On mismatch, kill and respawn.

**Robust:** Add a version field to the daemon handshake protocol so the CLI can query the running daemon's version directly.

## Related

This may be connected to #1086 ("Auto Connect inconsistency") — some reports there could be stale daemon artifacts rather than actual connect logic bugs.

PR #1189, opened 2026-04-07
fix(screenshot): use chrome-headless-shell on Chrome 147+ to avoid Page.captureScreenshot hang

syn-otani:
## Summary

On stock macOS (Darwin 25.2.0 arm64) with a fresh `agent-browser install`, `agent-browser screenshot` reliably hangs and then fails with:

```
✓ Example Domain
  https://example.com/
✗ CDP command timed out: Page.captureScreenshot
✗ Failed to read: Resource temporarily unavailable (os error 35) (after 5 retries - daemon may be busy or unresponsive)
```

Reproduces on `agent-browser` 0.24.0 / 0.24.1 / 0.25.3 with the Chrome downloaded by `agent-browser install` (`chrome-147.0.7727.50` and `.56`). `open` and `snapshot -i` work fine, only `screenshot` hangs.

I traced this to three compounding issues. Each commit addresses one of them.

## Root cause

1. **Chrome for Testing 147.x headless has a regression in `Page.captureScreenshot`.** Connecting to the same Chrome binary with a raw CDP WebSocket and sending `Page.captureScreenshot` (either `fromSurface: true` or `false`) also hangs — it's not an `agent-browser` serialization bug. Playwright does not repro because it bundles a different binary: `chrome-headless-shell`, from the same 147 release train, which ships the compositor fix that full Chrome for Testing is missing. Playwright also sets `--enable-features=CDPScreenshotNewSurface`.
2. **Client socket read timeout (30 s) == daemon CDP command timeout (30 s).** Whenever a CDP command gets close to the ceiling, the CLI's blocking `read_line` on the Unix socket races the daemon's response. The CLI loses, gets `EAGAIN` → `os error 35`, and because `is_transient_error` treats EAGAIN as retryable it burns through all 5 retries before reporting "daemon may be busy or unresponsive" — even though the daemon is still healthy and finishing the command.
3. **`Page.captureScreenshot` also hangs on occluded/background targets** on Chrome 147+ headless because the compositor stops producing frames. Playwright works around this with `Emulation.setFocusEmulationEnabled` + `Page.bringToFront` before each capture.

## Changes

Three small, independent commits:

### 1. `fix(install)`: download `chrome-headless-shell` alongside Chrome; prefer it for headless runs
- `install.rs` now pulls both `chrome` and `chrome-headless-shell` entries from the same `last-known-good-versions` manifest. Older manifests without a shell entry still work (Chrome-only install).
- Adds `find_installed_headless_shell()` / `chrome_headless_shell_in_dir()` mirroring the existing Chrome search, matching the flat layout of the `chrome-headless-shell-mac-arm64.zip` archive (and the equivalent Linux / Windows archives).
- `launch_chrome` prefers the shell whenever `headless` is on and no `--executable-path` / extensions are set. Headed runs and extension loads continue to use full Chrome unchanged.
- `--enable-features=NetworkService,NetworkServiceInProcess` is extended with `CDPScreenshotNewSurface` so the modern compositor capture path is used, matching Playwright's switch list.

### 2. `fix(timeout)`: decouple client socket read timeout from daemon CDP timeout
- `connection.rs`: client socket read timeout is now configurable via `AGENT_BROWSER_READ_TIMEOUT_SECS` (default **75 s**).
- `cdp/client.rs`: daemon CDP command timeout is now configurable via `AGENT_BROWSER_CDP_TIMEOUT_SECS` (default **60 s**, up from 30 s).
- The client read timeout is deliberately larger than the CDP timeout so the daemon always has a chance to serialise an error response before the client gives up. This eliminates the spurious `os error 35 (after 5 retries)` error on slow commands.

### 3. `fix(screenshot)`: prepare target and add `fromSurface` fallback before capture
- Adds `prepare_target_for_capture`: best-effort `Emulation.setFocusEmulationEnabled` + `Page.bringToFront` + a `Runtime.evaluate` forced layout. All three calls swallow errors so older browsers and CDP providers without the `Emulation` domain still succeed.
- Adds `AGENT_BROWSER_SCREENSHOT_FROM_SURFACE` opt-out (default `1`) and a one-shot retry with the opposite `fromSurface` value if the first call times out. This is a belt-and-braces guard in case a future Chrome build regresses the compositor path again.

## Verification

Before (v0.25.3, unpatched, macOS Darwin 25.2.0 arm64):
```
$ agent-browser --session s open https://example.com
✓ Example Domain
$ agent-browser --session s screenshot /tmp/out.png
✗ Failed to read: Resource temporarily unavailable (os error 35) (after 5 retries - daemon may be busy or unresponsive)
```

After (same machine, same Chrome download step):
```
$ agent-browser install
...
✓ Chrome 147.0.7727.56 installed successfully
✓ chrome-headless-shell 147.0.7727.56 installed successfully
$ agent-browser --session s open https://example.com
✓ Example Domain
$ agent-browser --session s screenshot /tmp/out.png
✓ Screenshot saved to /tmp/out.png    # < 1s, 16579 bytes
$ agent-browser --session s close
✓ Browser closed
```

`ps` confirms `launch_chrome` now picks up the headless shell automatically:
```
/Users/.../chrome-147.0.7727.56/chrome-headless-shell --type=renderer --headless=old ...
  --enable-features=CDPScreenshotNewSurface,NetworkService,NetworkServiceInProcess ...
```

## Tests
- `cargo test --release --bin agent-browser`: **622 passed; 0 failed; 52 ignored** on this branch.
- No test expectations needed to change — the existing `test_build_args_*` assertions are about `--headless` / feature-list presence only, not exact string equality.

## Compatibility
- **Existing installs** continue to boot: if only `chrome-*` exists under `~/.agent-browser/browsers/`, `find_installed_headless_shell()` returns `None` and `launch_chrome` falls back to full Chrome exactly as before. Re-running `agent-browser install` self-heals by downloading the shell into the same versioned directory.
- **Extensions** (`--extension`) still require full Chrome — `launch_chrome` skips the shell path when any extension is loaded.
- **Custom browsers** (`--executable-path`, cloud providers, iOS) are untouched.
- **Timeout env vars** are additive; unset values just pick the new safer defaults.

## Related reports
- #322 (OPEN): macOS `EAGAIN` crashes CLI instead of retrying — this PR removes the root cause of the `os error 35` path for the screenshot case specifically.
- #1148 (OPEN): `PR_SET_PDEATHSIG` kills Chrome in tokio multi-threaded runtime — orthogonal but same general "daemon hangs / CDP times out" bucket; not addressed here.
- #1127 (CLOSED): Stale daemon causes silent CDP failures — orthogonal, not addressed here.

Happy to split into three PRs if that's preferred, or to drop either of the two defensive commits (timeout decoupling, fromSurface fallback) and keep only the `chrome-headless-shell` switch, which is the fix that actually unblocks users.

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -635,21 +635,44 @@ fn is_transient_error(error: &str) -> bool {
 
 /// Socket read timeout for responses from the daemon.
 ///
-/// Must be strictly larger than the daemon-side CDP command timeout
-/// (see `native/cdp/client.rs::CDP_COMMAND_TIMEOUT_SECS`) so that the daemon
-/// always has time to materialise and return an error/success response before
-/// this client-side socket read gives up. Otherwise commands like
-/// `Page.captureScreenshot`, which can take close to the full CDP timeout on
-/// headless Chrome 147+, regularly trigger `EAGAIN (os error 35)` on macOS
-/// (and `os error 11` on Linux) even though the daemon is still healthy.
+/// Must be strictly larger than the worst-case time the daemon can spend
+/// on a single command so that the client doesn't give up before the
+/// daemon can materialise an error/success response. Otherwise commands
+/// like `Page.captureScreenshot`, which can take close to the full CDP
+/// timeout on headless Chrome 147+, regularly trigger `EAGAIN (os error
+/// 35)` on macOS (and `os error 11` on Linux) even though the daemon is
+/// still healthy.
+///
+/// The worst case today is `capture_screenshot_base64` in
+/// `native/screenshot.rs`, which on timeout retries once with the
+/// opposite `fromSurface` value. That gives a worst-case daemon wall
+/// time of `2 × cdp_command_timeout`. We add a safety margin on top
+/// (JSON serialisation of a large base64 payload, socket write-back,
+/// scheduler jitter) to avoid races at the exact boundary.
 ///
 /// Can be overridden via the `AGENT_BROWSER_READ_TIMEOUT_SECS` env var.
 fn socket_read_timeout_secs() -> u64 {
-    std::env::var("AGENT_BROWSER_READ_TIMEOUT_SECS")
+    if let Some(override_secs) = std::env::var("AGENT_BROWSER_READ_TIMEOUT_SECS")
         .ok()
         .and_then(|s| s.parse::<u64>().ok())
         .filter(|&n| n > 0)
-        .unwrap_or(75)
+    {
+        return override_secs;
+    }
+
+    // Keep this in sync with the CDP command timeout so the two knobs
+    // don't silently drift. We can't import the constant from
+    // `native::cdp::client` here because the binary is cfg-gated on
+    // per-command features; read the same env var instead.
+    let cdp_secs = std::env::var("AGENT_BROWSER_CDP_TIMEOUT_SECS")
+        .ok()
+        .and_then(|s| s.parse::<u64>().ok())
+        .filter(|&n| n > 0)
+        .unwrap_or(60);
+
+    // `2 × cdp` covers the screenshot fallback retry, `+ 15 s` covers
+    // serialisation / write-back overhead.
+    cdp_secs.saturating_mul(2).saturating_add(15)
 }
 
 fn send_command_once(cmd: &Value, session: &str) -> Result<Response, String> {


syn-otani:
Thanks for catching that — you're right, the 75 s default was below the worst-case daemon wall time. Fixed in 258a335.

The client socket read timeout is now computed as `2 × cdp_command_timeout + 15 s` so it always accommodates the screenshot `fromSurface` fallback retry (which can spend up to `2 × CDP timeout` in the daemon). With the default 60 s CDP timeout this comes out to **135 s**. Sourcing it from the same `AGENT_BROWSER_CDP_TIMEOUT_SECS` env var also means users who tune the CDP timeout up or down automatically get a matching socket timeout without having to set a second knob. `AGENT_BROWSER_READ_TIMEOUT_SECS` still overrides the computed default for anyone who wants an exact value.

Re-verified on Chrome 147.0.7727.56 headless-shell: `open` → `screenshot` → `close` still completes in < 1 s on the happy path, so this is purely a ceiling change for the pathological fallback case.

PR #1197, opened 2026-04-08
fix: prevent recording deadlocks when ffmpeg stalls

OfriHarlev:
## Summary
- prevent `record stop` from hanging forever when the screenshot capture loop wedges
- keep the streaming recording architecture and continuously drain `ffmpeg` stderr so encoder logging cannot deadlock the process
- add targeted timeout coverage for screenshot capture and recording shutdown
- add an ignored native e2e regression test that proves `video_stop` flushes a usable WebM instead of hanging

## Problem
`agent-browser record start` could succeed while `record stop` hung indefinitely and never flushed a WebM.

The root cause is in the streaming recording pipeline:
- the background task captures JPEG frames with `Page.captureScreenshot`
- those frames are streamed into a long-lived `ffmpeg` process over stdin
- `ffmpeg` writes progress and diagnostics to stderr continuously
- if stderr is piped but never drained, that pipe can fill up
- once stderr backpressure blocks `ffmpeg`, it stops consuming stdin frames and wedges the recording task

That leaves `record stop` waiting forever for the task to finish.

## Solution
This change keeps the existing real-time streaming design and hardens it in two ways:

1. It continues to pipe `ffmpeg` stderr, but drains it in a background task and keeps only a small tail for diagnostics.
2. It adds explicit timeouts around:
   - each `Page.captureScreenshot` call
   - the recording task shutdown path in `record stop`

If capture stalls, `record stop` now fails fast instead of hanging forever.

## Testing
I verified this change at three levels.

### Unit / focused validation
- `cargo fmt --manifest-path cli/Cargo.toml`
- `cargo test --manifest-path cli/Cargo.toml recording -- --nocapture`

### New regression coverage
Added ignored native e2e coverage in `cli/src/native/e2e_tests.rs` for:
- launch browser
- navigate
- `video_start`
- wait briefly
- `video_stop`
- assert the `.webm` exists, is non-trivial in size, and is accepted by `ffprobe`

Run with:
- `cargo test --manifest-path cli/Cargo.toml e2e_video_recording_stop_flushes_webm -- --ignored --test-threads=1`

### Before / after verification
Using that same e2e test:
- on this patched branch, it passes
- on clean `upstream/main`, the same test fails because `video_stop` hangs and hits the 5 second timeout

### End-to-end verification
I also verified this outside the test harness with:
- raw `agent-browser open -> record start -> sleep -> record stop`
- full ProofShot `start -> stop` flow using the patched binary

Both now produce valid WebM output instead of hanging during `record stop`.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -542,7 +542,7 @@ impl DaemonState {
         }
     }
 
-    /// Spawn a background task that polls screenshots and pipes them to ffmpeg.
+    /// Spawn a background task that polls screenshots and streams them into ffmpeg.
     async fn start_recording_task(
         &mut self,
         client: Arc<CdpClient>,
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -384,6 +384,69 @@ async fn e2e_snapshot_and_click_ref() {
 
 #[tokio::test]
 #[ignore]
+async fn e2e_video_recording_stop_flushes_webm() {
+    let mut state = DaemonState::new();
+
+    let resp = execute_command(
+        &json!({ "id": "1", "action": "launch", "headless": true }),
+        &mut state,
+    )
+    .await;
+    assert_success(&resp);
+
+    let resp = execute_command(
+        &json!({ "id": "2", "action": "navigate", "url": "https://example.com" }),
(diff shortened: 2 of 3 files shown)


mitch-fultz:
+1 on this

Issue #509, opened 2026-02-20
Screenshots with --annotate hang


iddogino:
When I run `npx agent-browser screenshot --annotate` on very simple pages (e.g. `example.com`) it works great. However, with pages that have a significant amount of interactive elements, it runs for a long while and times out.

I can also see the chrome process consuming a lot of CPU as it is going:
<img width="592" height="74" alt="Image" src="https://github.com/user-attachments/assets/23927771-3b97-4146-b17c-60ea94bc1204" />

Confirmed across Mac and Linux runtimes.

aksh-ahuja:
i am facing similar issue

digitall-it:
Adding a data point that **contradicts the reported reproduction condition** — on v0.15.1 / macOS (darwin 25.3.0), `--annotate` hangs on pages with **very few** interactive elements, not many.

**My reproduction:**
```bash
agent-browser --session s1 open https://example.com
agent-browser --session s1 wait --load networkidle
agent-browser --session s1 screenshot --annotate /tmp/test.png
# ↑ hangs indefinitely, must Ctrl+C
```
example.com has only ~2 interactive elements (one link, no form inputs).

**In contrast**, running the same command on pages with a rich accessibility tree (e.g. The Internet Heroku test suite, news sites) completed successfully in a few seconds.

This suggests the bug may not be strictly about the **quantity** of elements but rather something about edge cases at both ends of the spectrum — possibly a timeout or loop that requires at least N elements to exit, or a CPU-bound annotation pass that stalls when there's nothing to annotate.

**Environment:**
- agent-browser v0.15.1
- macOS darwin 25.3.0
- npm global install

PR #1200, opened 2026-04-09
fix: replace captureScreenshot polling with startScreencast for recording

shtefcs:
## Problem

`record start` uses `Page.captureScreenshot` in a 100ms polling loop (10 CDP commands/second). Each screenshot command competes with user commands (`eval`, `navigate`, `snapshot`) for Chrome's compositor, causing intermittent CDP timeouts during long recording sessions on JS-heavy pages.

We run agent-browser in production for cloud browser automation with recording enabled. On sites with heavy JavaScript (analytics, service workers, lazy-loading), sessions reliably fail after ~20 commands:

1. The recording task sends `Page.captureScreenshot` every 100ms
2. On heavy pages, each screenshot takes 100-500ms+ (compositor busy with layout/paint)
3. User commands queue behind the screenshot — the `ws_tx` Mutex serializes WebSocket writes, and Chrome's compositor serializes Page-domain work
4. After enough contention, a user command exceeds its timeout
5. Once one command times out, cascading failures follow

**Recording OFF = zero timeouts** across 38+ commands on the same site. **Recording ON = CDP timeouts** after ~20 commands, consistently reproducible.

## Fix

Replace `Page.captureScreenshot` polling with `Page.startScreencast` event subscription. Chrome pushes JPEG frames automatically as `Page.screencastFrame` events — zero CDP command round-trips for frame capture. The only CDP write is the lightweight `Page.screencastFrameAck`.

This is the same approach the stream server already uses in `stream/cdp_loop.rs` for live browser preview, where it runs successfully without any contention issues.

| | Before (captureScreenshot) | After (startScreencast) |
|---|---|---|
| CDP commands/sec from recording | **10** (one per frame) | **0** (frames arrive as events) |
| Compositor contention | Every 100ms | None |
| ws_tx Mutex contention | Every 100ms | Only for ack (lightweight) |
| Chrome controls frame rate | No (we poll) | Yes (everyNthFrame) |

Also includes two reliability improvements from PR #1197's diagnosis:
- **ffmpeg stderr drain**: background task keeps stderr flowing to prevent pipe backpressure from deadlocking ffmpeg (credit: @OfriHarlev's analysis in #1197)
- **Stop timeout**: `record stop` aborts after 5s if the capture task is wedged

### Automated
- `cargo test recording -- --nocapture`: 10 passed, 0 failed (all existing tests pass)

### Production validation
Tested in our cloud browser automation service (Railway, headed Chrome with Xvfb, proxy):

| Test | Recording | Method | Commands | CDP Timeouts | Page Frozen |
|------|-----------|--------|----------|-------------|-------------|
| Baseline | OFF | N/A | 38 | 0 | 0 |
| Before fix | ON | captureScreenshot | ~20 | 5-10 | Multiple, session dies |
| **After fix** | **ON** | **startScreencast** | **35+** | **0** | **0** |
| **After fix (back-to-back)** | **ON** | **startScreencast** | **35+** | **0** | **0** |

Both post-fix tests ran on the same server without redeploying between them, confirming this is not a fresh-server effect.

## Related issues and PRs

- **#1197** — Addresses ffmpeg stderr backpressure (a symptom of the same pipeline). Our change eliminates the root cause (captureScreenshot contention) while incorporating #1197's stderr drain as defense-in-depth.
- **#1189** — Documents Chrome 147+ `Page.captureScreenshot` hang regression. Switching to screencast sidesteps this entirely for recording.
- **#1151** — Our earlier PR to record the existing page instead of creating a new BrowserContext (different issue, complementary fix).
- **#509 / #842** — Previous reports of CDP saturation from too many concurrent commands, confirming that command volume on the WebSocket causes contention.

## Compatibility

- `Page.startScreencast` is supported by all Chromium-based browsers (Chrome, Edge, Brave). Not supported by LightPanda — the change falls back gracefully with an error message.
- Frame quality and resolution are configurable via the same parameters as before (JPEG, quality 80, 1920x1080).
- Chrome controls the frame rate based on compositor activity, which naturally produces ~10-30fps during active page changes — sufficient for automation recordings.

### DIFF
--- a/cli/src/native/recording.rs
+++ b/cli/src/native/recording.rs
@@ -3,14 +3,17 @@ use std::process::Stdio;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::io::AsyncWriteExt;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::sync::oneshot;
 
 use super::cdp::client::CdpClient;
-use super::cdp::types::{CaptureScreenshotParams, CaptureScreenshotResult};
 
-const CAPTURE_INTERVAL_MS: u64 = 100;
 const CAPTURE_FPS: u32 = 10;
+// Keep only the tail of ffmpeg stderr so diagnostics remain available without
+// allowing the pipe to block the encoder.
+const FFMPEG_STDERR_TAIL_BYTES: usize = 8 * 1024;
+// Bound stop latency so a wedged task does not hang the CLI forever.
+const RECORDING_STOP_TIMEOUT_MS: u64 = 5_000;
 
 pub struct RecordingState {
     pub active: bool,
@@ -120,8 +123,15 @@ fn build_ffmpeg_command(output_path: &str) -> tokio::process::Command {
     cmd
 }
 
-/// Spawn a background task that captures screenshots at a fixed interval
-/// and pipes them to ffmpeg in real-time.
+/// Spawn a background task that records browser frames via `Page.startScreencast`.
+///
+/// Chrome pushes frames automatically as CDP events — no polling, no
+/// `Page.captureScreenshot` commands competing with user commands for the
+/// compositor. The only CDP write is `Page.screencastFrameAck`.
+///
+/// The previous `Page.captureScreenshot` approach sent 10 CDP commands/second
+/// that contended with user commands (`eval`, `navigate`, `snapshot`) for
+/// Chrome's compositor, causing intermittent timeouts on JS-heavy pages.
 pub fn spawn_recording_task(
     client: Arc<CdpClient>,
     session_id: String,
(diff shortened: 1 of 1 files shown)


hiSandog:
The switch to screencast looks directionally right, but I think the ack path in `cli/src/native/recording.rs` should use the event session id rather than the closed-over `session_id`.

Right now `Page.screencastFrameAck` is sent with `Some(&session_id)`. In the sibling live-preview path (`cli/src/native/stream/cdp_loop.rs`), the ack uses `evt.session_id.as_deref()`. That is a bit safer because CDP can deliver frame events under a specific attached session, and reattach / target swaps can make the originally captured session id stale.

If the ack goes to the wrong session, Chrome stops sending further frames and recording effectively stalls until `record stop` hits the new timeout. I would mirror the `cdp_loop.rs` pattern here and ack against `evt.session_id` from the frame event itself.

shtefcs:
Good catch, you're right. The ack should use `evt.session_id` to handle session reattach and target swaps, exactly like `cdp_loop.rs` does. Fixed in 5672f1f.

The `Page.stopScreencast` still uses the closed-over `session_id` , that's intentional since we're stopping the screencast we started on that specific session. If the session became stale, the stop is a no-op (Chrome already stopped sending frames when the session changed).

OfriHarlev:
@shtefcs any reason you didn't also port over the integration test from #1197? 

I reworked it for your PR if you're open to including it. (Did not test this on your PR only tested my original test on my PR)

```
 #[tokio::test]
  #[ignore]
  async fn e2e_video_recording_stop_flushes_webm() {
      let mut state = DaemonState::new();

      let resp = execute_command(
          &json!({ "id": "1", "action": "launch", "headless": true }),
          &mut state,
      )
      .await;
      assert_success(&resp);

      let resp = execute_command(
          &json!({ "id": "2", "action": "navigate", "url": "https://example.com" }),
          &mut state,
      )
      .await;
      assert_success(&resp);

      let output_path = std::env::temp_dir()
          .join(format!("agent-browser-e2e-video-{}.webm", uuid::Uuid::new_v4()))
          .to_string_lossy()
          .to_string();

      let resp = execute_command(
          &json!({ "id": "3", "action": "video_start", "path": output_path }),
          &mut state,
      )
      .await;
      assert_success(&resp);

      // Force visual change so `Page.startScreencast` pushes multiple frames.
      // Without this, a static page delivers ~1 frame and the encoded webm
      // can be borderline small. This also keeps the test meaningful if the
      // capture mechanism changes again in the future.
      for i in 0..6 {
          let resp = execute_command(
              &json!({
                  "id": format!("motion-{}", i),
                  "action": "eval",
                  "expression": format!(
                      "document.body.style.background = 'hsl({}, 80%, 50%)'; \
                       document.body.innerText = 'frame {}';",
                      i * 60, i
                  ),
              }),
              &mut state,
          )
          .await;
          assert_success(&resp);
          tokio::time::sleep(tokio::time::Duration::from_millis(250)).await;
      }

      let resp = tokio::time::timeout(
          tokio::time::Duration::from_secs(5),
          execute_command(&json!({ "id": "4", "action": "video_stop" }), &mut state),
      )
      .await
      .expect("video_stop should not hang");
      assert_success(&resp);

      let metadata = std::fs::metadata(&output_path).expect("Recorded video should exist");
      assert!(
          metadata.len() > 4 * 1024,
          "Recorded video should be non-trivial size, got {} bytes",
          metadata.len()
      );

      let ffprobe = std::process::Command::new("ffprobe")
          .args(["-hide_banner", &output_path])
          .output()
          .expect("ffprobe should be installed for e2e recording validation");
      assert!(
          ffprobe.status.success(),
          "ffprobe should accept the recorded webm: {}",
          String::from_utf8_lossy(&ffprobe.stderr)
      );

      let resp = execute_command(&json!({ "id": "99", "action": "close" }), &mut state).await;
      assert_success(&resp);
      let _ = std::fs::remove_file(&output_path);
  }
  ```

PR #1208, opened 2026-04-10
fix: inherit viewport dimensions in recording context

### LINKED ISSUE #1207: [Feature Request] Allow videos to be recorded at different resolutions besides 1280x578
I noticed `agent-browser record start` has a fixed resolution of 1280x578 which is really short, and haven't found a way to change that. This seems to have fixed a past bug with custom video resolutions, but it doesn't seem to work anymore? https://github.com/vercel-labs/agent-browser/pull/718

So just requesting a way to change the video resolution for recordings.

Thanks for making this btw, I'm finding it to be a really neat and useful tool!

jin-2-kakaoent:
## Summary
- When `record start` creates a new browser context, it now re-applies the current viewport settings so the recording resolution matches what the user configured instead of falling back to the default 1280×720
- Both `set viewport` and `set device` paths now store viewport state on `DaemonState` for re-application
- Follows the same "store & re-apply to new context" pattern used by `download_path` and `ignore_https_errors`

## Before / After

| Configured viewport | Before (recording resolution) | After (recording resolution) |
|---|---|---|
| `set viewport 1920 1080` | 1280×578 | 1920×1080 |
| `set viewport 800 600` | 1280×578 | 800×600 |
| `set device "iphone 16 pro"` | 1280×578 | 1206×2622 |

## Test plan
- [x] Added e2e test `e2e_recording_inherits_viewport` — sets 800×600 viewport, starts recording, verifies `window.innerWidth`/`innerHeight` in the recording context
- [x] Manual testing with `ffprobe` across three resolutions (1920×1080, 800×600, iPhone 16 Pro)

Closes #1207

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -248,6 +248,9 @@ pub struct DaemonState {
     pub engine: String,
     /// Default timeout for wait operations, from AGENT_BROWSER_DEFAULT_TIMEOUT env var.
     pub default_timeout_ms: u64,
+    /// Last viewport settings (width, height, deviceScaleFactor, mobile),
+    /// re-applied to new contexts (e.g., recording).
+    pub viewport: Option<(i32, i32, f64, bool)>,
 }
 
 impl DaemonState {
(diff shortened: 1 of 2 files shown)


ctate:
Works gerat. Thanks @jin-2-kakaoent

glin:
This works great. Thanks!!

PR #1215, opened 2026-04-11
feat: add --background flag for non-intrusive browser tab operations

yangyaofei:
## Summary

related: #1247

Add `--background` flag to prevent agent-browser from stealing browser focus when creating or switching tabs. This is essential when using `--cdp` or `--auto-connect` to control a user's existing Chrome browser where the user is actively browsing.

## Problem

When connected to an existing Chrome via `--cdp` or `--auto-connect`, `tab new` and `tab switch` commands activate the target tab via CDP's `Target.createTarget` (which activates by default) and `Page.bringToFront`. This pulls the user's current tab away, making it impossible for an AI agent to operate in the background without disrupting the user.

Additionally, the `navigate` (`open`) command never called `Page.bringToFront`, so opened pages were not brought to the foreground even in normal (non-background) mode.

## Solution

Add a `--background` boolean flag (config key: `background`, env var: `AGENT_BROWSER_BACKGROUND`) that works as a **per-command flag** — each command can independently specify whether it should operate in background mode, regardless of how the daemon was started.

When `--background` is set:

- **`tab new`**: passes `background: true` to CDP `Target.createTarget`, creating the tab without activating it
- **`tab switch`**: skips `Page.bringToFront`, switching only the internal active page index
- **`navigate` (`open`)**: skips `Page.bringToFront`, navigating the page in the background
- **`click --new-tab`**: creates the new tab in the background

The agent can still fully interact with background tabs (snapshot, click, fill, etc.) since those operations don't require the tab to be in the foreground.

## Usage

```bash
# Per-command: open in foreground, then work in background
agent-browser --background tab new https://news.ycombinator.com
agent-browser --background snapshot -i
agent-browser --background click @e1
agent-browser --background tab 0

# Config file (agent-browser.json) — sets default for all commands
{"background": true}

# Environment variable — sets default for all commands
AGENT_BROWSER_BACKGROUND=1 agent-browser snapshot -i

# CLI flag overrides config/env for that specific command
agent-browser --background false tab new https://example.com


yangyaofei:
@ctate hi, can you give a glance of this, let agent-browser work in background.

PR #1218, opened 2026-04-12
fix: prefer DevToolsActivePort websocket path over HTTP discovery in --auto-connect

### LINKED ISSUE #1206: Is it normal to always have to allow twice when connecting to an existing Chrome instance via --auto-connect?
0.25.3

### LINKED ISSUE #1210: Prefer DevToolsActivePort websocket path first in --auto-connect to avoid duplicate remote-debugging prompts
`--auto-connect` looks like it does extra legacy discovery probes before using the exact websocket path from `DevToolsActivePort`.

On Chromium 146 with `chrome://inspect/#remote-debugging` enabled, this seems to cause two separate remote-debugging permission prompts:

1. one during legacy discovery/probing
2. another for the real websocket attach

Current flow appears to be:
- read `DevToolsActivePort`
- call legacy discovery (`/json/version`, `/json/list`, generic `/devtools/browser`)
- only then fall back to the exact websocket path from `DevToolsActivePort`

For UI remote-debugging mode, it would be better to:
- read `DevToolsActivePort`
- try `ws://127.0.0.1:<port><ws_path>` first
- only fall back to legacy `/json/*` discovery if that exact websocket attach fails

Why:
- in this mode, `/json/version` and `/json/list` may be gated or useless
- the exact websocket path from `DevToolsActivePort` is the real target
- using it first should reduce the browser-side permission flow to a single prompt instead of multiple prompts

This seems related to #1206, but this issue is specifically about the discovery order and a concrete fix.

jin-2-kakaoent:
Fixes #1210
Fixes #1206

## Summary

- Reverses the discovery order in `auto_connect_cdp()`: tries the exact WebSocket path from `DevToolsActivePort` first, falls back to legacy HTTP endpoints (`/json/version`, `/json/list`) only on failure
- Adds `verify_ws_endpoint()` to validate the WebSocket URL is a live CDP server before returning it
- Removes redundant synchronous `is_port_reachable()` TCP probe that blocked the async runtime

## Why

On Chrome M144+, `--auto-connect` triggered **two separate remote-debugging permission prompts** because legacy HTTP discovery probes (`/json/version`, `/json/list`) were attempted before using the exact WebSocket path already available in `DevToolsActivePort`. Each probe triggered its own permission flow.

## Before → After

| | Before (HTTP first) | After (ws_path first) |
|---|---|---|
| Discovery order | HTTP `/json/version` → `/json/list` → `/devtools/browser` → exact ws_path | exact ws_path → HTTP fallback |
| Permission prompts | 2 (one per discovery method) | 1 |
| Connection time | ~5s (HTTP timeouts) | ~0.2s |
| Stale URL handling | Returns unverified URL | Validates via `Browser.getVersion` handshake |

## Test plan

- [x] Unit tests: 3 new tests covering ws_path preference, HTTP fallback, and both-fail scenarios
- [x] All 625 existing tests pass
- [x] Real Chrome (M147) with `--remote-debugging-port`: `--auto-connect` connects in 0.2s
- [x] Stale `DevToolsActivePort` (dead port): clean error instead of hang
- [x] `--cdp` mode unaffected
- [x] `close` does not kill external Chrome

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -1863,7 +1863,7 @@ mod tests {
     }
 
     // -------------------------------------------------------------------
-    // auto_connect_cdp discovery-order tests (#1210, #1206, #1003, #1086)
+    // auto_connect_cdp discovery-order tests (#1210, #1206)
     // -------------------------------------------------------------------
 
     /// When DevToolsActivePort provides a ws_path and the port is reachable,
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -654,16 +654,7 @@ pub async fn auto_connect_cdp() -> Result<String, String> {
 
     for dir in &user_data_dirs {
         if let Some((port, ws_path)) = read_devtools_active_port(dir) {
-            // Try HTTP endpoint first (pre-M144)
-            if let Ok(ws_url) = discover_cdp_url("127.0.0.1", port, None).await {
-                return Ok(ws_url);
-            }
-            // M144+: direct WebSocket — verify the port is actually listening
-            // before returning, otherwise a stale DevToolsActivePort file
-            // (left behind after Chrome exits/crashes) produces a confusing
-            // "connection refused" error instead of falling through.
-            if is_port_reachable(port) {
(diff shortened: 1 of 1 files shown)


ctate:
Thank you so much for this fix @jin-2-kakaoent!

edxeth:
Amazing, thank you!

PR #1220, opened 2026-04-12
fix: preserve the active page when removing earlier tabs

### LINKED ISSUE #1219: `tab close` of an earlier tab can silently retarget the active page
## Summary

Closing or removing a tab that sits **before** the current active tab can silently move the session to the wrong page.

The current tab-removal bookkeeping only clamps `active_page_index` when it falls off the end of the list. It does **not** preserve the same logical active page when an earlier tab is removed, so the active index can shift to a different tab without any explicit tab switch.

## Reproduction

Given tabs `[A, B, C, D]` with `C` active:

1. Open or attach to four tabs so the session tracks them in that order
2. Switch to tab `2` (`C`)
3. Close tab `0` (`A`) with `tab close 0`

After the removal, the tab list becomes `[B, C, D]`.

## Expected Behavior

- The active page should remain `C`
- Internally, `active_page_index` should shift from `2` to `1` to keep pointing at the same page

## Actual Behavior

- `active_page_index` stays at `2` unless it is out of range
- The session now points at `D` instead of `C`
- Subsequent commands can run against the wrong page even though the user never switched tabs

## Root Cause

In `cli/src/native/browser.rs`, both explicit tab close and target-removal paths remove an entry from `pages`, but the active-index update only handles the out-of-range case. It does not decrement the active index when the removed page was located before the active one.

## Impact

- Commands after `tab close` can target the wrong page
- Session state can drift after external target destruction events as well
- This is a correctness issue in multi-tab workflows because the session silently changes scope

## Proposed Fix

- Centralize page-removal bookkeeping in a helper that adjusts `active_page_index` based on the removed index
- Decrement `active_page_index` when the removed page was before the active page
- Clamp the index when the active page itself or the last page is removed
- Add regression coverage for both explicit close and target-removal flows

officialasishkumar:
## Summary

- preserve the same logical active page when a tab before it is removed
- apply the same adjustment to both `tab close` and destroyed-target cleanup
- add regression tests for earlier-tab removal, later-tab removal, last-tab clamping, and the empty-page case

Closes #1219.

## Validation

- `cargo fmt --manifest-path cli/Cargo.toml`
- `cargo test --manifest-path cli/Cargo.toml active_page_index_after_removal`
- `cargo test --manifest-path cli/Cargo.toml tab_`
- manual CLI validation with `AGENT_BROWSER_ARGS=--no-sandbox`: open A/B/C/D, switch to C, run `tab close 0`, verify `get title` still returns `C` and the active tab is now index `1`
- `cargo clippy --manifest-path cli/Cargo.toml --all-targets -- -D warnings`

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -111,6 +111,26 @@ fn update_page_target_info_in_pages(pages: &mut [PageInfo], target: &TargetInfo)
     false
 }
 
+fn active_page_index_after_removal(
+    active_page_index: usize,
+    removed_index: usize,
+    remaining_pages: usize,
+) -> usize {
+    if remaining_pages == 0 {
+        return 0;
+    }
+
+    if removed_index < active_page_index {
+        return active_page_index - 1;
+    }
+
+    if active_page_index >= remaining_pages {
+        return remaining_pages - 1;
+    }
+
+    active_page_index
+}
+
 /// Converts common error messages into AI-friendly, actionable descriptions.
 pub fn to_ai_friendly_error(error: &str) -> String {
     let lower = error.to_lowercase();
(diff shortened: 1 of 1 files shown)


ctate:
Thanks for the fix @officialasishkumar

PR #1232, opened 2026-04-13
fix: print `get box` and `get styles` output in text mode

### LINKED ISSUE #1231: `get box` and `get styles` print only "Done" in text mode
## Bug

In the default (non-JSON) text output mode, `agent-browser get box <selector>` and `agent-browser get styles <selector>` print only `✓ Done` instead of the element's bounding box or computed styles. No data is shown.

Other `get` subcommands in the same family (`get text`, `get html`, `get value`, `get count`, `get attr`) all print their data correctly in text mode, so this is specific to `box` and `styles`.

## Steps to reproduce

```bash
agent-browser snapshot -i
agent-browser get html @e1      # prints the element's innerHTML — OK
agent-browser get box @e1       # prints: ✓ Done
agent-browser get styles @e1    # prints: ✓ Done
```

## Expected

`get box @e1` should print the bounding rect (`x`, `y`, `width`, `height`).
`get styles @e1` should print the computed CSS styles for the element.

## Actual

Both commands print `✓ Done` and no element data.

## Workaround

`--json` mode is unaffected — the underlying daemon returns the correct data over the IPC socket; only the text-mode output path is broken:

```bash
agent-browser --json get box @e1
# {"success":true,"data":{"x":10,"y":200,"width":320,"height":48}}

agent-browser --json get styles @e1
# {"success":true,"data":{"styles":{"color":"rgb(0,0,0)",...}}}
```

## Environment

- Platform: macOS (Darwin 25.3.0)
- Branch: main

nguyenyou:
## Summary

Both `get box <selector>` and `get styles <selector>` currently print `✓ Done` in text mode instead of the element's bounding rect / computed styles. The daemon returns the correct data over IPC — only the text-mode output path is broken. `--json` mode is unaffected.

Fixes #1231.

## Root cause

`output.rs` routes element-data responses by probing `data.get("<key>")` fields matching the `{ "<key>": ... }` shape returned by every `get`-style handler (`handle_innertext` → `{"text": ...}`, `handle_innerhtml` → `{"html": ...}`, `handle_inputvalue` → `{"value": ...}`, `handle_count` → `{"count": ...}`, etc.).

Two mismatches broke this for `box` and `styles`:

1. **`handle_boundingbox` did not wrap its return value.** It returned `{x, y, width, height}` raw, so the existing `data.get("box")` probe at `output.rs:481` never matched. This is the only `get`-style handler that didn't follow the "wrap in a named key" convention.
2. **No `data.get("styles")` probe existed.** `handle_styles` correctly returned `{"styles": {...}}`, but `output.rs` only had a `styles`-related block nested under `data.get("elements")` (for the `inspect` command, which has a different shape). So `get styles` responses fell all the way through the cascade.

In both cases the response reached `output.rs:~894` and was printed as a generic `✓ Done`.

## Changes

- **`cli/src/native/actions.rs`** — `handle_boundingbox` now wraps its result as `Ok(json!({ "box": bbox }))`, making it consistent with every other `get`-style handler. The existing `data.get("box")` probe now matches.
- **`cli/src/output.rs`** — added a top-level `data.get("styles")` probe that pretty-prints the computed-styles object, placed right after the bounding-box probe so element-inspection output stays grouped.

Diff is 2 files, 9 insertions, 1 deletion.

## Reproduction

```bash
agent-browser snapshot -i

# Before the fix:
agent-browser get box @e1       # ✓ Done
agent-browser get styles @e1    # ✓ Done

# After the fix:
agent-browser get box @e1
# {
#   "height": 38,
#   "width": 600,
#   "x": 392,
#   "y": 270
# }

agent-browser get styles @e1
# {
#   "color": "rgb(0, 0, 0)",
#   "display": "block",
#   ...
# }
```

## Test plan

- [x] `cargo check --manifest-path cli/Cargo.toml` passes cleanly
- [x] Manual repro on `https://example.com` after `bun run build:native`
- [x] `agent-browser --json get box @e1` and `--json get styles @e1` still work unchanged (JSON mode bypasses the output cascade)
- [ ] No new unit tests added — the change is a shape adjustment in a return value and a new branch in an output cascade; both are covered by existing integration-style manual testing. Happy to add tests if maintainers prefer.

## Non-goals

- No behavior change for `--json` mode.
- No new fields in the response — only a wrapping change to align with the existing convention.
- No changes to the CLI parser or the `get` subcommand surface.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -4258,7 +4258,7 @@ async fn handle_boundingbox(cmd: &Value, state: &mut DaemonState) -> Result<Valu
         &state.iframe_sessions,
     )
     .await?;
-    Ok(bbox)
+    Ok(json!({ "box": bbox }))
 }
 
 async fn handle_innertext(cmd: &Value, state: &mut DaemonState) -> Result<Value, String> {
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -485,6 +485,14 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
             );
             return;
         }
+        // Computed styles (from `get styles`)
+        if let Some(styles) = data.get("styles") {
+            println!(
+                "{}",
+                serde_json::to_string_pretty(styles).unwrap_or_default()
+            );
+            return;
+        }
         // Element styles
         if let Some(elements) = data.get("elements").and_then(|v| v.as_array()) {
             for (i, el) in elements.iter().enumerate() {


nguyenyou:
https://github.com/vercel-labs/agent-browser/pull/1233

PR #1240, opened 2026-04-14
feat: split domain filtering into navigationDomains and resourceDomains

### LINKED ISSUE #991: [Feature Request] - Separate navigation domain allowlist from sub-resource domain allowlist
## Problem

`allowedDomains` currently blocks both **agent-initiated navigation** and **page-initiated sub-resource requests** (scripts, fetch/XHR, WebSocket, EventSource). This means that to use `allowedDomains` with any real-world web app, you must allowlist every third-party domain the app depends on — defeating the purpose of the restriction.

A typical production web app depends on numerous third-party services (APIs, analytics, feature flags, error tracking, payment providers, CDNs, auth providers, etc.). To make the app functional with `allowedDomains`, you end up with 15+ domains in the allowlist. At that point, the security boundary is no longer "the agent can only interact with my app" — it's a broad network allowlist that's hard to audit and easy to get wrong.

## The distinction that matters

There are two fundamentally different threat models:

1. **Agent navigation** — the agent (via `open`, `click`, link following) navigates to a URL. This is the high-risk action: the agent choosing to visit an untrusted site or an internal IP.
2. **Page sub-resources** — JavaScript on an already-loaded trusted page makes fetch/XHR/WebSocket calls to its own backends and third-party services. This is low-risk: it's the app's own code running normally.

Currently both are gated by the same allowlist, forcing users to choose between a functional app and a tight security boundary.

## Proposed solution

Split `allowedDomains` into two controls:

| Control | Scope | Default |
|---------|-------|---------|
| `navigationDomains` | Agent-initiated navigation (`open`, link clicks, form submissions, redirects) | Allow all (current behavior) |
| `resourceDomains` | Page-initiated sub-resource requests (fetch, XHR, WebSocket, scripts, images) | Allow all (current behavior) |

The existing `allowedDomains` would continue to work as-is (restricting both), for backwards compatibility.

### Example

```json
{
  "navigationDomains": ["myapp.vercel.app", "*.myapp.com"],
  "resourceDomains": ["*"]
}
```

This says: "the agent can only navigate to my app, but once there, the app can load whatever it needs."

## Use case

We're deploying agent-browser for AI-assisted QA testing of a Next.js app on Vercel. Our security policy requires restricting where the agent can navigate, but the app depends on many third-party services for normal operation. With the current `allowedDomains`, we must either:

1. Allowlist all third-party domains (weakens the navigation boundary)
2. Don't use `allowedDomains` at all (no navigation restriction)
3. Accept a broken app (blocked sub-resources cause runtime errors)

Separating navigation from sub-resource restrictions would let us enforce a tight navigation boundary while keeping the app functional.

ariel-rahmane:
## Summary

Splits `allowedDomains` into two new independent controls so users can restrict agent navigation without needing to allowlist every third-party domain their app depends on:

- **`navigationDomains`** — restricts only agent-initiated navigation (open, click, form submit)
- **`resourceDomains`** — restricts only page-initiated sub-resources (fetch, XHR, scripts, WebSocket, images)

Both fall back to `allowedDomains` when not set, preserving full backwards compatibility. All three options support config file, CLI flags, and environment variables:

| Config key | CLI flag | Env var |
|------------|----------|---------|
| `allowedDomains` | `--allowed-domains` | `AGENT_BROWSER_ALLOWED_DOMAINS` |
| `navigationDomains` | `--navigation-domains` | `AGENT_BROWSER_NAVIGATION_DOMAINS` |
| `resourceDomains` | `--resource-domains` | `AGENT_BROWSER_RESOURCE_DOMAINS` |

### Example

Lock navigation to your app while allowing pages to load their own dependencies:

```json
{
  "navigationDomains": ["myapp.com", "*.myapp.com"]
}
```

Or via CLI:

```bash
agent-browser --navigation-domains "myapp.com,*.myapp.com" open https://myapp.com
```

### Additional fix

This PR also fixes a pre-existing bug where `allowedDomains` was serialized as a JSON array in the launch command (`json!(domains)`) but the daemon read it as a string via `.as_str()`. The array format silently returned `None`, causing the domain filter to never activate through the launch command path. Now correctly serialized as a comma-joined string.

Closes #991

## Test plan

- [x] All 642 existing tests pass
- [x] 7 new unit tests for split filter logic (navigation-only, resource-only, both, fallback, override, URL check, is_active)
- [ ] Verify `navigationDomains` restricts page navigation while sub-resources load freely
- [ ] Verify `allowedDomains` alone still restricts both navigation and resources (backwards compat)
- [ ] Verify CLI flags and env vars work for all three options


LiamLeonardo (approved):

ctate:
Thank you @ariel-rahmane - this is a great change. Can you please sign your commits?

ariel-rahmane:
> Thank you @ariel-rahmane - this is a great change. Can you please sign your commits?

Thanks for the review @ctate! Commits signed

PR #1241, opened 2026-04-15
fix: load storage state at launch when --state / AGENT_BROWSER_STATE is set

### LINKED ISSUE #1164: Linux v0.24.1: state save/--state and --session-name write state but do not restore cookies or localStorage
## Summary

On Linux with `agent-browser 0.24.1`, both of these persistence flows fail to restore state across restarts in my testing:

1. `--session-name <name>` auto-save/restore
2. `state save <file>` followed by `--state <file>` on relaunch

This is **not** just an empty-state-file problem. In my repro, the saved JSON contains the expected cookie and `localStorage` entry, but the next launch still comes back with neither restored.

This looks related to, but stronger than:

- #677 (`--session-name` writes empty state and restores nothing on macOS)
- #254 (`state load` / `--state` confusion)

## Environment

- agent-browser: `0.24.1`
- Install method: global via Bun (`/home/ubuntu/.bun/bin/agent-browser`)
- OS: Ubuntu Linux on x86_64 VPS
- Kernel: `6.17.0-19-generic`
- Shell: `zsh`

Note: on this host Chrome requires `--args "--no-sandbox"` to launch due the local Linux/AppArmor environment. That launch workaround is required here, but the persistence bug below reproduces cleanly once the browser is running.

## Repro 1: explicit state save + --state restore

```bash
agent-browser close --all >/dev/null 2>&1 || true
rm -f /tmp/ab-state-example.json

agent-browser --args "--no-sandbox" open https://example.com
agent-browser wait 500
agent-browser eval 'localStorage.setItem("persist_test","ok"); document.cookie="ab_cookie=1; max-age=3600; path=/"; JSON.stringify({ls: localStorage.getItem("persist_test"), cookie: document.cookie})'
# => {"ls":"ok","cookie":"ab_cookie=1"}

agent-browser state save /tmp/ab-state-example.json
agent-browser close

jq '.origins, .cookies' /tmp/ab-state-example.json
# => file contains the expected localStorage entry and cookie

agent-browser --state /tmp/ab-state-example.json --args "--no-sandbox" open https://example.com
agent-browser wait 500
agent-browser eval 'JSON.stringify({ls: localStorage.getItem("persist_test"), cookie: document.cookie})'
# => {"ls":null,"cookie":""}

agent-browser close
```

### Actual output from the save step

```json
{"ls":"ok","cookie":"ab_cookie=1"}
```

### Actual output after relaunch with `--state`

```json
{"ls":null,"cookie":""}
```

## Repro 2: --session-name auto persistence

```bash
agent-browser close --all >/dev/null 2>&1 || true
agent-browser state clear --all >/dev/null 2>&1 || true

agent-browser --session-name autoonly --args "--no-sandbox" open https://example.com
agent-browser wait 500
agent-browser eval 'localStorage.setItem("auto_probe","works"); document.cookie="auto_cookie=yes; path=/"; "SET"'
agent-browser close

find ~/.agent-browser/sessions -maxdepth 1 -type f -printf '%f\n' | sort
# => autoonly-default.json

agent-browser --session-name autoonly --args "--no-sandbox" open https://example.com
agent-browser wait 500
agent-browser eval 'JSON.stringify({ls: localStorage.getItem("auto_probe"), cookie: document.cookie})'
# => {"ls":null,"cookie":""}

agent-browser close
```

## Expected

If a state file contains a cookie and localStorage entry for the target origin, relaunching with that state should restore both.

Likewise, `--session-name` should auto-restore the previously saved state.

## Actual

- `state save` writes a plausible state file
- `--state <file>` relaunch still restores neither cookie nor localStorage
- `--session-name` writes a state file but relaunch restores neither cookie nor localStorage

## Why this matters

This blocks one of the main workflows advertised in the docs/README for authenticated agent automation:

- save auth/session once
- reuse it across later agent-browser runs

At the moment, on this machine, the write path appears to work but the restore path does not.

tomdale:
Fixes #1164 

The `--state <path>` flag and `AGENT_BROWSER_STATE` env var are documented as loading saved browser state (cookies + localStorage) at launch, but `load_state()` was never called after the browser started. The flag was fully plumbed through parsing, env propagation, and validation but the actual loading step was never wired up. This has been the case since the flag was introduced in the native Rust rewrite.

## Reproduction

```bash
# Save a cookie to a state file
agent-browser cookies set test_cookie hello_world
agent-browser state save /tmp/repro-state.json
agent-browser close

# Relaunch with --state and check cookies
agent-browser --state /tmp/repro-state.json open https://example.com
agent-browser --json cookies get
agent-browser close
```

**Before** (v0.25.4):
```json
{"success": true, "data": {"cookies": []}, "error": null}
```

**After** (this PR):
```json
{"success": true, "data": {"cookies": [{"name": "test_cookie", "value": "hello_world", ...}]}, "error": null}
```

## Changes

- **`handle_launch()`**: Call `state::load_state()` with the `storageState` path after browser launch
- **`auto_launch()`**: Extract `storage_state` from `LaunchOptions` before the struct is moved, then call `try_load_storage_state()` after `try_auto_restore_state()` in all four branches (CDP, auto-connect, provider, local Chrome)
- **New helper `try_load_storage_state()`**: Encapsulates the load pattern to avoid duplication across branches
- **4 new e2e tests** covering every state persistence path

## Test plan

- [x] `e2e_state_flag_restores_cookies` — `storageState` in explicit launch command (was broken, now passes)
- [x] `e2e_state_env_restores_cookies_on_auto_launch` — `AGENT_BROWSER_STATE` env via auto-launch (was broken, now passes)
- [x] `e2e_session_name_auto_restores_cookies` — `--session-name` auto-save/restore (regression guard)
- [x] `e2e_explicit_state_load_restores_cookies` — explicit `state load` command (baseline)
- [x] Full e2e suite: 57/57 passed, 0 regressions
- [x] Unit tests: 635/635 passed

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -1504,6 +1504,9 @@ async fn auto_launch(state: &mut DaemonState) -> Result<(), String> {
     }
     let engine = env::var("AGENT_BROWSER_ENGINE").ok();
 
+    // Extract storage_state before options is moved into BrowserManager::launch.
+    let storage_state_path = options.storage_state.clone();
+
     // Store proxy credentials for Fetch.authRequired handling
     let has_proxy_auth = options.proxy_username.is_some();
     if has_proxy_auth {
@@ -1527,6 +1530,7 @@ async fn auto_launch(state: &mut DaemonState) -> Result<(), String> {
         state.start_dialog_handler();
         state.update_stream_client().await;
         try_auto_restore_state(state).await;
+        try_load_storage_state(state, &storage_state_path).await;
         return Ok(());
     }
 
@@ -1538,6 +1542,7 @@ async fn auto_launch(state: &mut DaemonState) -> Result<(), String> {
         state.start_dialog_handler();
         state.update_stream_client().await;
         try_auto_restore_state(state).await;
+        try_load_storage_state(state, &storage_state_path).await;
         return Ok(());
     }
 
@@ -1572,6 +1577,7 @@ async fn auto_launch(state: &mut DaemonState) -> Result<(), String> {
                     state.update_stream_client().await;
                     write_provider_file(&state.session_id, &p);
                     try_auto_restore_state(state).await;
+                    try_load_storage_state(state, &storage_state_path).await;
                     return Ok(());
                 }
                 Err(e) => {
@@ -1604,6 +1610,7 @@ async fn auto_launch(state: &mut DaemonState) -> Result<(), String> {
     }
 
     try_auto_restore_state(state).await;
+    try_load_storage_state(state, &storage_state_path).await;
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


ctate:
Thanks @tomdale!

PR #1242, opened 2026-04-15
feat: add JSON Schema for agent-browser config files

michael-farah:
Adds agent-browser.schema.json describing all config options with types and descriptions. Enables IDE autocomplete and validation when referenced via $schema in agent-browser.json or
~/.agent-browser/config.json.

README and docs site updated to document the schema reference.


ctate:
Very helpful - thanks @michael-farah!

michael-farah:
Thank you, glad it was helpful @ctate!

PR #1264, opened 2026-04-18
fix: truncated snapshot refs over remote CDP

### LINKED ISSUE #1187: snapshot --json returns truncated accessibility tree over remote CDP WebSocket
## Description

When connecting to a remote browser via CDP WebSocket (`--cdp wss://...`), the `snapshot -i -c --json` command returns a significantly truncated accessibility tree compared to connecting to a local browser.

## Environment

- **agent-browser:** v0.25.3 (regression from v0.19.0 where this worked correctly)
- **Remote browser:** Headless Chrome via WSS (managed browser service)
- **Connection:** `--cdp wss://<host>/session/connect/<id>`
- **OS:** macOS (Darwin 24.6.0)

## Steps to Reproduce

1. Connect to a remote headless Chrome instance via CDP WebSocket
2. Navigate to a page with a Radix UI popover containing radios, combobox, tabs
3. Open the popover via `eval` (click the trigger)
4. Run `agent-browser --cdp wss://... snapshot -i -c --json`

## Expected

Full accessibility tree (~3000–5000 chars) including all interactive elements inside the popover (radio buttons, combobox, tabs, buttons).

## Actual

Truncated output (~715 chars) containing only page-level elements. Popover content (rendered via Radix UI portal at `document.body`) is missing from the accessibility tree.

## Workaround

Pin to `agent-browser` v0.19.0 which correctly captures the full accessibility tree over remote CDP WebSocket.

## Impact

Any workflow using snapshot output to extract interactive element refs over a remote CDP connection gets incomplete data. For example, filtering `data.refs` for `role == "radio"` returns 0 results from the truncated snapshot, causing automated steps to fail.

## Notes

- **Local browser** (`--cdp ws://localhost:...` or internal Chromium) works correctly with v0.25.3 — same page produces 4849+ char snapshots
- The `--json` output structure is correct (`{success, data: {origin, refs, snapshot}}`) — it's the `snapshot` string content that's truncated
- Non-`--json` output may also be affected (not tested separately)
- This is a regression: the same remote CDP setup worked correctly on v0.19.0

semimikoh:
## Summary

  Fixes #1187.

  Remote CDP sessions can return an incomplete accessibility tree for pages with dynamic portal content, which causes `snapshot -i -c
  --json` to miss interactive elements such as radios, comboboxes, tabs, and buttons.

  This PR extends the existing DOM scan used by snapshot generation so it also records semantic interactive elements. When a visible DOM
  control is missing from the AX tree, the snapshot now adds a fallback ref for that element.

  ## Changes

  - Record semantic interactive DOM elements during snapshot scanning
  - Add fallback refs only when the element is absent from the AX tree
  - Deduplicate fallback refs by `backendNodeId`
  - Preserve selector scoping for fallback refs
  - Keep existing AX-based output unchanged when Chrome already reports the element

  ## Testing

  - `cargo test --manifest-path cli/Cargo.toml snapshot::tests`
  - Local smoke test for `snapshot -i -c --json` with button and `role="radio"`

### DIFF
--- a/cli/src/native/snapshot.rs
+++ b/cli/src/native/snapshot.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use serde_json::Value;
 
@@ -180,6 +180,13 @@ struct CursorElementInfo {
     text: String, // textContent from the DOM element (fallback when ARIA name is empty)
     hidden_input_kind: Option<HiddenInputKind>,
     hidden_input_checked: Option<String>, // "true", "false", or "mixed" (tristate)
+    fallback_role: Option<String>,
+    checked: Option<String>,
+    expanded: Option<bool>,
+    selected: Option<bool>,
+    disabled: Option<bool>,
+    required: Option<bool>,
+    value_text: Option<String>,
 }
 
 struct RoleNameTracker {
@@ -406,7 +413,9 @@ pub async fn take_snapshot(
     for (idx, _) in &nodes_with_refs {
         if let Some(bid) = tree_nodes[*idx].backend_node_id {
             if let Some(cursor_info) = cursor_elements.get(&bid) {
-                tree_nodes[*idx].cursor_info = Some((*cursor_info).clone());
+                if !cursor_info.hints.is_empty() || cursor_info.hidden_input_kind.is_some() {
+                    tree_nodes[*idx].cursor_info = Some((*cursor_info).clone());
+                }
             }
         }
     }
@@ -556,6 +565,15 @@ pub async fn take_snapshot(
         }
     }
 
+    append_missing_dom_interactive_refs(
+        &mut output,
+        &tree_nodes,
+        &cursor_elements,
+        selector_backend_ids.as_ref(),
+        ref_map,
+        frame_id,
+    );
+
     if options.compact {
         output = compact_tree(&output, options.interactive);
     }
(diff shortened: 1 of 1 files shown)


vinitslal:
Hey @ctate — this PR fixes #1187 which we reported. We've been pinned at v0.19.0 waiting for this.

The approach looks solid — fallback refs for DOM elements missing from the AX tree over remote CDP, without changing existing AX-based output. CI is green and it's mergeable.

Could this get a review? Happy to help test a pre-release build if needed.

semimikoh:
Thanks @vinitslal  for the nudge! Happy to address any feedback once this gets a review.

ctate:
@semimikoh Thanks for the PR! The fallback-ref approach for #1187 is solid, but `e2e_snapshot_cursor_interactive` fails on this branch (passes on `main`):

```
Standard <a> element should not have cursor-interactive info:
- link "Link" [ref=e1] clickable [cursor:pointer]
```

Removing the `interactiveTags` skip causes native elements (`<a>`, `<button>`, etc.) that are already in the AX tree to pick up cursor-interactive hints. The semantic fallback collection needs to be separated from the cursor-interactive hint path so native elements only get fallback refs when absent from the AX tree, not extra annotations when present.

Also, please sign your commits before the next push.

Happy to re-review once the E2E regression is fixed.

semimikoh:
@ctate 

  Thanks for the review! I pushed a follow-up commit (`c8e5643`) that separates semantic DOM fallback collection from the cursor-
  interactive hint path.

  Native interactive elements and interactive ARIA roles are still collected as fallback candidates, but their cursor hints are suppressed
  when they are already represented in the AX tree, so `<a>` / `<button>` no longer get extra `clickable [cursor:pointer]` annotations.

  Verified locally:

  - `cargo fmt -- --check`
  - `cargo test native::snapshot::tests`
  - `cargo test e2e_snapshot_cursor_interactive -- --ignored --test-threads=1`

  The new commit is signed as requested.

PR #1278, opened 2026-04-20
feat(react): always-on suspense event log

gaojude:
## Summary

Adds an always-on React Suspense transition recorder, exposed as
`agent-browser react suspense-log [--clear] [--json] [--no-source-maps]`.
Preserves `react suspense` as-is (single-shot current-tree snapshot).

The recorder installs alongside `installHook.js` when `--enable react-devtools`
is passed at launch. It wraps `hook.onCommitFiberRoot` and walks the fiber
tree on each commit, diffing `memoizedState` against the previous commit to
detect suspend/resolve edges. On the suspend edge, it synchronously captures
`suspendedBy` / `ownerStack` / `awaiterStack` via `inspectElement` — before
React clears them on resolve.

Full design rationale and the PPR two-phase recipe is in
[`docs/suspense-log-pr-plan.md`](https://github.com/vercel-labs/agent-browser/blob/jude/suspense-log/docs/suspense-log-pr-plan.md)
(lives in the sibling `next-browser` repo where this was co-designed).

Commit structure (five logical pieces, four commits — the scripts change
and the daemon action share the same types module so they ship together):

1. `feat(react): add always-on Suspense transition recorder` — extracts the
   shared ops decoder, adds `SUSPENSE_LOG_INIT`, daemon action, CLI verb,
   and a small `output.rs` tweak so `{ "report": "..." }` responses don't
   get swallowed by the `✓ Done` fallback.
2. `feat(react): source-map resolution for suspense-log stack frames` — adds
   `sourcemap = "9"`, per-query resolver, untagged `MaybeResolvedFrame` enum.
3. `test(react): e2e coverage for react suspense-log` — two `#[ignore]`
   tests in `e2e_tests.rs` plus a React 19 ESM fixture.
4. `docs(react): skill reference for suspense-log + PPR recipe`.

## Notable design deviations from the original plan

- **Fiber-walk, not operations-subscribe.** The plan proposed subscribing
  to `hook.emit("operations", ...)`. In React 19 without a full DevTools
  backend (no `Agent.setRendererInterfaces`), runtime commits don't emit
  operations events to hook listeners — only the initial flush does.
  Walking fibers on `onCommitFiberRoot` is the one source of truth for
  transitions that doesn't depend on bridge state. Scripts.rs docstring
  explains this at length.
- **React 19 ESM fixture, not UMD.** React 19 dropped UMD builds. The E2E
  fixture uses `esm.sh` ESM modules via `<script type="module">`. Longer
  timers (800ms / 1600ms) so concurrent React's commit batching doesn't
  coalesce the suspended state away before the test observes it.

## Test plan

- [x] `cargo build --manifest-path cli/Cargo.toml` — clean
- [x] `cargo test --manifest-path cli/Cargo.toml` — 710 passed, 0 failed
- [x] `cargo test e2e_react_suspense_log -- --ignored --test-threads=1` —
      2 passed (`_captures_transitions`, `_errors_without_enable`)
- [x] `react suspense` continues to work against the React 19 fixture

## Out of scope

- Reimplementing `react suspense` in terms of the log — kept separate per
  the plan; both work side-by-side with the shared decoder.
- Next.js `/__nextjs_original-stack-frames` integration — skill concern.

### DIFF
--- a/.changeset/smooth-otters-sneeze.md
+++ b/.changeset/smooth-otters-sneeze.md
@@ -1,0 +1,5 @@
+---
+"agent-browser": minor
+---
+
+feat(react): add always-on Suspense transition log (`react suspense-log`)
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -65,6 +65,7 @@ dependencies = [
  "sha2",
  "similar",
  "socket2",
+ "sourcemap",
  "tempfile",
  "time",
  "tokio",
@@ -214,6 +215,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
+name = "base64-simd"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "339abbe78e73178762e23bea9dfd08e697eb3f3301cd4be981c0f78ba5859195"
+dependencies = [
+ "outref",
+ "vsimd",
+]
+
+[[package]]
 name = "bit_field"
 version = "0.10.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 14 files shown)


gaojude:
Closing — validated against a real Next.js 16 canary app with Cache Components enabled (client suspender via `usePathname`, server suspender via `cookies()`). The existing `react suspense` in #1257 already captures both, classifier-labeled, post-resolve:

```
# 4 boundaries: 2 dynamic holes, 2 static
Top actionable hole: CombinedPage — usePathname (SSR) (client-hook)
Most common root cause: cookies (request-api) affecting 1 boundary
```

The recorder's premise — that `suspendedBy` is GC'd on resolve and only capturable mid-transition — does not hold in dev mode. React 19's debug trace retains postponement reasons on the Initial Paint record, and `inspectElement` retrieves them after everything has settled. A post-resolve one-shot is sufficient for the PPR diagnostic workflow this was built for.

The recorder's remaining value (transition timing, runtime re-suspensions) is narrower than the description implied and doesn't justify the extra surface area right now. Leaving the branch intact for reference.

PR #1297, opened 2026-04-24
feat(paths): support AGENT_BROWSER_ and XDG base directories

### LINKED ISSUE #1361: Support XDG Base Directory paths for agent-browser state, config, and installs
## Problem

On Linux, `agent-browser` writes state/config/cache directly under `~/.agent-browser`. This creates home-directory clutter on systems that follow the XDG Base Directory spec.

Concrete examples in the current codebase:

- `cli/src/install.rs`: `get_browsers_dir()` uses `~/.agent-browser/browsers`, so `agent-browser install` installs browser binaries there.
- `cli/src/native/state.rs`: `get_state_dir()` uses `~/.agent-browser` for sessions, auth storage, and the encryption key.
- `cli/src/flags.rs`: user config lookup reads `~/.agent-browser/config.json`.
- Several temp/artifact paths also use `~/.agent-browser/tmp/...` for traces, profiles, screenshots, HAR files, etc.

There is already `AGENT_BROWSER_SOCKET_DIR` for sockets, but I could not find an equivalent override for the main state/config/install root. The docs mention `AGENT_BROWSER_HOME`, but the Rust paths above appear to call `dirs::home_dir().join(".agent-browser")` directly.

## Suggested Behavior

On Linux, prefer XDG locations when the relevant environment variables are set, falling back to the spec defaults when they are not:

- Config: `${XDG_CONFIG_HOME:-$HOME/.config}/agent-browser/config.json`
- State/auth/sessions: `${XDG_STATE_HOME:-$HOME/.local/state}/agent-browser`
- Browser installs/data: `${XDG_DATA_HOME:-$HOME/.local/share}/agent-browser/browsers`
- Cache/temp artifacts: `${XDG_CACHE_HOME:-$HOME/.cache}/agent-browser`

An explicit override would also work, for example:

- `AGENT_BROWSER_HOME`
- `AGENT_BROWSER_CONFIG_DIR`
- `AGENT_BROWSER_STATE_DIR`
- `AGENT_BROWSER_CACHE_DIR`
- `AGENT_BROWSER_DATA_DIR`

## Why

This keeps Linux home directories clean and lets NixOS/Home Manager users configure `agent-browser` declaratively without a persistent `~/.agent-browser` directory.

tunnckoCore:
## Summary
- add a shared path resolver so config, persistent data, cache artifacts, and runtime files stop defaulting to `~/.agent-browser`
- prefer `AGENT_BROWSER_*_DIR` overrides first, then fall back to the matching `XDG_*` directories for config, data, cache, and runtime locations
- update help text, README, skill docs, and docs pages so the new directory layout and precedence are documented consistently

---

Because... stop fricking polluting my HOME dir.

---

edit 3 after: closes #1361

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,9 @@ Alternatively, use `--session-name` to automatically save and restore cookies an
 agent-browser --session-name twitter open twitter.com
 
 # Login once, then state persists automatically
-# State files stored in ~/.agent-browser/sessions/
+# State files stored in $AGENT_BROWSER_DATA_DIR/agent-browser/sessions/
+# or $XDG_DATA_HOME/agent-browser/sessions/
+# (usually ~/.local/share/agent-browser/sessions/)
 
 # Or via environment variable
 export AGENT_BROWSER_SESSION_NAME=twitter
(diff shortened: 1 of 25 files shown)


tunnckoCore:
Closing this in favor of a smaller PR (#1362) based on latest upstream with a narrower path-resolution implementation, and focus on Linux only.

PR #1317, opened 2026-05-03
daemon: set default idle timeout to 30 minutes

pikapi12321:
## Problem

`AGENT_BROWSER_IDLE_TIMEOUT_MS` previously defaulted to **disabled** (no auto-shutdown). The daemon relied entirely on callers like hermes-agent to set this env var. When a daemon was launched without it:

- The daemon lives **indefinitely**
- Chrome processes leak and can spin to 100% CPU
- Observed in the wild: 8 renderer/gpu/utility processes consuming all cores for **6+ days**

## Fix

Change the default from disabled to **30 minutes** (1,800,000 ms). Set to 0 to restore the old opt-in behaviour.

The idle timer resets on every command received, so active sessions are unaffected. This is a safety net — hermes-agent already sets a 5-minute idle timeout which overrides this default.

## Changes

- `cli/src/native/daemon.rs`: `.filter(|&ms| ms > 0)` → `.unwrap_or(1_800_000)` with zero-check
- `cli/src/output.rs`: update help text

### DIFF
--- a/cli/src/native/daemon.rs
+++ b/cli/src/native/daemon.rs
@@ -113,11 +113,12 @@ pub async fn run_daemon(session: &str) {
     }
 
     // Auto-shutdown the daemon after this many ms of inactivity (no commands received).
-    // Disabled when unset or 0.
+    // Defaults to 30 minutes (1,800,000 ms).  Set to 0 to disable.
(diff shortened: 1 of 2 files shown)


pikapi12321:
Any body?

PR #1341, opened 2026-05-09
perf: avoid redundant CDP round-trips in evaluate and navigate

pedro-saraiva88:
## Summary

Read-only handlers that decorate their JSON response with the current page URL were issuing an extra `Runtime.evaluate location.href` after every operation, doubling the CDP round-trip count for ~13 commonly-used commands. `BrowserManager::navigate()` was also waiting on lifecycle events that Chrome never fires for same-document navigations.

## Measured impact

Benchmarked on a warm daemon, headless Chrome, `https://example.com`, 5 runs each:

| Op | Before | After |
| --- | --- | --- |
| `evaluate` | 0.37 ms | 0.25 ms (-32%) |
| `navigate` (same-doc / hash routing) | per-call timeout | returns immediately |
| `navigate` (full) | 19.48 ms | 19.48 ms |

The headline win is `evaluate` because it's in the hot path of every interaction loop (snapshot → ref → act). `navigate` is unchanged for full navigations — see "What's intentionally not in this PR" below.

## Changes

1. **`BrowserManager::cached_url()` / `cached_title()`** — read from `pages[active_page_index]`, which is already kept in sync by `navigate()` and Target events. No new state, no new sync paths.

2. **`handle_evaluate` and 8 sibling read-only handlers** (`handle_content`, `handle_snapshot`, `handle_gettext`, `handle_getattribute`, `handle_isvisible`, `handle_isenabled`, `handle_ischecked`, ...) now call `mgr.cached_url()` instead of `mgr.get_url().await`. Per the VADE review, four navigation-adjacent handlers (`handle_back`, `handle_forward`, `handle_reload`, `handle_waitforurl`) and the user-facing `handle_url` keep `get_url().await` because they change the URL or are expected to return live state.

3. **`BrowserManager::navigate()`** skips `wait_for_lifecycle` when the `Page.navigate` response carries no `loaderId`. That is Chrome's signal for a same-document navigation (e.g. hash routing). Without this guard the previous code would wait for `Page.loadEventFired` / `Page.domContentEventFired` events that Chrome never fires in that case, until the per-call timeout — which would surface as user-visible latency on SPA route changes done via `agent-browser open`.

## Compatibility

- `cached_url()` / `cached_title()` are additive helpers; nothing was removed from the public API.
- `handle_url`, `handle_back`, `handle_forward`, `handle_reload`, `handle_waitforurl` are deliberately unchanged behaviour — they always return the live URL, since the cache hasn't been refreshed yet at those points.
- The `loaderId.is_some()` guard in navigate is purely defensive — same-doc navigations could not have been waiting on those events anyway, they were just timing out silently.

## Testing

```bash
cd cli
cargo build --release             # clean
cargo clippy --tests -- -D warnings   # clean
cargo fmt --check                 # clean
cargo test -- --test-threads=1    # 717 passed (parallel mode hits a pre-existing env-var race in test_default_timeout_ms_*)
```

End-to-end smoke against a local Filament/Laravel app: `agent-browser open http://localhost:8080/admin/login` followed by `agent-browser eval "location.href"` and `agent-browser screenshot --full <path>` — page renders, screenshot captures the actual login page, eval returns the navigated URL.

## Self-reported during this PR

While dogfooding this PR, an earlier revision tried to consolidate the URL+title fetch in `navigate()` into a single `evaluate("JSON.stringify([...,...])")` call to save one round-trip. That hunk left the daemon's active session pointing at a different target after navigation completed — `eval` would run against `about:blank` or the Chrome debug index page (`127.0.0.1:<cdpPort>`). The two-round-trip URL+title fetch is restored in 429cbb5; bisected, isolated, and explained in the commit. The current diff has no perf claim that depends on that hunk.

## What's intentionally not in this PR

The remaining ~18 ms gap on `Page.navigate` is head-of-line blocking in the CDP read loop: the reader processes events serially via `event_tx.send()` (a `tokio::broadcast` channel), and a slow subscriber can stall the reader until response delivery. Fixing that means either splitting the response-routing path from the event broadcast path or using `try_send` on the broadcast — both warrant their own PR with dedicated benchmarks. Happy to follow up.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -3102,7 +3102,9 @@ async fn handle_back(state: &mut DaemonState) -> Result<Value, String> {
     let mgr = state.browser.as_ref().ok_or("Browser not launched")?;
     mgr.evaluate("history.back()", None).await?;
     tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
-    let url = mgr.cached_url();
+    // history.back() does not flow through navigate(), so the cached URL on
+    // PageInfo is still the pre-navigation value. Read live to avoid staleness.
+    let url = mgr.get_url().await.unwrap_or_default();
     state.ref_map.clear();
     Ok(json!({ "url": url }))
 }
@@ -3120,7 +3122,9 @@ async fn handle_forward(state: &mut DaemonState) -> Result<Value, String> {
     let mgr = state.browser.as_ref().ok_or("Browser not launched")?;
     mgr.evaluate("history.forward()", None).await?;
     tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
-    let url = mgr.cached_url();
+    // history.forward() does not flow through navigate(), so the cached URL is
+    // still the pre-navigation value. Read live to avoid staleness.
+    let url = mgr.get_url().await.unwrap_or_default();
     state.ref_map.clear();
     Ok(json!({ "url": url }))
 }
@@ -3160,7 +3164,11 @@ async fn handle_reload(state: &mut DaemonState) -> Result<Value, String> {
     })
     .await;
 
-    let url = mgr.cached_url();
+    // Page.reload triggers a fresh navigation that does not flow through
+    // BrowserManager::navigate(), so the cached URL is still the pre-reload
+    // value (the URL is the same after a same-page reload, but for SPAs that
+    // reload at a hash anchor this still matters). Read live.
+    let url = mgr.get_url().await.unwrap_or_default();
     state.ref_map.clear();
     Ok(json!({ "url": url }))
 }
@@ -5400,7 +5408,9 @@ async fn handle_waitforurl(cmd: &Value, state: &DaemonState) -> Result<Value, St
     let timeout_ms = state.timeout_ms(cmd);
 
     wait_for_url(&mgr.client, &session_id, url_pattern, timeout_ms).await?;
-    let url = mgr.cached_url();
+    // wait_for_url returned because the URL just changed; the cached URL is
+    // by definition the previous one. Read live.
+    let url = mgr.get_url().await.unwrap_or_default();
     Ok(json!({ "url": url }))
 }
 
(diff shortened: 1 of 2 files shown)


pedro-saraiva88:
Good catch — fixed in eff14ba.

`handle_back`, `handle_forward`, `handle_reload` and `handle_waitforurl` now read the URL live with `mgr.get_url().await` instead of `cached_url()`. The cache is only kept in sync by `BrowserManager::navigate()` and by Target events, so these four code paths were indeed returning the pre-navigation URL — the exact moment when staleness is most user-visible. Each site has a short comment explaining why the cache is bypassed.

The nine remaining handlers that decorate read-only responses with `origin` keep `cached_url()` since they do not change the URL.

Validated with `cargo build`, `cargo clippy --tests -- -D warnings`, `cargo fmt --check`, and `cargo test -- --test-threads=1` (717 passed).

pedro-saraiva88:
Self-reported regression: dropped the navigate eval consolidation in 429cbb5.

While dogfooding this on a real Filament/Laravel app I found that the `JSON.stringify([location.href, document.title])` change in `BrowserManager::navigate()` left the daemon's active session pointing at a different target after navigation completed — the navigate response had the correct values, but a subsequent `agent-browser eval` would run against `about:blank` or the Chrome debug index page (`127.0.0.1:<cdpPort>`). Reproduced reliably with `agent-browser open http://localhost:8080/admin/login` followed by `agent-browser eval "location.href"`. Bisected to that single hunk; reverting it fixes the regression while the rest of the PR stays clean.

Updated PR contents:

- `cached_url()` / `cached_title()` accessors (additive)
- 9 read-only handlers using `cached_url()` for response decoration → `evaluate` -32% (0.37ms → 0.25ms)
- `back` / `forward` / `reload` / `waitforurl` keep using live `get_url().await` (per @vercel VADE review)
- `wait_for_lifecycle` skipped for same-document navigations (`loader_id == None`) — correctness fix, kept

`navigate()` itself stays at the original two-round-trip URL+title fetch. The 0.3ms saving from consolidating those is interesting future work but not worth the foot-gun.

Validated: `cargo build --release`, `cargo clippy --tests -- -D warnings`, `cargo fmt --check`, `cargo test -- --test-threads=1` (717 passed), plus a manual end-to-end smoke against a local Filament app — page renders, screenshot captures the actual page (was a black image with the buggy version), `eval` returns the right URL.

PR #1370, opened 2026-05-16
Make tab switch foregrounding optional via --no-activate

Prasanna721:
## Problem

`tab <ref>` bundles two operations: it sets the session's active tab pointer (logical select) and brings the tab to the foreground via `Page.bringToFront` (physical activate). For interactive agents driving one tab at a time, that's the right default. For background syncers that need to address every tracked tab in turn (to read each one's snapshot), every poll cycle ends up calling `bringToFront` on every tab in sequence, fighting the user's foreground in headed mode.

## Change

Add a `--no-activate` flag to `tab <ref>` that skips the `Page.bringToFront` call but still updates `active_page_index` and re-enables CDP domains on the target session. Default behavior is unchanged.

## Mechanics

The infrastructure already exists: every tab is attached via `Target.attachToTarget` with its own `session_id` stored in `PageInfo`, and `active_session_id()` resolves through `active_page_index`. The flag just makes the existing `bringToFront` step optional.

- `commands.rs`: parser for `tab <ref>` accepts `--no-activate`, sets `noActivate: true` in the action JSON. Unknown flags still error.
- `actions.rs::handle_tab_switch`: reads `noActivate`, forwards inverse as `activate: bool` to `tab_switch_by_id`.
- `browser.rs::tab_switch` / `tab_switch_by_id`: take `activate: bool`; `Page.bringToFront` gated by `if activate { … }`.
- `output.rs`: `tab --help` documents the new flag.

## Use case

I'm building [preprint](https://github.com/Prasanna721/preprint), a file-system view of a real browser for agents. Every tab is projected as a markdown file kept in sync by a daemon. Without this flag the daemon refocuses every tab on every poll, which fights the user in `--preview` mode. With it the daemon can address each tab logically without touching foreground.

## Tests

New parser tests in `commands.rs`:
- `tab tN --no-activate` parses to `{action: tab_switch, tabId: tN, noActivate: true}`.
- `tab tN` defaults to no `noActivate` field (activates).
- Unknown flags on `tab tN` error.

Existing tab tests pass.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1261,11 +1261,29 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
                     }
                     Ok(cmd)
                 }
-                Some(tab_ref) => Ok(json!({
-                    "id": id,
-                    "action": "tab_switch",
-                    "tabId": tab_ref,
-                })),
+                Some(tab_ref) => {
+                    let mut cmd = json!({
+                        "id": id,
+                        "action": "tab_switch",
+                        "tabId": tab_ref,
+                    });
+                    let mut i = 1;
+                    while i < rest.len() {
+                        match rest[i] {
+                            "--no-activate" => {
+                                cmd["noActivate"] = json!(true);
+                                i += 1;
+                            }
+                            other => {
+                                return Err(ParseError::UnknownSubcommand {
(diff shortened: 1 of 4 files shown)


Prasanna721:
Closing — premature, opened without owner confirmation. Will re-open when ready.

PR #1374, opened 2026-05-18
feat(dashboard): live annotation toggle in viewport toolbar

techcomthanh:
## Summary

Adds a `#` toggle in the dashboard viewport toolbar (left of the snapshot icon) that injects the same numbered red boxes used by `screenshot --annotate` onto the **live page**, so element refs are visible in the streaming viewport without taking a screenshot.

Motivation: when picking elements interactively from the dashboard, `--annotate` only labels stills. The labels disappear the moment streaming continues, which is awkward for human-in-the-loop selection. This toggle reuses the existing overlay injector so labels stay on screen while the user identifies the target, then can be removed in one click.

CLI equivalent (also new):
```
agent-browser --session <name> annotate-overlay on
agent-browser --session <name> annotate-overlay off
```

The numbers shown on the live overlay match `@eN` from `snapshot -i` — the action takes a fresh snapshot on enable so the ref map is current.

## Changes

**Rust (`cli/`)**
- `native/screenshot.rs`: extract a public `set_live_annotation(client, session_id, ref_map, on)` that reuses the existing `inject_annotation_overlay` / `remove_annotation_overlay` helpers (no new injection code).
- `native/actions.rs`: new `annotate_overlay` action. On enable, clears + refills the ref map via `snapshot::take_snapshot` so overlay numbers match the current DOM.
- `commands.rs`: new `annotate-overlay on|off` CLI verb.

**Dashboard (`packages/dashboard/`)**
- `components/viewport.tsx`: new `Hash` icon toggle button placed left of the snapshot button; local `annotateOn` state; calls `runCmd(\"annotate-overlay\", on?\"on\":\"off\")`. A 300ms-debounced `useEffect` on `url` re-applies the overlay after navigation so labels track the new DOM.

## Test plan

- [x] `cargo build --release --manifest-path cli/Cargo.toml` — clean build.
- [x] `pnpm --filter dashboard build` — clean build.
- [x] Smoke: `agent-browser open https://example.com` → `annotate-overlay on` → DOM contains `#__agent_browser_annotations__` overlay; viewer shows numbered boxes. `off` removes them. Verified via `agent-browser eval` post-toggle.
- [x] Dashboard button reflects state (background tint when active) and tooltip flips between \"Show/Hide element labels\".
- [ ] Multi-tab / iframe sanity (not exercised).
- [ ] WebDriver backend (current implementation is Chrome-CDP only, matches `screenshot --annotate`).

## Notes / Open questions

- Auto-refresh after navigation is heuristic (300ms `setTimeout` on `url` change). SPA route changes that don't update `activeUrlAtom` won't trigger it; users re-click the toggle. A CDP `Page.frameNavigated` hook on the server side would be more robust — happy to follow up if preferred.
- Icon choice (`Hash` from `lucide-react`) is a guess; open to alternatives.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -741,6 +741,13 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
             Ok(cmd)
         }
 
+        // === Annotate overlay (dashboard live-view toggle) ===
+        "annotate-overlay" => {
+            let arg = rest.first().copied().unwrap_or("on");
+            let on = matches!(arg, "on" | "true" | "1" | "show" | "enable");
+            Ok(json!({ "id": id, "action": "annotate_overlay", "on": on }))
+        }
+
         // === Eval ===
         "eval" => {
             // Check for flags: -b/--base64 or --stdin
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -1293,6 +1293,7 @@ pub async fn execute_command(cmd: &Value, state: &mut DaemonState) -> Value {
         "close" => handle_close(state).await,
         "snapshot" => handle_snapshot(cmd, state).await,
         "screenshot" => handle_screenshot(cmd, state).await,
+        "annotate_overlay" => handle_annotate_overlay(cmd, state).await,
         "click" => handle_click(cmd, state).await,
         "dblclick" => handle_dblclick(cmd, state).await,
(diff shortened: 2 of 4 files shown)


techcomthanh:
Addressed in 259430c: reset `annotateOn` to `false` in a `useEffect` keyed on `sessionName`. Now switching sessions clears the toggle so it matches the new session's actual overlay state (none injected yet). User re-enables explicitly per session.

PR #1375, opened 2026-05-19
fix: reject invalid dashboard ports

hiSandog:
## Summary
- Parse dashboard start --port through a small helper instead of silently falling back to 4848
- Report missing, invalid, or out-of-range port values as CLI errors
- Add unit coverage for default, valid, missing, and invalid dashboard port arguments

## Validation
- git diff --check

I could not run cargo fmt or cargo test locally because this environment does not have cargo/rustc on PATH.


hiSandog:

PR #1389, opened 2026-05-26
add AGENT_BROWSER_DASHBOARD_VIEWPORT_ONLY env to config if dashboard

lidingshan:
Add an environment variable `AGENT_BROWSER_DASHBOARD_VIEWPORT_ONLY` to decide if the dashboard should only show the viewport. This will allow the user to interact with the backend application when required. For example, let user to login the system before starting the automation process. The reason to hide other UI elements is to provide simple UX so the user won't be confused.


lidingshan (commented):
Have reviewed files and without issues.

lidingshan (commented):
Reviewed the changes.

PR #1393, opened 2026-05-26
fix: parse boolean flag values case-insensitively

hiSandog:
## Summary
- Parse optional boolean flag values case-insensitively so `FALSE` and `TRUE` behave like `false` and `true`.
- Consume case-insensitive boolean values in `clean_args` so they do not leak into command arguments.
- Add unit coverage for parsing and argument cleanup.

## Validation
- `git diff --check`
- Not run: `cargo test --manifest-path cli/Cargo.toml flags` because `cargo`/`rustc` are not installed in this environment.


hiSandog:

PR #1395, opened 2026-05-27
fix(cli): prevent env-derived flags from triggering implicit re-launch

### LINKED ISSUE #1299: `--headed` window dies and Chrome is killed+relaunched on every CLI invocation (macOS 26 / Darwin 25.x)
## Summary

On macOS 26 (Tahoe), every CLI command sent to a `--headed` daemon causes the daemon to **kill the entire Chrome process tree and relaunch a new Chrome in headless mode**. The visible window disappears within ~1s. The bug is `--headed`-specific (headless mode is unaffected) and reproduces with or without `--profile`.

The CLI command itself returns a result, but:
- **Without `--profile`**: `get url` returns `about:blank` (page state is lost — looks like #1211).
- **With `--profile`**: `get url` returns the original URL because the profile restores the page on relaunch — making the bug subtler and easier to miss, but the kill+relaunch still happens.

Reproducible on a fresh ephemeral profile, no env vars, no custom args. Affects both **0.25.3** and **0.26.0**.

## Environment

- macOS 26 (Darwin 25.4.0), Apple M4 (arm64)
- agent-browser 0.25.3 and 0.26.0 (npm global install)
- Chrome for Testing 147.0.7727.56 (bundled)
- No custom config, no env vars set

## Repro — bug present with or without --profile

```bash
agent-browser close --all

# Variant 1: no --profile at all
agent-browser open --headed https://example.com
sleep 5
ps -ax -o pid,command | grep "Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing " \
  | grep -v Helper | grep -v grep | awk '{print "BEFORE:", $1}'
agent-browser get url      # → about:blank (page state lost)
ps -ax -o pid,command | grep "Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing " \
  | grep -v Helper | grep -v grep | awk '{print "AFTER: ", $1}'
# Output:
# BEFORE: 11172
# AFTER:  11211
# Window disappears.

# Variant 2: with --profile
agent-browser close --all
agent-browser open --headed --profile /tmp/ab-test https://example.com
sleep 5
# (snapshot before)
agent-browser --profile /tmp/ab-test get url   # → https://example.com/ (looks fine!)
# (snapshot after)
# Output:
# BEFORE: 12612
# AFTER:  12651
# Window also disappears, but the URL looks right because the profile restored the page.
```

In both variants, every renderer / gpu / network helper PID also changes between snapshots — entire process tree replaced.

## Headless is NOT affected

Same repro without `--headed`:

```bash
agent-browser close --all
agent-browser open https://example.com   # headless
# BEFORE: 7653
agent-browser get url                     # → https://example.com/
# AFTER:  7653   ← same PID, no relaunch
```

So the trigger is specifically the `--headed` window — the daemon seems to consider it disposable on every new client connection.

## Workaround

Setting `AGENT_BROWSER_HEADED=true` as a global env var (e.g. shell rc) avoids the visible symptom: the relaunched Chrome inherits the env var and starts in headed mode again, so the window stays visible. The kill+relaunch still happens though — this is a workaround, not a fix.

## Notes / related

- Looks like a macOS counterpart to #1148 (`PR_SET_PDEATHSIG kills Chrome after ~10s idle`), which was Linux-only and fixed in 0.25.2. Different mechanism, same effect.
- Distinct from #1211 (active page → about:blank with `--profile`): the original reporter says it works "without --profile" — on macOS 26 we see the same about:blank without --profile, plus the additional symptom of the visible window dying. May be the same root cause but with broader manifestation on macOS 26.
- Distinct from #1263 (Chrome orphan after network blip): no network event needed, triggered purely by a second CLI invocation.

xiaodankasi:
## Problem

When `AGENT_BROWSER_EXECUTABLE_PATH` (or other launch-related env vars / config values) is set, every CLI invocation that hits the daemon sends an implicit `launch` command — even when the user only runs a read-only action like `get url`.

Because this second launch omits flags the *original* launch included (e.g. `--proxy`), the computed `launch_hash` differs from the running browser's hash, causing the daemon to **kill and re-launch the browser**. The new browser instance starts at `about:blank`, breaking subsequent commands.

This was introduced in PR #996 (commit `05d86fa`, shipped in v0.24.1) which added hash-based relaunch detection.

## Root Cause

The launch condition in `main.rs` checks `flags.executable_path.is_some()` (among others), which is `true` whenever the env var `AGENT_BROWSER_EXECUTABLE_PATH` is set — regardless of whether the user explicitly passed `--executable-path` on the CLI. This means an env-only configuration silently triggers a launch command on every invocation.

## Fix

Guard the implicit launch with `has_explicit_launch_flags` — a boolean that is `true` only when the user passes launch-relevant flags directly on the CLI (using the `cli_*` tracking fields introduced in PR #996 itself). When the daemon is **already running** and no explicit CLI launch flags are present, the launch command is skipped entirely.

This preserves PR #996's design intent: users can still explicitly change browser options mid-session with `--headed`, `--proxy`, etc., triggering a hash-aware relaunch. But env/config-derived values alone will no longer cause unintended relaunches.

## Reproduction

```bash
# On a machine with AGENT_BROWSER_EXECUTABLE_PATH set:
agent-browser open --proxy http://proxy:port --session test
agent-browser --session test get url   # ← triggers relaunch, returns about:blank
```

## Testing

Verified on Linux (Alibaba Linux, Chrome 140) where `AGENT_BROWSER_EXECUTABLE_PATH=/usr/bin/google-chrome-140`:
- `open --proxy` → `get url` (no explicit flags): browser stays alive, returns correct URL ✓
- `open --proxy` → `get url --proxy` (explicit flag): correctly triggers relaunch ✓
- `open --headed` → `get url`: no spurious relaunch ✓
- Fresh start (no daemon): launch proceeds normally ✓

Closes #1299

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1072,24 +1072,41 @@ fn main() {
         }
     }
 
-    // Launch headed browser or configure browser options (without CDP or provider)
-    if (flags.headed
-        || flags.cli_headed  // User explicitly set --headed (even if false)
-        || flags.executable_path.is_some()
-        || flags.profile.is_some()
-        || flags.state.is_some()
-        || flags.proxy.is_some()
-        || flags.args.is_some()
-        || flags.user_agent.is_some()
-        || flags.allow_file_access
-        || should_send_hide_scrollbars_launch_option(
-            flags.cli_hide_scrollbars,
-            flags.hide_scrollbars,
-        )
-        || flags.color_scheme.is_some()
-        || flags.download_path.is_some()
-        || flags.engine.is_some()
-        || !flags.extensions.is_empty())
+    // Launch headed browser or configure browser options (without CDP or provider).
+    // When the daemon is already running, only re-send a launch command if the user
+    // explicitly passed launch-relevant CLI flags (indicating intent to change options).
+    // Environment-variable or config-file derived values alone must NOT trigger a
+    // re-launch — doing so causes hash mismatches that kill the running browser (#1299).
+    let has_explicit_launch_flags = flags.cli_headed
+        || flags.cli_executable_path
+        || flags.cli_profile
+        || flags.cli_state
+        || flags.cli_proxy
+        || flags.cli_args
+        || flags.cli_user_agent
+        || flags.cli_allow_file_access
+        || flags.cli_extensions
+        || flags.cli_download_path
+        || flags.cli_hide_scrollbars;
+
+    if (!daemon_result.already_running || has_explicit_launch_flags)
+        && (flags.headed
+            || flags.cli_headed // User explicitly set --headed (even if false)
+            || flags.executable_path.is_some()
+            || flags.profile.is_some()
+            || flags.state.is_some()
+            || flags.proxy.is_some()
+            || flags.args.is_some()
+            || flags.user_agent.is_some()
+            || flags.allow_file_access
+            || should_send_hide_scrollbars_launch_option(
+                flags.cli_hide_scrollbars,
+                flags.hide_scrollbars,
+            )
+            || flags.color_scheme.is_some()
+            || flags.download_path.is_some()
+            || flags.engine.is_some()
+            || !flags.extensions.is_empty())
         && flags.cdp.is_none()
         && flags.provider.is_none()
         && !flags.auto_connect


xiaodankasi:
Hi @ctate, could you take a look when you have a moment? This fixes a bug where env-derived flags (e.g. `AGENT_BROWSER_EXECUTABLE_PATH`) trigger an implicit re-launch on every CLI invocation, causing hash mismatches that kill the running browser. It also closes #1299. Thanks!

xiaodankasi (commented):
Re: Vercel bot comment about missing `cli_color_scheme` and `cli_engine` —

These `cli_*` tracking fields do not exist in the codebase. `flags.rs` only defines: `cli_headed`, `cli_executable_path`, `cli_profile`, `cli_state`, `cli_proxy`, `cli_args`, `cli_user_agent`, `cli_allow_file_access`, `cli_extensions`, `cli_download_path`, and `cli_hide_scrollbars`.

Adding `cli_color_scheme` / `cli_engine` would be a separate enhancement. Also worth noting: `color_scheme` is not included in `launch_hash()`, so it does not affect relaunch behavior regardless.

TreedsSlop:
+1 for fixing this. I had the same issue and spent way too long debugging it.
Two gaps I saw, but I'm not entirely sure about:

1. Browser-death mid-session silently degrades the session. Once this lands, a
   read-only command (get url / eval) after the browser has died sends no envelope,
   so the daemon's implicit auto_launch (actions.rs:3312) rebuilds options from
   launch_options_from_env() — env/config defaults, NOT the session's effective
   options. A session that was `open --headed` comes back headless; viewport/proxy/
   engine/args are lost too. The user ran a read-only command, so re-passing launch
   flags on every command is non-intuitive. This needs a daemon-side complement:
   persist the effective LaunchOptions and reuse them on auto-relaunch.

2. has_explicit_launch_flags is missing several options that the current
   should_send_local_launch_config triggers on, so explicit CLI flags get silently
   ignored mid-session:
   - --engine and --color-scheme are still in the inner condition (~line 1108) but
     have no cli_* field, so they can never pass the outer gate on their own.
   - --init-script, --enable, --webgpu DO have cli_* fields (cli_init_scripts,
     cli_enable, cli_webgpu) that are simply not in the gate list.
   - --allowed-domains is dropped from both conditions — it's hash-relevant and
     security-relevant (restrict_webrtc is derived from it).
   All of these are in launch_hash (actions.rs:4351), so the envelope is the only
   way they reach the daemon; without it the flag is just dropped.

PR #1398, opened 2026-05-29
fix: harden idle timeout parsing

hiSandog:
## Summary

- avoid byte-splitting the idle timeout suffix so non-ASCII units return an error instead of panicking
- use checked multiplication when converting s/m/h values to milliseconds
- add parser regression coverage for Unicode units and overflow

## Validation

- git diff --check
- cargo test parse_idle_timeout --manifest-path cli/Cargo.toml (not run: cargo is not installed in this local environment)


hiSandog:

PR #1399, opened 2026-05-31
Fix comma-containing Chromium --args handling

simonellefsen:
## Summary

Fix Chromium launch arg parsing so comma-containing switches survive the CLI-to-daemon handoff intact.

This came out of a real headed-browser failure in Klub Lotto automation. We were launching with `--window-size=1600,1200`, but agent-browser split that value at the comma before it reached Chrome. In practice that made headed sessions behave unexpectedly, and removing the launch arg restored normal visible browser behavior.

## Root cause

`AGENT_BROWSER_ARGS` was parsed by splitting on every comma and newline. That works for lists like:

- `--no-sandbox,--disable-blink-features=AutomationControlled`

But it breaks single Chromium switches whose value contains a comma, especially:

- `--window-size=1600,1200`

That value was being rewritten into two invalid args before launch:

- `--window-size=1600`
- `1200`

## What changed

- add a dedicated launch-arg parser for `AGENT_BROWSER_ARGS`
- keep newline separation
- keep comma separation between adjacent switches for backwards compatibility
- preserve commas inside a single switch value, such as `--window-size=1600,1200`
- add regression tests for parsing and env-to-launch forwarding
- add a headed launch regression test to prove the exact `--window-size=1600,1200` switch reaches the Chromium launch layer unchanged
- update CLI help, README, docs, and the agent skill to document the intended `--args` behavior clearly

## Verification

- `cargo test --manifest-path cli/Cargo.toml parse_launch_args`
- `cargo test --manifest-path cli/Cargo.toml window_size`
- `cargo test --manifest-path cli/Cargo.toml`

## Notes

This PR is intentionally focused on the parsing and forwarding bug. The launch layer already preserved a correctly formed `--window-size=...` arg once it reached Chromium options intact; the breakage was earlier in the env parsing path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -710,7 +710,7 @@ This is useful for multimodal AI models that can reason about visual layout, unl
 | `--extension <path>` | Load browser extension (repeatable; or `AGENT_BROWSER_EXTENSIONS` env) |
 | `--init-script <path>` | Register a page init script before the first navigation (repeatable; or `AGENT_BROWSER_INIT_SCRIPTS` env) |
 | `--enable <feature>` | Built-in init scripts: `react-devtools` (repeatable or comma-list; or `AGENT_BROWSER_ENABLE` env) |
-| `--args <args>` | Browser launch args, comma or newline separated (or `AGENT_BROWSER_ARGS` env) |
+| `--args <args>` | Browser launch args, newline-separated or comma-separated between switches. Commas inside one switch are preserved, e.g. `--window-size=1600,1200` (or `AGENT_BROWSER_ARGS` env) |
 | `--user-agent <ua>` | Custom User-Agent string (or `AGENT_BROWSER_USER_AGENT` env) |
 | `--proxy <url>` | Proxy server URL with optional auth (or `AGENT_BROWSER_PROXY` env) |
 | `--proxy-bypass <hosts>` | Hosts to bypass proxy (or `AGENT_BROWSER_PROXY_BYPASS` env) |
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1139,12 +1139,7 @@ fn main() {
         }
 
         if let Some(ref a) = flags.args {
-            // Parse args (comma or newline separated)
-            let args_vec: Vec<String> = a
-                .split(&[',', '\n'][..])
-                .map(|s| s.trim().to_string())
-                .filter(|s| !s.is_empty())
-                .collect();
+            let args_vec = crate::native::actions::parse_launch_args(a);
             cmd_obj.insert("args".to_string(), json!(args_vec));
         }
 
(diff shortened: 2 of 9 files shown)


curtisholtt:
Tested this PR against the window positioning use case on macOS (agent-browser 0.26.0). The `parse_launch_args` function logic is correct and handles the core issue. A few additional findings from our debugging:

**1. Same bug exists in three code paths, PR fixes one**

The comma-splitting is applied in at least three places, but the PR only fixes `launch_options_from_env()` (the `AGENT_BROWSER_ARGS` env var path). The same broken split also happens when args come from:

- The `--args` CLI flag (`flags.rs` → `DaemonOptions.args` as `Option<&str>` → daemon IPC)
- The `"args"` key in `config.json` (routes through `DaemonOptions` the same way)

We tested with `--args "--window-position=819,32,--window-size=862,1019"` and `"args": "..."` in config.json — neither worked in 0.26.0. So this PR eliminates one of three paths but the other two need the same `parse_launch_args` treatment applied wherever the raw string is eventually split into `Vec<String>` for `LaunchOptions.args`.

**2. Headed mode never gets a default `--window-size`**

In `build_chrome_args()` (chrome.rs), the default `--window-size` injection only fires when `headless && !has_extensions`:

```rust
if !has_window_size && options.headless && !has_extensions {
    let (w, h) = options.viewport_size.unwrap_or((1280, 720));
    args.push(format!("--window-size={},{}", w, h));
}
```

In headed mode with no custom `--args`, Chrome opens at whatever default it chooses (1200x975 on macOS Tahoe, Apple Silicon). The viewport from config.json (`"viewport": {"width": 873, "height": 872}`) isn't applied at launch either — it's only applied at runtime via the `set viewport` command. So without this PR's fix AND an explicit `--window-size` in args, headed mode gives users an inconsistent initial window.

**3. macOS caps JS window.moveTo() — CDP is the workaround**

Even with args correctly passed, if someone wants a specific window position (not just size), `--window-position` may not land at the requested coordinates on macOS. The JavaScript `window.moveTo()` in the renderer is subject to macOS window manager caps — we observed requested (840, 0) landing at (807, 0) — the menu bar and edge snap zones clamp values.

The CDP method `Browser.setWindowBounds` bypasses this entirely and hits the exact coordinates. We built a small script around it (connect to the CDP WebSocket, call `Browser.getWindowForTarget` + `Browser.setWindowBounds`). This only matters for position, not size — `--window-size` works fine once the comma parsing is fixed.

simonellefsen:
Thanks for the thorough review and for testing all three paths — that was exactly the right call.

Fixed in the latest commit: the `--args` flag and `config.json` `"args"` paths both converged at a naive `.split(&[',', '\n'][..])` in `main.rs` that was independent of the env var fix. Made `parse_launch_args` `pub(crate)` and plugged it in there, so all three sources now go through the same parser. Three new tests cover the fixed path directly.

Finding 2 (headed mode not getting a default `--window-size`) and finding 3 (macOS `window.moveTo` clamping / CDP workaround) are noted — happy to address those in a follow-up PR if that's the preferred approach, or I can fold them in here if you'd rather keep it together.

simonellefsen:
Rebased onto current `main` — this was conflicting and is now mergeable, as a single commit.

The branch was 79 commits behind and conflicted in the `actions.rs` test module, where a textual rebase would have meant hand-resolving hunks of unrelated upstream tests. I re-applied the change onto current `main` instead. The logic is unchanged from the original.

Confirmed the bug is still live on `main` before redoing the work — `cli/src/main.rs` and `launch_options_from_env` both still split on every comma, so `--args "--window-size=1600,1200"` still becomes `["--window-size=1600", "1200"]` and Chromium ignores both.

Two things worth noting since the diff moved:

- `main` has since gained `test_build_args_custom_window_size_not_overridden`, so the *default-override* half of this is already handled upstream. The comma-splitting half is not, and that's what remains here.
- The other comma-split env vars (`AGENT_BROWSER_ENABLE`, `AGENT_BROWSER_INIT_SCRIPTS`, `AGENT_BROWSER_EXTENSIONS`) are deliberately left alone — those are lists where every comma really is a separator. Only the two launch-arg entry points change.

Tests: 4 for `parse_launch_args` (comma-in-value preserved, split between adjacent switches, newline separation, trimming/empties) plus a `build_chrome_args` regression test. `cargo fmt --check` clean; 1237 pass, 0 fail.

Issue #1407, opened 2026-06-02
Bug: [Windows] CLI hangs when PowerShell captures output after close-then-open sequence


OnePaiGu:
### Description
On Windows, when using PowerShell to capture CLI output (via $output = & agent-browser ... or 2>&1 redirection), the CLI process hangs and does not return an exit code after executing close followed by open . The process only exits after running another close command.

This does not affect manual execution (without output capture), and does not occur on macOS or Linux.

### Reproduction

```powershell
# Step 1: Open a page (works fine)
$output = & agent-browser open --headed "https://example.com"

# Step 2: Close the session
$output = & agent-browser close

# Step 3: Open again — HANGS here, never returns
$output = & agent-browser open --headed "https://example.com"
# Only returns after running: agent-browser close
```

The same issue occurs with `2>&1` redirection:

```powershell
agent-browser close
agent-browser open --headed "https://example.com" 2>&1
# HANGS
```

### Expected Behavior
The open command should exit and return control to PowerShell immediately after the daemon is launched, regardless of whether output is being captured.

### Actual Behavior
The CLI process hangs indefinitely. PowerShell waits for all write ends of its output pipe to close before considering output complete. The daemon child process holds an inherited copy of the pipe write-end handle, preventing PowerShell from detecting that output is finished.

### Root Cause
When the CLI spawns the daemon process via Rust's std::process::Command::spawn() , the Windows implementation defaults to bInheritHandles = TRUE in the underlying CreateProcessW call. This causes the daemon to inherit all inheritable handles from the parent CLI process, including:

- The stdout pipe handle created by PowerShell for $output = capture
- The stderr pipe handle created by PowerShell for 2>&1 redirection
After the CLI exits (closing its own pipe handles), the daemon still holds the inherited pipe write-end handle. PowerShell's output capture waits for all write ends to close, causing the hang.

This only manifests after close then open because the first open starts a new daemon, close kills it, and the second open spawns a new daemon that inherits the current PowerShell session's pipe handles.

### Proposed Fix
Use the Windows CreateProcessW API directly in cli/src/connection.rs to spawn the daemon with bInheritHandles = FALSE , preventing the daemon from inheriting any handles from the parent process.

Key changes:

[...]

PR #1408, opened 2026-06-02
fix: Windows handle inheritance leak

OnePaiGu:
## Root Cause

**Windows handle inheritance leak.**

When the CLI process spawns the daemon via `Command::spawn()`, the Rust standard library on Windows defaults to `bInheritHandles = TRUE`. This means the daemon child process inherits all handles marked as "inheritable" from the parent process, including:

- The **stdout pipe handle** created by PowerShell for `$output =` capture
- The **stderr pipe handle** created by PowerShell for `2>&1` redirection

Timeline:

```
PowerShell                    CLI Process                  Daemon Process
    |                             |                             |
    |-- create pipe (read+write)->|                             |
    |   (write end -> CLI stdout) |                             |
    |                             |                             |
    |   $output = & CLI ...       |                             |
    |   (waiting for pipe read    |                             |
    |    end to close)            |                             |
    |                             |-- spawn(daemon) ----------->|
    |                             |   bInheritHandles=TRUE      |
    |                             |                             |
    |                             |-- exit, close own handles ->|
    |                             |                             |
    |   pipe write end still held by daemon!                    |
    |   -> PowerShell keeps waiting --------------------------->|
    |                             |                             |
    |   until daemon also closes  |<-- close kills daemon ------|
    |   -> PowerShell returns     |                             |
```

## Proposed Fix
Use the Windows CreateProcessW API directly in cli/src/connection.rs to spawn the daemon with bInheritHandles = FALSE , preventing the daemon from inheriting any handles from the parent process.

Key changes:

1. New spawn_daemon_no_inherit function : Calls CreateProcessW directly with bInheritHandles = FALSE and explicit creation flags ( CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS | CREATE_BREAKAWAY_FROM_JOB | CREATE_UNICODE_ENVIRONMENT ).
2. New build_env_block function : Constructs a complete environment block (parent env vars + custom vars like AGENT_BROWSER_DAEMON=1 ), since passing a non-null lpEnvironment means the child only uses the provided variables.
3. New WindowsDaemonChild struct : Replaces std::process::Child on Windows. Only holds the process handle for detecting early daemon exit, without any pipe handles.
4. Modified ensure_daemon function : Uses spawn_daemon_no_inherit instead of Command::spawn() on Windows.

## Verification

| Scenario | Before Fix | After Fix |
|----------|-----------|-----------|
| Manual `open --headed` | Exits normally | Exits normally |
| `$output = & open --headed` | **Hangs, no exit** | Exits normally, `$LASTEXITCODE=0` |
| `close` then `open --headed` (script) | **Hangs, no exit** | Exits normally |
| `open` when daemon already running | Exits normally | Exits normally |
| Daemon startup failure detection | Reads stderr for error info | Shows exit code |

## Environment
- OS : Windows (Server 2022, also reproducible on Windows 10/11)
- agent-browser version : 0.27.0
- Rust : 1.96.0 (stable)
- PowerShell : 5.1 (also affects PowerShell 7.x)

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -40,7 +40,7 @@ rust-embed = "8"
 libc = "0.2"
 
 [target.'cfg(windows)'.dependencies]
-windows-sys = { version = "0.52", features = ["Win32_System_Threading", "Win32_Foundation"] }
+windows-sys = { version = "0.52", features = ["Win32_System_Threading", "Win32_Foundation", "Win32_System_Diagnostics_Debug"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -576,6 +576,140 @@ fn kill_stale_daemon(session: &str) {
     cleanup_stale_files(session);
 }
 
+#[cfg(windows)]
+struct WindowsDaemonChild {
+    process_handle: windows_sys::Win32::Foundation::HANDLE,
+}
+
+#[cfg(windows)]
+impl WindowsDaemonChild {
+    fn try_wait(&mut self) -> Option<u32> {
+        use windows_sys::Win32::System::Threading::{WaitForSingleObject, GetExitCodeProcess};
+        use windows_sys::Win32::Foundation::WAIT_OBJECT_0;
+
+        let result = unsafe { WaitForSingleObject(self.process_handle, 0) };
+        if result == WAIT_OBJECT_0 {
+            let mut exit_code: u32 = 0;
+            unsafe { GetExitCodeProcess(self.process_handle, &mut exit_code) };
+            Some(exit_code)
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(windows)]
+impl Drop for WindowsDaemonChild {
+    fn drop(&mut self) {
+        unsafe {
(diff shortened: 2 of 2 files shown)


OnePaiGu:
This PR fixes #1407 ,which addresses the Windows handle-inheritance leak.

lNimien:
Independent confirmation that this is still worth merging: the leak reproduces on **0.37.1**, so it has survived from 0.27.0 through roughly ten minor releases.

Measured on Windows 11 Pro 10.0.26200 with PowerShell 5.1, full numbers in [my comment on #1407](https://github.com/vercel-labs/agent-browser/issues/1407#issuecomment-5679312392): a cold `open` with output captured never returns (killed at 75s on 0.37.1), the same call wrapped in `cmd /c '... > log 2>&1'` completes in 1.05s, and a warm call with the daemon already up returns in 0.39s. That isolates the hang to the daemon spawning invocation, exactly as described here.

One detail that argues for fixing it at this layer rather than in the npm wrapper: on 0.37.1 the node shim still spawns the native binary with `stdio: 'inherit'` (`bin/agent-browser.js:115`), so it only forwards whatever handles the shell handed it. Setting `bInheritHandles = FALSE` at the `CreateProcessW` call is the only place that actually stops the daemon from capturing the caller's pipe.

PR #1409, opened 2026-06-03
fix: preserve encrypted state extension on rename

hiSandog:
## Summary
- keep renamed encrypted state files on the .json.enc suffix
- add a unit test covering encrypted state rename output

## Validation
- git diff --check
- Not run: cargo fmt/test because cargo/rustc are not available in this environment


hiSandog:

PR #1413, opened 2026-06-04
Fix iframe-blind snapshot/click; add snapshot --frames for transparent iframe discovery

simonellefsen:
## Problem

Two related iframe issues:

**1. Bug: `frame <sel>` context was silently ignored**
After switching into an iframe with `frame <sel>`, all subsequent commands (`snapshot`, `click`, `fill`, `get`, etc.) still targeted the main page. Three call sites in `BrowserManager` called `getPage()` instead of `getFrame()`, bypassing the active frame state entirely. Cross-origin iframes were completely invisible.

**2. UX: discovering iframe content required prior DOM knowledge**
Even with the bug fixed, you had to know an iframe existed, find its selector, run `frame <sel>`, then `snapshot`. Non-frontend developers had no way to discover that content was inside an iframe without inspecting the DOM manually.

### `snapshot --frames` / `-F` — the headline feature

One flag gets you everything: interactive elements from the main page and every iframe, merged into a single ref list, no knowledge of the DOM structure required:

```bash
# Before: had to know the iframe existed, find its selector, switch context manually
agent-browser open "https://danskespil.dk/klublotto/dagens-ordknuden"
agent-browser frame "iframe.kl-game__iframe"   # had to know this
agent-browser snapshot -i                      # only then could you see buttons

# After: one command, everything visible
agent-browser open "https://danskespil.dk/klublotto/dagens-ordknuden"
agent-browser snapshot -i -F
# - button "Log ind" [ref=e10]
# - link "Spil & Quiz" [ref=e6]
# ... (main page elements)
#
# # [iframe: iframe.kl-game__iframe]
# - button "q" [ref=e84]
# - button "w" [ref=e85]
# ...
# - button "Retur" [ref=e114]
# - button "" [ref=e115]    ← backspace

# And clicking just works — no frame switching needed at all
agent-browser click @e85    # types "w" inside the cross-origin iframe
agent-browser click @e114   # submits — refs are fully transparent
```

Refs from iframes carry a `frameSelector` internally. `getLocatorFromRef` uses `page.frameLocator(selector)` for those refs so the caller never needs to know or care that the element is inside an iframe.

### Bug fix: `frame <sel>` now actually works (`src/browser.ts`, `src/snapshot.ts`)

`getSnapshot`, `getLocatorFromRef`, and `getLocator` now call `getFrame()` instead of `getPage()`. `getEnhancedSnapshot` / `findCursorInteractiveElements` widened from `Page` to `Page | Frame`. The `frame` command is still useful when you want to work exclusively inside one iframe across many commands — it's more efficient than re-running `--frames` on every snapshot.

### Internals

`getEnhancedSnapshot` is refactored into a `snapshotFrame` inner function (no counter reset) plus a thin public wrapper, so `getMultiFrameSnapshot` can share one ref counter across all frames and avoid collisions between main and iframe refs.

### Help text fixes (`cli/src/output.rs`)

- `frame <sel|main>` was implemented in the Rust CLI but never added to `--help`
- `-C / --cursor` was parsed but undocumented
- `-F / --frames` added to both parser and help text

## Changed files

| File | Change |
|---|---|
| `src/browser.ts` | `getSnapshot`, `getLocatorFromRef`, `getLocator` use `getFrame()`; route `frames:true` to `getMultiFrameSnapshot`; use `frameLocator()` for iframe refs |
| `src/snapshot.ts` | `Page \| Frame` type; refactor into `snapshotFrame` + wrapper; add `getMultiFrameSnapshot`; add `frameSelector` to `RefMap` |
| `src/protocol.ts` | Add `frames` field to snapshot schema |
| `src/actions.ts` | Pass `frames` through to `browser.getSnapshot` |
| `cli/src/commands.rs` | Parse `-F / --frames` for snapshot command |
| `cli/src/output.rs` | Add `frame` to Navigation; add `-C` and `-F` to Snapshot Options |
| `.agents/…/commands.md` | Document `-C`, `-F`; expand Frames section |

## Test plan

- [ ] `snapshot -i -F` returns elements from main page and all iframes in one output, refs work with `click`/`fill`/`get` without any `frame` command
- [ ] `frame <sel>` + `snapshot -i` still works for single-iframe focus workflow
- [ ] `snapshot -i` without `-F` is unchanged (no regression, no extra latency)
- [ ] `--help` shows `frame` under Navigation, `-C` and `-F` under Snapshot Options
- [ ] Sandboxed/unloaded iframes are skipped silently (no crash)

### DIFF
--- a/.agents/skills/agent-browser/references/commands.md
+++ b/.agents/skills/agent-browser/references/commands.md
@@ -21,11 +21,15 @@ agent-browser connect 9222    # Connect to browser via CDP port
 agent-browser snapshot            # Full accessibility tree
 agent-browser snapshot -i         # Interactive elements only (recommended)
 agent-browser snapshot -C         # Also include cursor:pointer / onclick elements
-agent-browser snapshot -i -C      # Interactive + cursor-interactive (widest net)
+agent-browser snapshot -F         # Include elements from all iframes automatically
+agent-browser snapshot -i -F      # Interactive elements from page + all iframes
+agent-browser snapshot -i -C -F   # Widest net: interactive + cursor + iframes
 agent-browser snapshot -c         # Compact output
 agent-browser snapshot -d 3       # Limit depth to 3
 agent-browser snapshot -s "#main" # Scope to CSS selector
 ```
+
+`-F / --frames` discovers all iframes automatically and merges their elements into the output. Refs from iframes work transparently with `click`, `fill`, `get`, etc. — no manual `frame` switching required. Use the `frame` command when you want to work exclusively inside one iframe.
 
 When inside an iframe (after `agent-browser frame <sel>`), snapshot and all interactions target the iframe's document — use `frame main` to return to the main page.
 
--- a/.agents/skills/agent-browser/references/commands.md
+++ b/.agents/skills/agent-browser/references/commands.md
@@ -1,0 +1,278 @@
+# Command Reference
+
+Complete reference for all agent-browser commands. For quick start and common patterns, see SKILL.md.
+
+## Navigation
+
+```bash
+agent-browser open <url>      # Navigate to URL (aliases: goto, navigate)
+                              # Supports: https://, http://, file://, about:, data://
+                              # Auto-prepends https:// if no protocol given
+agent-browser back            # Go back
+agent-browser forward         # Go forward
+agent-browser reload          # Reload page
+agent-browser close           # Close browser (aliases: quit, exit)
+agent-browser connect 9222    # Connect to browser via CDP port
+```
+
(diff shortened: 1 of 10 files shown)


simonellefsen:
This PR targets Node.js TypeScript files that no longer exist in the codebase. The project has migrated to a pure Rust CLI architecture as of v0.27.0. 

The iframe frame-discovery feature proposed in this PR is no longer applicable to the new Rust architecture, which handles frames natively through its element tracking system. The Rust CLI already has frame context awareness built in (see cli/src/native/element.rs).

Closing in favor of focusing on the active Rust-based architecture.

PR #1419, opened 2026-06-06
fix: parse unquoted cookie flags

hiSandog:
## Summary
- allow cookie import to parse unquoted cURL cookie arguments like `-b sid=abc`
- support unquoted `--cookie token=xyz` values while preserving quoted header handling
- add unit coverage for the new parsing cases

## Validation
- `git diff --check`
- Not run: `cargo test parse_curl_cookies` because `cargo` is not installed in this environment


hiSandog:

PR #1425, opened 2026-06-08
Ready: fix(window): open `window new` in the default browser context

soichisumi:
## Summary

- Make `window new` open in the browser's **default** context, so it inherits the profile's existing login, instead of a fresh incognito-like context that is logged out.
- Preserve the old behavior behind a new `window new --isolated` flag.
- Give `CreateTargetParams` typed `newWindow` / `background` / `focus` / `browserContextId` fields + a `CreateTargetParams::new(url)` constructor.

## Why

`window new` creates a fresh browser context and opens the page inside it (`handle_window_new`): `Target.createBrowserContext` then `Target.createTarget { browserContextId }`. That context is incognito-like — it does **not** share the profile's cookies / localStorage / login. So when you attach to a logged-in Chrome (`--cdp` / `--profile`) and run `window new`, the new window is **logged out**, and automating a logged-in site in it fails. Same root cause as #890.

Repro — attach to a Chrome already logged into some site, then:

```
agent-browser --cdp <port> window new
agent-browser --cdp <port> tab <new-tab-id>
agent-browser --cdp <port> open https://<logged-in-site>/   # logged out
```

## What

- `handle_window_new` no longer calls `Target.createBrowserContext` by default; it creates the target in the default context.
- New opt-in `window new --isolated` reproduces the previous isolated-context behavior. The docs only described "New window" (never isolation), so the old default was effectively accidental.
- `CreateTargetParams` gains optional `newWindow` / `background` / `focus` / `browserContextId` (camelCase, omitted when `None`) and a `::new(url)` constructor; the four existing url-only call sites are migrated to it.
- Unit tests for the serialization contract and `window new [--isolated]` parsing.

## How

Open the window with `Target.createTarget { url, newWindow: true, focus: false, background: false }` and **no** `browserContextId`:

- `newWindow: true` -> a real OS window (not a tab).
- `focus: false` (+ `background: false`) -> opens without stealing OS focus from the user's foreground window.
- no `browserContextId` -> created in the default context -> login preserved.

The `::new(url)` constructor leaves all options `None`, so the migrated url-only call sites stay byte-for-byte identical on the wire.

Caveat for reviewers: per the Page Visibility spec the selected tab of a non-minimized, non-fully-occluded window stays `visible` even without OS focus, but a minimized / fully-occluded window still maps to `hidden`; `focus: false` is experimental and worth a sanity check on Linux/Windows; `newWindow` is not supported by the headless shell.

## Test plan

- [x] [auto] `cargo test --bin agent-browser` — `create_target_params_*` (serialization, incl. `browserContextId` omitted by default) and `test_window_new` / `test_window_new_isolated` / `test_window_new_rejects_unknown_flag` pass.
- [x] [one-shot-e2e] Built `--release` and attached to a logged-in Chrome over CDP: `window new` -> owned window reports `loggedIn: true`; `window new --isolated` -> `loggedIn: false`. Single run; the behavioral CDP path is not in CI.
- [x] [manual] Verified the four migrated `CreateTargetParams` call sites still serialize url-only via `::new()`.

Refs #890. Future work (out of scope): per-command tab addressing + per-tab ref maps (cf. closed #306) for full multi-client safety.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1538,14 +1538,30 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
         "window" => {
             const VALID: &[&str] = &["new"];
             match rest.first().copied() {
-                Some("new") => Ok(json!({ "id": id, "action": "window_new" })),
+                Some("new") => {
+                    // `window new` opens a window in the shared (logged-in)
+                    // context; `--isolated` opts into a fresh isolated context.
+                    let mut cmd = json!({ "id": id, "action": "window_new" });
+                    for arg in rest.iter().skip(1) {
+                        match *arg {
+                            "--isolated" => cmd["isolated"] = json!(true),
+                            other => {
+                                return Err(ParseError::UnknownSubcommand {
+                                    subcommand: other.to_string(),
+                                    valid_options: &["--isolated"],
+                                });
+                            }
+                        }
+                    }
+                    Ok(cmd)
+                }
                 Some(sub) => Err(ParseError::UnknownSubcommand {
                     subcommand: sub.to_string(),
                     valid_options: VALID,
                 }),
                 None => Err(ParseError::MissingArguments {
                     context: "window".to_string(),
-                    usage: "window <new>",
+                    usage: "window <new> [--isolated]",
                 }),
             }
         }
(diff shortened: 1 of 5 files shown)


soichisumi:
This fix is really needed for me and I think this feature suites for agent-browser(new window inherit current targetted browsers context). plz review~ 🙏 
cc: @ctate

joshuavial:
Landing lane: claiming this PR stack (1425, 1426, 1427) for processing.

joshuavial (approved):
Verification passed. Enabling auto-merge.

soichisumi:

PR #1427, opened 2026-06-08
Ready: feat: keep non-foreground tabs operable (focus-emulate + tab new --background)

soichisumi:
## Summary

- `focus-emulate <on|off>` — emulate a focused/visible page on the active tab so automation works on a backgrounded tab.
- `tab new --background` — open a new tab without foregrounding it.
- Together they let an agent open and drive a tab in the background, without raising windows or stealing the user's current tab/focus.

## Why

When automating a browser that a human is also using, an agent often needs to work on a tab that is **not** the foreground tab. Chrome backgrounds non-foreground tabs: `document.visibilityState` becomes `hidden` and rendering / `requestAnimationFrame` is throttled, so clicks and visibility/focus-gated handlers silently no-op. The usual workaround — foregrounding the tab (`Page.bringToFront`) — steals the user's current tab, which is disruptive when a person is sharing the browser.

## What

- **`focus-emulate <on|off>`**: a new command that toggles CDP `Emulation.setFocusEmulationEnabled` on the active tab.
- **`tab new --background`**: a new flag on `tab new` that creates the target with CDP `background: true`. The tab still becomes the daemon's active page, so subsequent commands target it.
- Unit tests for both command parsers, and docs/help updates for both.

## How

- `focus-emulate on` sends `Emulation.setFocusEmulationEnabled { enabled: true }` to the active page's session, so it reports as focused + visible (and resumes rendering) even while backgrounded — without raising the window or taking OS focus. `off` reverts it.
- `tab new --background` passes `background: true` to `Target.createTarget` (built with `serde_json::json!`, matching the existing ad-hoc `createTarget` call), so the tab opens without foregrounding. No change to the typed `CreateTargetParams`, keeping the diff isolated.

## Test plan

- `cargo test --bin agent-browser` — parser tests pass: `test_focus_emulate_on` / `test_focus_emulate_off` / `test_focus_emulate_requires_on_or_off`, `test_tab_new_background` / `test_tab_new_background_with_url`.
- Built and run against a real Chrome over CDP: a background tab reports `visibilityState=hidden` and `requestAnimationFrame ≈ 0` ticks/s; after `focus-emulate on` it reports `visible`, `hasFocus()=true`, and `≈60` ticks/s. `tab new --background` opens a tab without foregrounding it, and the new tab is still the active page for subsequent commands.
- Docs/help updated for both new commands (`README.md`, `docs/src/app/commands/page.mdx`, `skill-data/core/references/commands.md`, `cli/src/output.rs`).

Note: this does not reintroduce scoped per-command tab dispatch (the removed `--tab`, #1250) — there is no per-command tab routing here; both additions act on the active tab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,9 +327,11 @@ agent-browser network har stop [output.har]    # Stop and save HAR (temp path if
 agent-browser tab                              # List tabs (shows `tabId` and optional label)
 agent-browser tab new [url]                    # New tab (optionally with URL)
 agent-browser tab new --label docs [url]       # New tab with a user-assigned label
+agent-browser tab new --background [url]        # New tab without foregrounding it
 agent-browser tab <t<N>|label>                 # Switch to a tab by id or label
 agent-browser tab close [t<N>|label]           # Close a tab (defaults to active)
 agent-browser window new                       # New window
+agent-browser focus-emulate <on|off>           # Keep a backgrounded tab operable (focus/visibility emulation)
 ```
 
 Tab ids are stable strings of the form `t1`, `t2`, `t3`. They're never reused within a session, so scripts and agents can keep referring to the same tab even after other tabs are opened or closed. Positional integers like `tab 2` are **not** accepted; the `t` prefix disambiguates handles from indices and mirrors the `@e1` convention used for element refs.
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -945,6 +945,24 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
         // === Inspect ===
         "inspect" => Ok(json!({ "id": id, "action": "inspect" })),
 
+        // === Focus emulation ===
+        // Make the active tab behave as a focused, visible page even while it is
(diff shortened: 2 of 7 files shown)


soichisumi:
also this is needed 🙏 
Cannot upgrade agent-browser wo these features https://github.com/vercel-labs/agent-browser/pull/1425 https://github.com/vercel-labs/agent-browser/pull/1426 https://github.com/vercel-labs/agent-browser/pull/1427

cc: @ctate

soichisumi:

soichisumi:
Rebased onto `main` at v0.34.0. Conflicts were confined to `cli/src/native/actions.rs` (restructured by #1589) — I kept main's `defer_url_until_controls` / pinned-session logic and only threaded the new `background` argument through `tab_new`.

Scope unchanged: `focus-emulate` + `tab new --background`, so non-foreground tabs still commit React state updates. No per-command tab routing — this composes with #1589's `--pin-tab` rather than overlapping it.

Happy to have this folded into a maintainer batch with credit like #1426, or split into two smaller PRs — whichever is easier for you.

PR #1440, opened 2026-06-11
fix: add SHA-256 checksum verification for downloaded binaries

### LINKED ISSUE #1422: scripts/postinstall.js downloads native binary without checksum verification
## Summary

The postinstall script fetches a platform-specific binary from GitHub Releases over HTTPS but performs no integrity check on the downloaded file. If the release artifacts were ever replaced (compromised GH account, supply chain attack on the release pipeline, or a MITM on the download), the malicious binary would be silently installed and executed with the user's permissions.

## Affected file

`scripts/postinstall.js`

The `downloadFile()` function validates only the HTTP status code (200 vs redirect) but performs no subsequent verification of the binary's authenticity or integrity:

```js
// Current behaviour — only checks HTTP status
if (response.statusCode !== 200) { ... }
// binary is written to disk and chmod'd with no hash check
```

## What's missing

After writing the binary to disk, the script should verify its SHA-256 hash against a known-good value shipped in the package itself. This is the standard approach used by similar native-binary packages:

- **esbuild** — ships per-platform hashes in each platform sub-package, verified by npm's integrity check
- **turbo** — same pattern

## Suggested fix

Add a `CHECKSUMS` map keyed by `${platform}-${arch}`, compute `sha256` of the downloaded buffer, and throw before `chmod`-ing if it doesn't match:

```js
const crypto = require('crypto');

const CHECKSUMS = {
  'linux-x64':   'abc123...',
  'darwin-arm64': 'def456...',
  // ...
};

const hash = crypto.createHash('sha256').update(downloadedBuffer).digest('hex');
const expected = CHECKSUMS[`${platform}-${arch}`];
if (hash !== expected) {
  throw new Error(`Binary checksum mismatch for ${platform}-${arch}.\nExpected: ${expected}\nGot:      ${hash}`);
}
```

The checksums would be generated as part of the release pipeline and committed to the repo (or embedded in `package.json`) alongside each release.

## Impact

Low likelihood day-to-day (requires a compromised Vercel Labs release pipeline or GH account), but high impact if triggered — the binary runs with the installing user's full permissions immediately on `npm install`. Adding checksum verification is a low-effort, high-value hardening step.

axelray-dev:
Fixes #1422

## What changed

Added SHA-256 checksum verification for the downloaded native binary in `scripts/postinstall.js`. After downloading, the script fetches the `SHA256SUMS` file from the same GitHub release and verifies the hash before making the binary executable.

If the `SHA256SUMS` file is not available (older releases), the install proceeds with a warning. If the checksum file is present but the hash does not match, the downloaded binary is deleted and the install fails.

Also added `scripts/generate-checksums.sh` to produce the `SHA256SUMS` file for a given release. This can be run manually or integrated into the release pipeline later.

## Changes

- `scripts/postinstall.js`: added `crypto` and `createReadStream` imports, `fetchText()` helper for HTTPS GET with redirect handling, `parseChecksums()` to parse the standard `<hash>  <filename>` format, `sha256File()` to compute the hash of a file on disk, `verifyChecksum()` to orchestrate the check, and a verification call after `downloadFile()` in `main()`.
- `scripts/generate-checksums.sh`: new script that downloads all 7 platform binaries from a release and writes `SHA256SUMS`.

## Tests

- `node --check scripts/postinstall.js` passes (no syntax errors).
- `bash -n scripts/generate-checksums.sh` passes.
- Ran `node scripts/postinstall.js` locally: the missing `SHA256SUMS` path is handled gracefully with a warning, and the binary download continues as before.
- No new dependencies; uses Node.js built-in `crypto` module.

## Notes

The `SHA256SUMS` file does not exist in current releases yet. The verification is backward-compatible: it warns and proceeds when the file is missing. Once the release pipeline generates `SHA256SUMS` alongside the binaries, the verification will enforce integrity automatically.

### DIFF
--- a/scripts/generate-checksums.sh
+++ b/scripts/generate-checksums.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+set -euo pipefail
+
+# Download release binaries and generate SHA256SUMS.
+# Usage: ./scripts/generate-checksums.sh [version]
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PROJECT_ROOT="$(dirname "$SCRIPT_DIR")"
+OUTPUT_FILE="$PROJECT_ROOT/SHA256SUMS"
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "$TMP_DIR"' EXIT
+
+GITHUB_REPO="vercel-labs/agent-browser"
+
+if [ "${1:-}" ]; then
+  VERSION="${1#v}"
+else
+  VERSION="$(
+    curl -fsSL "https://api.github.com/repos/$GITHUB_REPO/releases/latest" \
+      | node -e "let data=''; process.stdin.on('data', c => data += c); process.stdin.on('end', () => console.log(JSON.parse(data).tag_name.replace(/^v/, '')));"
+  )"
+fi
+
+BASE_URL="https://github.com/$GITHUB_REPO/releases/download/v$VERSION"
+
+BINARIES=(
+  "agent-browser-linux-x64"
+  "agent-browser-linux-arm64"
+  "agent-browser-win32-x64.exe"
+  "agent-browser-darwin-x64"
+  "agent-browser-darwin-arm64"
+  "agent-browser-linux-musl-x64"
+  "agent-browser-linux-musl-arm64"
+)
+
+echo "Generating checksums for v$VERSION"
+rm -f "$OUTPUT_FILE"
+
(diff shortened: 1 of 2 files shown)


axelray-dev:
Closing this draft while I rethink the approach. The current implementation adds runtime checksum verification but does not address how checksums should be generated and distributed through the release pipeline. I want to confirm the intended trust model before proceeding -- specifically whether checksums should come from trusted build artifacts shipped inside the npm package, and how release.yml would need to change. Happy to reopen with a revised design once there is maintainer direction on the architecture.

Issue #1351, opened 2026-05-13
High LLM turn count due to frequent `snapshot` calls when using `agent-browser` skills


nttdata-jiang:
# Issue: High LLM turn count due to frequent `snapshot` calls when using `agent-browser` skills (Windows 11, LangChain E2E agent)

## Summary
I migrated a local E2E test agent from **LangChain + MCP Playwright tools + LLM** to **`agent-browser` skills** to reduce prompt/token size.

While token usage improved, I observed a **significant increase in LLM tool-interaction turns**, and logs show a large portion of those turns are `snapshot` calls.

I’d like to ask whether there is room for improvement on the `agent-browser` side (or recommended best practices) to reduce the number of explicit `snapshot` requests required by the LLM—e.g., returning an updated element reference tree automatically after UI actions.

## Background
- I have an E2E testing agent driven by an LLM.
- Input is a test case file with **natural-language steps** and **expected results**.
- The agent operates a real browser to execute the test.

I originally used **MCP Playwright** as the browser automation toolset. To reduce token consumption, I switched browser automation to **`agent-browser skills`**.

## Environment
- OS: Windows 11
- Node.js: >= 24.14.0
- `agent-browser`: `^0.26.0` (npm dependency)
- Orchestration: LangChain (tools) + LLM (agent)
- Baseline toolchain: MCP Playwright tools
- New toolchain: `agent-browser` (skill-driven CLI usage)

## What I changed
- Before: LLM calls MCP Playwright tools for browser operations.
- After: LLM calls a restricted tool that executes `agent-browser ...` commands according to `agent-browser skills get core`.

## Metrics (same cases + same environment)
I ran ~150 test cases under the same environment and compared the total number of LLM invocations/tool-interaction turns.

| Item | Agent Browser | Playwright | Diff (absolute) | Diff (%) |
| --- | ---:| ---:| ---:| ---:|
| LLM call count | 3,351 | 2,028 | +1,323 | +65.2% |

## Observation
From execution logs, roughly **~30% of the tool calls** are `snapshot` during the `agent-browser` run.

I understand why `snapshot` is required:
- The reference tree (`@eN` refs) can become stale after navigation or DOM updates.
- The official skills also recommend re-snapshotting after page changes.

However, in practice this causes a **large increase in the number of LLM turns**, which has a direct impact on:
- total runtime
- cost
- rate-limit pressure

## Question / Request
[...]

Issue #853, opened 2026-03-16
SPA Scraping Pattern: Preserving accessibility refs with tab isolation


ejaircastillo:
# SPA Scraping Pattern: Preserving accessibility refs with tab isolation

## Summary

After extensive testing with Google Maps accessibility tree scraping, I discovered a critical pattern for sustainable ref-based web scraping on SPAs with dynamic feeds. The solution: **never navigate the source tab, always use new tabs for details**.

## Problem Statement

When scraping Google Maps search results using `agent-browser`'s accessibility tree snapshots and sequential item extraction:

**Initial state (first snapshot):**
```text
article "Hostel Siri"[ref=e36]
article "Hotel Santa Cruz" [ref=e37]
article "Hostel Elcira" [ref=e38]
article "Dptos El Buen Descanso"[ref=e39]
article "Hotel Laguna Azul" [ref=e40]
```

**After: click e36 → view detail → back()**
```text
article "Hotel Patagonia"[ref=e26] ← NEW item appeared!
article "Hotel Santa Cruz" [ref=e27] ← All refs reassigned
article "Hostel Siri" [ref=e28]
article "Dptos El Buen Descanso" [ref=e29]
article "Hostel Elcira"[ref=e30] ← Position changed
article "Hotel Laguna Azul" [ref=e31]
```

**Result:** Sequential extraction fails at item 3+ because refs are invalid and items have reordered.

**Root cause:** Google Maps (and similar SPAs) completely re-render the feed DOM on return from detail pages, reassigning all accessibility tree refs.

## Solution: Tab Isolation Pattern

**Core principle:** Keep tab 0 (the search results feed) untouched. Navigate details in ephemeral tabs.

### Implementation

```typescript
// 1. Initial snapshot in tab 0 (never leaves this tab)
const listings = extractListingRefs(ab("snapshot -i -c"));

// 2. For each item
for (const name of listings.map(l => l.name)) {
  // Re-snapshot tab 0 to get fresh ref (resilient to lazy-load)
  const feedSnapshot = ab("snapshot -i -c");
  const current = extractListingRefs(feedSnapshot).find(l => l.name === name);
  
  // Open detail in new tab (tab 0 stays untouched)
  ab(`click "@${current.linkRef}" --new-tab`);
  ab("wait 2000");
  ab("tab 1");  // Switch to detail (always tab 1 for first detail)
  
  // Extract from detail page
[...]

PR #1444, opened 2026-06-12
Fix stale snapshot refs silently actuating the wrong element after DOM-mutation renumbering

### LINKED ISSUE #1443: Stale snapshot refs silently actuate the wrong element after DOM-mutation renumbering
**Version:** agent-browser 0.27.2 (macOS arm64, Chrome 149; also reproduced on 0.25.3 Linux/Docker)

## Summary

Snapshot element refs (`e1`, `e2`, …) are positional within the latest snapshot's registry. When the DOM mutates and a new snapshot renumbers the registry, a ref remembered from the *previous* snapshot resolves against the *new* numbering — and silently clicks/fills whichever element now occupies that number, returning `✓ Done` / exit 0.

Out-of-range refs already fail safely (`✗ Unknown ref: e9` in `cli/src/native/element.rs`) — this ask is narrowly the **remembered-ref-after-renumbering** case: the ref still exists, but denotes a different element than when it was emitted. For LLM agents, that's the worst failure shape: not an error, not a no-op, but a wrong action reported as success.

## Minimal repro (5 commands, no real site needed)

```bash
agent-browser open 'data:text/html,<button onclick="document.title=this.textContent">Alpha</button><button onclick="document.title=this.textContent">Beta</button><button onclick="document.title=this.textContent">Gamma</button>'

agent-browser snapshot -i
# - button "Alpha" [ref=e1]
# - button "Beta"  [ref=e2]   <- agent remembers: e2 == Beta
# - button "Gamma" [ref=e3]

# Page inserts a node (what cmdk/Radix multi-selects do on every committed selection)
agent-browser eval "document.body.insertBefore(Object.assign(document.createElement('button'),{textContent:'NEW'}),document.body.firstChild); 'ok'"

agent-browser snapshot -i
# - button "NEW"   [ref=e1]
# - button "Alpha" [ref=e2]   <- everything shifted +1
# - button "Beta"  [ref=e3]
# - button "Gamma" [ref=e4]

agent-browser click @e2      # agent still believes e2 == Beta
# ✓ Done
agent-browser get title
# Alpha                       <- wrong element clicked, success reported
```

(Without the second snapshot, `click @e2` correctly hits Beta — the old registry mapping is honored. The hazard is specifically remembered-ref + renumbering.)

## Real-world impact

Combobox/multi-select widgets (cmdk, Radix, HeadlessUI) insert a DOM node per committed selection, shifting every later ref +1 — an LLM agent selecting multiple options from one snapshot silently checks *adjacent* options on each subsequent click. In a large agent-automation audit of ours, this mechanism shipped wrong category selections to multiple live product-directory listings, each wrong click acknowledged with `✓ Done`. #853 documents the same renumbering in the wild (Google Maps feed re-render: `Hostel Siri` e36→e28, sequential extraction grabbing wrong items), and #1351's measurement that ~30% of agent calls are defensive re-snapshots "because refs can become stale" quantifies the tax agents pay working around it.

## Prior art

- **PR #1112 ("Make refs strict across snapshots")** described this exact defect and fix shape — *"Fail stale refs instead of silently rebinding them after DOM changes, and preserve ref ids for surviving nodes across repeated snapshots"* — but was self-closed seconds after opening and never reviewed; no successor exists.
- **#806** (merged, v0.20.6) handles the *sibling* mode — a ref whose cached `backend_node_id` died is re-queried by role+name — but doesn't apply here: after renumbering, the registry entry is perfectly valid, just bound to a different element.
- **PR #1249** (merged) fixed `ref_map` leakage across tab switches precisely because "it could click the wrong one" — precedent that silent wrong-element actuation is treated as a bug.
- **v0.26.0 stable tab ids** rejected bare integers "so agents can't mistake stable handles for positional indices" — exactly the right principle; element refs (which the `t1` convention explicitly mirrors, per PR #1250) are the remaining positional handles.

## Proposals (either resolves it)

1. **Bind refs to stable node identity** — mint refs against `backendNodeId` and preserve ids for surviving nodes across snapshots (PR #1112's shape). A remembered ref keeps meaning the node it was minted for; if detached, fail like `Unknown ref` does today: `✗ stale ref: node no longer in DOM — take a new snapshot`. As a bonus this directly reduces the defensive re-snapshot churn measured in #1351.

2. **Snapshot epoch validation** — emit a generation marker with each snapshot (in the ref, e.g. `e42@s7`, or a header line) and have action commands reject refs minted under an older generation with a teaching error, mirroring the bare-integer tab id error: `✗ ref e2 is from snapshot 6; the registry renumbered (now 7) — re-snapshot and use fresh refs`.

Happy to contribute a PR for either direction if maintainers indicate a preference.

iddogino:
Fixes #1443

## Problem

Snapshot refs (`e1`, `e2`, ...) were positional in the latest snapshot's registry. When the DOM mutated and a new snapshot renumbered the registry, a ref remembered from the previous snapshot resolved against the new numbering, silently clicking or filling whichever element now occupied that number and returning `✓ Done`. For LLM agents this is the worst failure shape: a wrong action reported as success (wrong multi-select options checked, wrong listings extracted; see #853 and the defensive re-snapshot tax measured in #1351).

```
agent-browser snapshot -i      # button "Beta" [ref=e2]   <- agent remembers e2 == Beta
agent-browser eval "...insert a node before the buttons..."
agent-browser snapshot -i      # everything shifted +1
agent-browser click @e2        # clicked Alpha, reported ✓ Done
```

## Fix

Bind refs to node identity (the shape described in PR #1112). Taking a snapshot now retires the previous ref generation instead of clearing it:

- **Surviving nodes reclaim their original ref id**, matched by `backendNodeId` + frame id. A remembered ref keeps meaning the node it was minted for.
- **New nodes get fresh numbers, and ref numbers are never reused within a page**, so a retired number can never be silently reoccupied.
- **Refs whose node left the DOM fail loudly** with a teaching error, mirroring the stable tab id error pattern:
  `Stale ref: e2 was minted by an earlier snapshot and its element is no longer in the DOM. Take a new snapshot and use a fresh ref.`
  Never-minted refs keep the existing `Unknown ref` error.
- Navigation, reload, tab switches, and launch still clear the registry entirely (the #1249 behavior is unchanged).
- The role+name re-query fallback for refs whose backend node died between snapshot and action (#806) is unchanged, and its e2e test still passes.

After the fix, the repro from the issue gives:

```
agent-browser snapshot -i
# - button "NEW"   [ref=e4]   <- new node gets a fresh ref
# - button "Alpha" [ref=e1]   <- surviving nodes keep their refs
# - button "Beta"  [ref=e2]
# - button "Gamma" [ref=e3]
agent-browser click @e2
agent-browser get title        # Beta
```

As a side effect, stable ref ids across snapshots reduce the defensive re-snapshot churn measured in #1351 and make repeated `snapshot` output diffable across re-renders.

## Tests

- Three e2e regression tests covering the issue's exact repro: a remembered ref keeps meaning the same element after a re-snapshot, surviving nodes keep their ids (new node gets a fresh one), and a removed node's ref fails loudly instead of actuating a neighbor. All three fail on `main` and pass with the fix.
- Seven `RefMap` unit tests for the retire/reclaim lifecycle: id preservation, stale vs unknown errors, numbers never reused, a node that disappears and reappears reclaims its original id, and full clear on navigation.
- Full suite: 745 unit tests and 73/74 e2e tests pass (the one failure is `e2e_recording_inherits_viewport`, which needs ffmpeg that the test environment lacks; unrelated to this change). `cargo fmt --check` and `cargo clippy -- -D warnings` are clean.

## Docs

Updated per AGENTS.md: README ref section, `skill-data/core/SKILL.md`, `skill-data/core/references/snapshot-refs.md` (ref lifecycle + troubleshooting), and the docs site selectors page now describe identity-bound refs, the stale-ref error, and that navigation still clears all refs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -897,6 +897,13 @@ control first, then run `snapshot` again before reusing refs.
 - **Fast**: No DOM re-query needed
 - **AI-friendly**: Snapshot + ref workflow is optimal for LLMs
 
+Refs are bound to the element they were minted for and survive repeated
+snapshots within a page: elements still in the DOM keep their ref ids, new
+elements get fresh ids, and ref numbers are never reused. Using a ref whose
+element left the DOM after a newer snapshot fails with a stale-ref error
+(`Stale ref: e2 ... Take a new snapshot`) instead of acting on a different
+element. Navigation clears all refs.
+
 ### CSS Selectors
 
 ```bash
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -2547,7 +2547,11 @@ async fn handle_snapshot(cmd: &Value, state: &mut DaemonState) -> Result<Value, 
         urls: cmd.get("urls").and_then(|v| v.as_bool()).unwrap_or(false),
     };
 
-    state.ref_map.clear();
+    // Retire (rather than clear) the previous generation of refs: surviving
+    // nodes reclaim their ref ids inside take_snapshot, and refs whose nodes
+    // are gone fail with a stale-ref error instead of silently rebinding to
+    // whichever element now holds their number (#1443).
+    state.ref_map.begin_rebuild();
     let tree = snapshot::take_snapshot(
         &mgr.client,
         &session_id,
@@ -2650,7 +2654,7 @@ async fn handle_screenshot(cmd: &Value, state: &mut DaemonState) -> Result<Value
     };
 
     if annotate {
-        state.ref_map.clear();
+        state.ref_map.begin_rebuild();
         let _ = snapshot::take_snapshot(
             &mgr.client,
             &session_id,
@@ -5705,10 +5709,7 @@ async fn handle_frame(cmd: &Value, state: &mut DaemonState) -> Result<Value, Str
     // If selector is a ref (@e1), resolve the iframe element from the ref map
     if let Some(sel) = selector {
         if let Some(ref_id) = super::element::parse_ref(sel) {
-            let entry = state
-                .ref_map
-                .get(&ref_id)
-                .ok_or_else(|| format!("Unknown ref: {}", ref_id))?;
+            let entry = state.ref_map.resolve(&ref_id)?;
             let backend_node_id = entry
                 .backend_node_id
                 .ok_or_else(|| format!("Ref {} has no backend node id", ref_id))?;
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -3073,6 +3073,280 @@ async fn e2e_click_stale_ref_falls_back_to_role_name() {
 }
 
 // ---------------------------------------------------------------------------
+// Regression: stale snapshot refs after DOM-mutation renumbering (#1443)
+//
+// Snapshot refs were positional within the latest snapshot's registry. When
+// the DOM mutated and a new snapshot was taken, every ref shifted, so a ref
+// remembered from the previous snapshot silently actuated whichever element
+// now occupied that number. Refs must instead stay bound to the node they
+// were minted for: surviving nodes keep their ref ids across snapshots and
+// refs whose nodes left the DOM fail loudly instead of rebinding.
+// ---------------------------------------------------------------------------
+
+const RENUMBER_PROBE_URL: &str = "data:text/html,\
+    <button onclick=\"document.title=this.textContent\">Alpha</button>\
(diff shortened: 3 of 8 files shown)


iddogino:
@ctate any thoughts on the approach here? This is a problem we have been running into a lot with agentic loops, would love to see if there is a way to make refs more stable...

PR #1450, opened 2026-06-15
Fix verbose flag argument cleanup

hiSandog:
Summary:
- Treat -v/--verbose and -q/--quiet as bare flags during argument cleanup.
- Preserve command arguments named true or false after those flags.
- Add regression coverage for verbose/quiet cleanup.

Validation:
- git diff --check
- Not run: cargo fmt or cargo test, because cargo is not available in this environment.


hiSandog:
Closing this as part of PR maintenance because it has been open for more than 15 days. Leaving the branch intact.

PR #1451, opened 2026-06-15
fix: reap failed Chrome launch child

### LINKED ISSUE #1401: Bug: failed Chrome launch can leave zombie child processes unreaped on Linux
## Summary

On Linux, failed Chrome launch attempts can leave unreaped zombie child processes under the `agent-browser` daemon.

The root cause appears to be a specific launch-failure path in `cli/src/native/cdp/chrome.rs`: when `try_launch_chrome()` falls back after `DevToolsActivePort` failure and then errors again, it calls `child.kill()` but does not `wait()` on that child before returning. Because `ChromeProcess` is never constructed on that path, the existing process-group cleanup in `ChromeProcess::kill()` / `Drop` does not run either.

That means repeated failed launch attempts can accumulate `[chrome] <defunct>` entries under the daemon parent.

## Environment

- OS: Ubuntu 24.04.4 LTS
- Arch: arm64
- agent-browser: 0.27.0
- Invocation context: daemon launched by an agent runner, but the root cause is in the generic native Chrome launch path

## Observed behavior

- A detached `agent-browser` daemon remained alive after interrupted browser work.
- Its child list contained many zombie `chrome` processes.
- In my case they accumulated under a single daemon PID and showed up as `[chrome] <defunct>`.
- This was reproducible from source inspection and local testing without relying on the original agent wrapper.

## Source-level root cause

The problematic path is in `try_launch_chrome()`:

- `wait_for_devtools_active_port()` fails
- stderr fallback also fails
- error path calls `child.kill()`
- function returns without `child.wait()`

Because the launch wrapper never reaches `ChromeProcess { .. }`, the later `Drop`-based cleanup is skipped.

## Minimal reproduction idea

One simple way to reproduce the bug in tests is to force a launch failure with an executable that exits immediately, for example `/bin/false`, and call `launch_chrome()`.

Repeated failures can leave unreaped child processes unless the failure path explicitly waits on the child.

## Expected behavior

A failed launch attempt should:

- kill the spawned Chrome process group
- reap the child before returning
- leave no zombie child processes behind

## Actual behavior

The failed launch path can return while the spawned child is still unreaped, leaving a zombie process behind.

## Related issues

This looks adjacent to, but more specific than:

- #1397
- #1371
- #1334

Those issues describe broader interruption/orphan cleanup problems. This issue is about a concrete unreaped-child bug in the native Chrome launch error path.

sjh9714:
## Summary

Fixes #1401.

Failed native Chrome launches could return after calling `child.kill()` without waiting on the spawned child. Since the error path never constructs `ChromeProcess`, the normal drop cleanup did not run, so failed launch attempts could leave unreaped Chrome children behind.

This adds a small failed-launch cleanup helper that kills the spawned child/process group and waits for the child before returning from the DevToolsActivePort/stderr fallback failure paths.

## Testing

- `PATH="$HOME/.cargo/bin:$PATH" cargo test test_terminate_failed_chrome_launch_reaps_child`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test native::cdp::chrome`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt -- --check`
- `PATH="$HOME/.cargo/bin:$PATH" cargo clippy`
- `git diff --check`

I also ran `PATH="$HOME/.cargo/bin:$PATH" cargo test`; it reached unrelated `native::stream::http` tests that fail in this checkout because the generated Unix socket path under the current workspace exceeds macOS `SUN_LEN` (`path must be shorter than SUN_LEN`).

Note: I used Codex while preparing this change, and reviewed the final diff and checks locally.

### DIFF
--- a/cli/src/native/cdp/chrome.rs
+++ b/cli/src/native/cdp/chrome.rs
@@ -60,6 +60,18 @@ impl ChromeProcess {
     }
 }
 
+fn terminate_failed_chrome_launch(child: &mut Child) {
+    let _ = child.kill();
+    #[cfg(unix)]
+    {
+        let pgid = child.id() as i32;
+        unsafe {
+            libc::kill(-pgid, libc::SIGKILL);
+        }
+    }
+    let _ = child.wait();
+}
+
 impl Drop for ChromeProcess {
     fn drop(&mut self) {
         self.kill();
@@ -401,7 +413,7 @@ fn try_launch_chrome(chrome_path: &Path, options: &LaunchOptions) -> Result<Chro
         Err(primary_err) => {
             // Fallback: scrape stderr (legacy behavior) for better diagnostics.
             let stderr = child.stderr.take().ok_or_else(|| {
-                let _ = child.kill();
+                terminate_failed_chrome_launch(&mut child);
                 cleanup_temp_dir(&temp_user_data_dir);
                 "Failed to capture Chrome stderr".to_string()
             })?;
@@ -409,7 +421,7 @@ fn try_launch_chrome(chrome_path: &Path, options: &LaunchOptions) -> Result<Chro
             match wait_for_ws_url_until(reader, deadline) {
                 Ok(url) => url,
                 Err(fallback_err) => {
-                    let _ = child.kill();
+                    terminate_failed_chrome_launch(&mut child);
                     cleanup_temp_dir(&temp_user_data_dir);
                     return Err(format!(
                         "{}\n(also tried parsing stderr) {}",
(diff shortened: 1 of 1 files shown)


guhyun9454:
Independent reproduction + confirmation on **headless aarch64 (ARM64) Linux** — this patch fixes exactly the leak I was hitting.

### Environment
- ARM64 Linux, headless (no display), sandbox-restricted host
- Chrome cannot complete launch in this environment, so every attempt fails — which makes the failed-launch cleanup path the *common* case here, not an edge case.

### What I observed (before this patch)
Running the test suite, each failed launch went through the retry loop (`launch_chrome`, 3 attempts):

```
[chrome] Launch attempt 1/3 failed, retrying in 500ms...
[chrome] Launch attempt 2/3 failed, retrying in 500ms...
[chrome] Launch attempt 3/3 failed, retrying in 500ms...
```

Each failed attempt left an **unreaped `<defunct>` Chrome child** behind. Over a single session they piled up — I watched Chrome processes accumulate to ~37, many in zombie state:

```
$ ps -eo pid,ppid,stat,comm | grep -i chrome
  <pid>  <ppid>  Z    chrome <defunct>
  <pid>  <ppid>  Z    chrome <defunct>
  ...
$ ps -eo stat,comm | grep -c 'Z.*chrome'
37
```

### Root cause matches this PR exactly
The failures land in `try_launch_chrome`: `spawn()` succeeds, but `wait_for_devtools_active_port` never sees the port (30s deadline) → `primary_err` → stderr fallback also fails → the `wait_for_ws_url_until` error path did `let _ = child.kill();` and returned **without `wait()`** (and without `killpg` for the process group). So the main Chrome process was signalled but never reaped → `<defunct>`, and group helpers were orphaned.

`terminate_failed_chrome_launch` (`child.kill()` + `killpg(-pgid, SIGKILL)` + `child.wait()`) addresses both: it reaps the direct child and tears down the whole process group. I confirmed the two patched paths (the missing-stderr path and the `ws_url` fallback-failure path) are the only `child.kill()`-without-`wait()` spots in `try_launch_chrome`; the `spawn()`-failure path has no child to reap. So this fully covers the accumulation I saw.

LGTM from a Linux/aarch64 reproduction standpoint — happy to re-test the branch on this host if useful.

Issue #1437, opened 2026-06-10
screenshot hangs / times out at CDP Page.captureScreenshot on macOS arm64 (headless Chrome 149) — open/snapshot work fine


yhersh:
## Summary

`agent-browser screenshot` reliably hangs and never produces a file, even though `open`, `snapshot`, `wait`, and `eval` against the same page all work. The browser clearly loads and renders (snapshot returns the correct a11y tree and the page title) — only the image-capture path fails. The daemon then becomes wedged for subsequent commands.

## Environment

- agent-browser **0.27.0** (Homebrew, `/opt/homebrew/bin/agent-browser`)
- macOS **26.3.1**, Apple Silicon (**arm64**)
- Bundled browser: **Chrome for Testing 149.0.7827.22** (`~/.agent-browser/browsers/chrome-149.0.7827.22`), launched `--headless=new --enable-unsafe-swiftshader` (software GL)

## Repro (trivial page — not site-specific)

```
$ agent-browser open https://example.com
✓ Example Domain
  https://example.com/

$ agent-browser snapshot -i
- heading "Example Domain" [level=1, ref=e1]
- link "Learn more" [ref=e2]

$ agent-browser screenshot /tmp/ex.png
# (hangs, no output, no file written; never returns until killed)
```

## Observed errors

Across attempts the screenshot fails two ways:

1. On a freshly warmed session:
   ```
   ✗ CDP command timed out: Page.captureScreenshot
   ```
2. When the daemon is already wedged from a prior hung screenshot:
   ```
   ✗ Failed to read: Resource temporarily unavailable (os error 35) (after 5 retries - daemon may be busy or unresponsive)
   ```

`os error 35` is `EAGAIN` on the daemon IPC socket — i.e. the daemon is blocked (presumably awaiting the never-completing `Page.captureScreenshot`), so it can't service the next command either. A wedged screenshot poisons the whole session until `agent-browser close --all` + `pkill -9 -f agent-browser-darwin`.

## Expected

`screenshot` writes a PNG (or errors quickly with an actionable message), and a failed capture must not wedge the daemon for unrelated subsequent commands.

## Notes / suspicion

[...]

Issue #1049, opened 2026-03-27
CLI hangs and throws os error 11 (Resource temporarily unavailable) on Ubuntu 24.04 VM


Iroha-desu:
**Description**
When running the `agent-browser open [url]` command on an Ubuntu 24.04.4 virtual machine, the CLI hangs without any response. After a while, it times out and throws an os error 11, indicating that the daemon has become unresponsive.

**Environment**

OS: Ubuntu 24.04.4 LTS (Running on VMware Workstation Pro 25H2u1)
Node.js Version: v22.22.0
Agent-Browser Version: v0.22.3

**Steps to Reproduce**
1. Install the package and system dependencies:
```bash
npm install -g agent-browser  
agent-browser install --with-deps
```

2. Attempt to open a webpage:
```
agent-browser open example.com
```

**Actual Behavior**
The terminal hangs for a period of time and eventually outputs the following error:
```
✗ Failed to read: Resource temporarily unavailable (os error 11) (after 5 retries - daemon may be busy or unresponsive)
```

DocSJ1980:
Same here, following.

Iroha-desu:
I manually closed this issue earlier, but since I noticed there is still an active PR associated with it, I'm reopening it so it can be tracked and automatically closed when the PR is merged. Thanks for the work on this!

Issue #962, opened 2026-03-22
The `--auto-connect` seems connect to current Chrome session multiple times


doggy8088:
When using `agent-browser` with the `--auto-connect` flag in the Codex CLI, it seems to connect to the current Chrome session multiple times. During a 10-minute task (one prompt), it likely requests permission about 3 to 5 times. Is it possible to approve remote debugging only once?

Issue #248, opened 2026-01-24
ARM64 Architecture Support: Chromium Not Available on aarch64


adrianpuiu:
Issue Description:                                                                   
                                                                                       
  ## Description                                                                       
  When running agent-browser on ARM64 architecture (aarch64), Playwright fails to      
  launch Chromium because it's not available for this platform.                        
                                                                                       
  ## Environment                                                                       
  - **Architecture:** ARM64 (aarch64)                                                  
  - **OS:** Linux                                                                      
  - **Node.js:** [check: `node --version`]                                             
  - **Playwright Version:** [check: `npm list playwright`]                             
                                                                                       
  ## Error                                                                             
  Error: browserType.launchPersistentContext: Chromium distribution 'chrome' is not    
  found at /opt/google/chrome/chrome                                                   
  Run "npx playwright install chrome"                                                  
                                                                                       
  Error: browserType.launchPersistentContext: Chromium distribution 'chrome' is not    
  found                                                                                
  Error: Failed to install browser:                                                    
  ╔═══════════════════════════════════════════════════════════╗                        
  ║ It looks like you are running 'npx playwright install' without first               
   ║                                                                                   
  ║ installing your project's dependencies.                                            
  ║                                                                                    
                                                                                       
  ## Steps to Reproduce                                                                
  1. Clone agent-browser repo                                                          
  2. Run on ARM64 system (Apple Silicon, AWS Graviton, etc.)                           
  3. Attempt to launch browser                                                         
  4. Observe Chromium not available error                                              
                                                                                       
  ## Expected Behavior                                                                 
  Browser should launch successfully on ARM64 platforms                                
                                                                                       
  ## Actual Behavior                                                                   
  Browser launch fails - Chromium not available for aarch64 architecture               
                                                                                       
  ## Possible Solutions                                                                
  1. **Firefox Support:** Use Playwright's Firefox browser (natively supports ARM64)   
  2. **Document Limitation:** Add ARM64 as unsupported platform in README              
  3. **Pre-built Binaries:** Include or link to ARM64 Chromium binaries                
  4. **Alternative:** Document Selenium + geckodriver as fallback for ARM64 users      
                                                                                       
  ## Related                                                                           
  - Playwright ARM64 support: https://github.com/microsoft/playwright/issues (search   
  for aarch64)

PR #1471, opened 2026-06-22
Skip launch reconfigure when the daemon is already running

guhyun9454:
## Summary

When the daemon is already running, any command that does **not** repeat the original `open`'s launch options (the normal case — you pass `--args`/`--executable-path`/etc. once to `open`, then run bare `snapshot`/`get`/`click`) silently triggers a **close + relaunch** of Chrome. If the host requires specific flags to start Chrome at all (e.g. `--no-sandbox` on headless Linux, containers, or CI), the relaunch starts Chrome **without** them, the launch never completes, and the daemon wedges — so every subsequent command hangs indefinitely.

## Root cause

Every non-trivial command first sends an implicit `launch` command to the daemon. `handle_launch` compares a hash of *this invocation's* `LaunchOptions` against the running browser's stored hash (`cli/src/native/actions.rs`):

```rust
let hash_changed = !is_external && state.launch_hash != Some(new_hash);
// ...
if needs_relaunch { close_current_browser(state).await?; /* then relaunch */ }
```

The three other launch paths in `main.rs` — `--auto-connect`, `--cdp`, and `--provider` — are already guarded with `&& !daemon_result.already_running` (#962). The **local-browser** launch/configure path was not. So when a follow-up command reaches that block (it fires whenever e.g. `--executable-path` is set, including via `AGENT_BROWSER_EXECUTABLE_PATH`), it re-sends a `launch` carrying this invocation's options. With `--args` omitted, `args` is `[]`, the hash differs from the browser launched with `[--no-sandbox]`, and the daemon kills the healthy Chrome and relaunches it without `--no-sandbox`.

This also contradicts what the CLI already prints one block earlier:

```
⚠ --args ignored: daemon already running. Use 'agent-browser close' first to restart with new options.
```

The flags are reported as *ignored*, yet they were still driving a relaunch.

## Reproduction (headless Linux aarch64, external Chromium)

```bash
export AGENT_BROWSER_EXECUTABLE_PATH=/path/to/chromium   # arm64 has no Chrome for Testing build
agent-browser open https://example.com --args "--no-sandbox"   # ✓ works
agent-browser get url                                          # hangs forever (before this PR)
```

On the wedged daemon: the daemon→Chrome CDP socket goes to `CLOSE-WAIT`, Chrome's debug port stops listening, and the daemon never recovers.

<details>
<summary>Daemon log with temporary instrumentation (AGENT_BROWSER_DEBUG=1), paths redacted</summary>

```
[daemon] Debug logging started for session: default
# --- agent-browser open https://example.com --args "--no-sandbox" ---
[DIAG] action=launch    skip_launch=true  needs_launch=false
[DIAG] action=navigate  has_process_exited=false is_connection_alive=true
[DIAG] action=navigate  skip_launch=false needs_launch=false
# open succeeds; browser + CDP port healthy

# --- agent-browser get url   (no --args this time) ---
[DIAG] action=launch    skip_launch=true  needs_launch=false
[DIAG handle_launch] new_hash=9936...560 stored_hash=Some(10229...914) hash_changed=true cmd_args=[] -> needs_relaunch=true
[DIAG handle_launch] closing browser then relaunching with args=[]
[cdp] WebSocket Error: WebSocket protocol error: Connection reset without closing handshake
# Chrome was just relaunched WITHOUT --no-sandbox -> never comes up -> daemon wedged -> command hangs
```

(`[DIAG ...]` lines are temporary `eprintln!`s added only to diagnose this; they are **not** part of the patch. `stored_hash` is the hash of the running browser launched with `args=[--no-sandbox]`; `new_hash` is from the bare `get url` invocation with `cmd_args=[]`.)

</details>

## Fix

Add `&& !daemon_result.already_running` to the local-browser launch/configure condition in `main.rs`, mirroring the guard already used on the `--auto-connect`, `--cdp`, and `--provider` paths. When the daemon is already running it keeps its existing browser and the launch-time options are ignored — exactly what the `already_running` warning already promises. The actual command still runs, and `execute_command`'s connection-liveness check (`has_process_exited() || !is_connection_alive()`) still relaunches when the browser genuinely died.

## Why it's safe

- Matches the documented "options ignored when daemon already running" behavior.
- Identical guard pattern to the three sibling launch paths (#962).
- A genuinely dead/exited browser is still relaunched via the existing liveness check in `execute_command`.
- Session reuse is preserved (verified: page/URL persists across commands instead of resetting to `about:blank`).

## MCP parity

The MCP server (`cli/src/mcp.rs`) executes every tool by spawning the CLI binary itself (`run_cli` → `env::current_exe()`), so it delegates through this same `main()` path and inherits the fix with no separate change. No new flags, commands, or output were added, so no `--help`/README/SKILL/docs updates are required.

## Testing

- `cargo fmt -- --check` and `cargo clippy -- -D warnings`: clean.
- `cargo test --bin agent-browser`: **800 passed, 0 failed.** The remaining browser-launching unit/parity tests (`native::actions::tests`, `native::parity_tests`) exercise a real Chrome launch and could not complete on the headless aarch64 host I tested on — Chrome cannot start there in their default configuration, so they retry and stall ("has been running for over 60 seconds"). That is an environment limitation independent of this one-line change (it touches only the CLI launch-skip condition, not those code paths); CI runs the full matrix on a working Chrome.
- Manual end-to-end on headless aarch64 (Ubuntu 20.04) with external Chromium: before the fix, `get url`/`snapshot`/`screenshot` after `open --args "--no-sandbox"` hang; after the fix the full `open → snapshot (refs) → screenshot` loop works and the browser is reused across commands.

## Related issues

Likely the same root failure (daemon wedges after the first command):
- #1437 — screenshot hangs at `Page.captureScreenshot`, then the daemon is wedged for subsequent commands (arm64)
- #1049 — CLI hangs / `os error 11` on Ubuntu VM
- #248 — ARM64: Chromium not available (why arm64 users must use `--executable-path`, which is what surfaces this bug)

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1231,6 +1231,18 @@ fn main() {
         && flags.cdp.is_none()
         && flags.provider.is_none()
         && !flags.auto_connect
+        // Skip when the daemon was already running. It already holds a
+        // configured browser, and these launch-time options are reported as
+        // ignored above (see the `already_running` warning). Re-sending a
+        // `launch` command here carries this invocation's options (e.g. an
+        // empty `args` list when `--args` was only passed to the original
+        // `open`), which differs from the running browser's launch hash and
+        // forces a needless close + relaunch. On hosts that require specific
+        // launch flags to start Chrome at all (e.g. `--no-sandbox` on headless
+        // Linux/containers), that relaunch starts Chrome without them and
+        // wedges the daemon. Mirrors the same guard on the --auto-connect,
+        // --cdp, and --provider launch paths (#962).
+        && !daemon_result.already_running
     {
         let mut launch_cmd = json!({
             "id": gen_id(),


dabuddha:
Another affected consumer, still reproducible on **0.34.0** (macOS arm64): [terminal-browser](https://github.com/zenbu-labs/terminal-browser)'s `action` bridge, which exists to drive an already-open browser over CDP.

Its call sequence is exactly the shape described here — one invocation carrying launch options, then bare ones:

```ts
// cli/src/action.ts
runAgent(binary, ["--session", s, "--cdp", port, "tab", "list", "--json"]);  // correct target
spawnSync(binary, ["--session", s, ...passthrough]);                         // relaunches
```

Reduced to agent-browser alone, against a browser it does not own:

```bash
# external Chrome, not launched by agent-browser
"Google Chrome for Testing" --remote-debugging-port=9333 \
  --user-data-dir=/tmp/ext-profile --headless=new https://example.org

curl -s http://127.0.0.1:9333/json/version   # Chrome/146.0.7680.153, healthy

agent-browser --session a --cdp 9333 get url   # https://example.org/   ✓
agent-browser --session a --cdp 9333 get title # Example Domain         ✓

agent-browser --session b connect 9333         # "[agent-browser] relaunched browser"
agent-browser --session b get url              # not example.org        ✗
```

Same endpoint, same version. `--cdp` repeated on every invocation is reliable; any follow-up invocation that drops it loses the attachment and a local browser is started instead.

One consequence worth noting: this makes the documented behaviour in `skill-data/electron/SKILL.md:88` unreachable on 0.34.0 —

> After `connect`, all subsequent commands target the connected app without needing `--cdp`.

Tested on 0.34.0 and 0.33.2, identical results.

PR #1487, opened 2026-06-25
Fix open with arguments accidentally losing state

gaearon:
The `next-dev-loop` Skill loses cookies when using `agent-browser`.

Claude narrowed this down to this case, where the `try_auto_restore_state` call was missing because we pass launch arguments like `--headed` to `open`. This is the fix proposed by Claude and the new tests that should fail without this fix.

This is vibed (after much trial and error) so read the code closely.

## Test Plan

New test. Also, I'll be doing some manual testing in a bit.


ctate:
Code LGTM - can you plz fix format/CI failures?

PR #1509, opened 2026-07-06
Add periodic restore state autosave

ctate:
* Save restore state periodically after browser-touching commands, honoring the quiet period and restore-save policy.
* Add the AGENT_BROWSER_AUTOSAVE_INTERVAL_MS control and cross-platform daemon ticks.
* Document autosave behavior across CLI help, README, docs, and core skill references.


gaearon (approved):
tested on a real app, works

PR #1511, opened 2026-07-07
fix(cli): preserve commas inside --args values

### LINKED ISSUE #1501: --args splits on every comma with no escaping, making it impossible to pass launch args whose values contain commas (e.g. --disable-features=A,B,C)
Summary

  The --args flag parses its value by splitting on every comma and newline character, with no quoting/escaping mechanism. This makes it impossible to pass any Chromium launch argument whose value itself contains a comma — most notably --disable-features=FeatureA,FeatureB,FeatureC, which is the canonical way to
  disable multiple Chrome features at once.

  The same limitation applies to the args field in config.json (it's a string that goes through the identical split).

asif786ka:
## Summary
- Add split_launch_args that rejoins comma-separated flag values
- Add regression tests for multi-value disable-features flags

Fixes #1501

## Test plan
- [x] cargo test split_launch_args

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cli/src/flags.rs
+++ b/cli/src/flags.rs
@@ -9,6 +9,37 @@ const CONFIG_DIR: &str = ".agent-browser";
 const CONFIG_FILENAME: &str = "config.json";
 const PROJECT_CONFIG_FILENAME: &str = "agent-browser.json";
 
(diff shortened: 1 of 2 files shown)


lithium-yinl (commented):
agree

Issue #1514, opened 2026-07-08
browseruse provider broken — hardcoded WebSocket endpoint returns 400


yookibooki:
The `browseruse` provider in `cli/src/native/providers.rs` fails to connect with:

```
Provider 'browseruse' connection failed: CDP WebSocket connect failed: HTTP error: 400 Bad Request
```

**Cause:** The provider hardcodes `wss://connect.browser-use.com?apiKey={key}` as the WebSocket URL. This endpoint now returns HTTP 200 with a JSON CDP info response instead of upgrading to WebSocket, so `tokio-tungstenite` gets a 400.

Browser Use API has changed. The current flow (as of their v3 API docs) is:

1. `POST https://api.browser-use.com/api/v3/browsers` with header `X-Browser-Use-API-Key`
2. Response includes `cdpUrl` (e.g. `https://{id}.cdp.browser-use.com`)
3. `GET {cdpUrl}/json/version` returns `webSocketDebuggerUrl`
4. Connect to that WebSocket URL

Other providers (Browserbase, Browserless, Kernel) already follow this REST→WebSocket pattern correctly. Browser Use is the only one with a stale implementation.

**Suggested fix:** Update `connect_browser_use()` to create a session via the v3 REST API, then resolve the WebSocket URL from `/json/version`, similar to how `connect_browserbase()` works.

PR #1519, opened 2026-07-08
fix: upgrade Browser Use to v3 and clean up orphaned remote sessions

yookibooki:
1. Browser Use was still using the legacy v2 API. Updated it to v3.
2. Remote provider sessions (Browser Use, Browserbase, etc.) weren't being closed when running `close all`. Added cleanup so stale sessions don't linger and count against plan limits.


yookibooki:
Closes #1514
Closes #1520

PR #1529, opened 2026-07-10
feat: add WebGPU launch preset

ctate:
- Add --webgpu, config, environment, and MCP support for WebGPU launches.
- Enable Chrome WebGPU flags, including Linux SwiftShader Vulkan support.
- Document the workflow and add a doctor probe that validates WebGPU pixels.


hiSandog:
What should happen when `--webgpu` is requested on a host where Vulkan or a compatible adapter is unavailable? A launch test for that case could pin whether the command fails with a clear doctor-style diagnostic or falls back to a non-WebGPU browser, rather than starting successfully and failing later during page execution. It may also be worth covering precedence when the CLI flag, environment variable, and config file disagree.

PR #1532, opened 2026-07-11
fix(native): revive discarded tabs on tab switch instead of hanging the daemon

### LINKED ISSUE #1528: Bug: `tab <label>` onto a Memory-Saver-discarded tab hangs forever in 0.31.1 and stalls the daemon queue (connect is fixed; same root cause as #1036)
## Summary

As of **0.31.1**, `connect 9222` against a browser with Memory-Saver-discarded tabs completes fine (the main path in #1036 — thanks). But **switching to a discarded tab via `tab <label>` still hangs forever**, and while it is stuck, every subsequent command (`snapshot`, `eval`, `tab`) queues behind it in the daemon and also hangs. The daemon recovers only after the hung CLI process disconnects.

## Environment

- agent-browser 0.31.1 (npm), macOS arm64 (Darwin 24.6)
- Google Chrome 149.0.7827.201, `--remote-debugging-port=9222`, Memory Saver on (default)

## Reproduction

```bash
# 1. Chrome with several background tabs, idle long enough for Memory Saver
#    to discard some (renderer process killed; CDP target stays listed).
#    Verify a tab is discarded: raw-CDP Page.enable / Runtime.evaluate to its
#    session gets no reply, ever (no error either).

agent-browser connect 9222   # ✅ works in 0.31.1 (used to hang, see #1036)
agent-browser tab            # lists tabs, incl. the discarded one, e.g. t3
agent-browser tab t3         # ❌ hangs forever
# in another shell, while the above is stuck:
agent-browser snapshot -i    # ❌ also hangs (daemon queue stalled)
```

Kill the hung CLI → daemon recovers, later commands work again.

## Root cause

Same as #1036: a discarded tab has **no renderer process**, so renderer-bound CDP commands (`Page.enable`, `Runtime.evaluate`, `Page.bringToFront`, …) never get a response. The `connect` attach path got a fix, but the tab-switch path still awaits a renderer-bound call with no timeout.

## Verified fix mechanism (no focus steal)

`Page.reload` is answered by the **browser process** and respawns a discarded tab's renderer **without focusing/activating the tab**. Verified: after sending `Page.reload` to the dead session and waiting ~2s, the renderer responds and the same `tab <label>` completes instantly. (`Target.activateTarget` also revives, but visibly steals the active tab — worse for connect-to-user's-browser workflows.)

Suggested: wrap renderer-bound commands in a timeout, and on timeout either auto-revive via `Page.reload` (maybe behind a flag, since it reloads the page) or fail the command with a clear "tab is discarded" error instead of stalling the daemon queue.

Railly:
## What

`tab <ref>` onto a tab that Chrome Memory Saver discarded hung the daemon (#1528). A discarded tab keeps its CDP session but has no renderer, so `Page.enable` never answered.

## Fix

- probe the renderer, and recover a dead one with `Target.activateTarget`: it reloads a discarded tab but only focuses a live one, so a probe false positive keeps page state
- commit the switch only after the tab is usable, and clear the caller's refs and frame scope only after it succeeds, so a failed switch no longer strands the caller
- treat a tab paused by a JavaScript dialog as alive, not discarded: report `dialogBlocked` and skip the evals that would otherwise hang on the paused renderer
- revive a discarded successor on tab close, and keep the close best-effort so a committed close never reports failure
- clear a cancelled probe's pending CDP request with a drop guard

## Result fields

Each appears on the default CLI line, `--json`, MCP, the README, the core skill, and the docs site:

- `revived`: the switch reloaded a discarded tab
- `dialogBlocked`: the switch target is paused by a dialog, resolve it with `dialog accept` or `dialog dismiss`
- `activeTabRevived`: the close reloaded the tab that became active

## Verification

- mock-CDP tests: revive, fail-fast without poisoning the active tab, responsive tab untouched, cancelled-probe leak, dialog-blocked not misclassified, close revives the successor, close succeeds when the successor cannot revive
- real Chrome 150: a discarded-tab switch revives in about 3.4s with a follow-up command in about 10ms; activating a live target fires no navigation and keeps JS state
- `cargo test`, `clippy`, and `fmt` green

Fixes #1528.


bluk1020:
Independent validation on exact head `25c81ecd65cfc8966a4580ed0d6bced6a4b6d1ff` (macOS arm64):

- 7/7 focused discarded-tab mock-CDP tests pass
- full native suite: 965 unit + 2 doctor CLI tests pass
- `cargo clippy -- -D warnings` clean
- arm64 release build succeeds
- installed the pinned binary locally and verified `doctor --offline --quick` (8 pass, 0 fail) plus an isolated open/snapshot/close smoke

No dependency, lockfile, workflow, postinstall, or download-path changes found in the diff. This fixes the issue locally without reproducing the global command-queue stall.

ctate (approved):

PR #1545, opened 2026-07-14
Security hardening

dnukumamras:
Extends the `--allowed-domains` containment to WebRTC, which previously escaped it entirely: `RTCPeerConnection` STUN/TURN egress uses UDP/TCP outside the CDP Fetch layer, so ICE server hosts were never checked.

- Wrap `RTCPeerConnection`/`webkitRTCPeerConnection` in the init script so every `iceServers[].urls` host (`stun:`/`stuns:`/`turn:`/`turns:`) is checked against the same allowlist helper as WebSocket/EventSource/sendBeacon; also covers `setConfiguration()`.
- Add `--force-webrtc-ip-handling-policy=disable_non_proxied_udp` when domain filtering is active, gated so a user-supplied policy wins.
- Coerce every url value with `String()` (matches WebIDL), validate a snapshot and hand Chrome the copy (no getter TOCTOU), fail closed on userinfo (`@`) and malformed input.
- Preserve prototype, `name`, and static methods (e.g. `generateCertificate`) for allowed hosts.
- Update security docs.

## Behavior

Happy path
- Allowed ICE host (e.g. `stun:example.com`): connection proceeds unchanged.
- Static methods and normal WebRTC APIs on allowed hosts keep working.

Unhappy path
- Disallowed host, comma/array url forms, TURN with `?transport=tcp`: throws `DOMException('SecurityError')`.
- `String`-object / `toString()` wrappers, `@` smuggling, malformed IPv6, getter-based TOCTOU: blocked (fail closed).

## Verification

`cargo build`, `cargo clippy --all-targets` (clean), `cargo test` (925 pass). New unit test asserts all patched constructors and the hardening stay in place.

## Residual gaps

- If `eval` is allowed, page scripts could restore the original constructor (documented limitation).
- DNS-only leakage via a STUN hostname lookup survives the UDP flag if the JS layer is bypassed.


ctate:
Addressed in #1546

PR #922, opened 2026-03-18
perf: fast-path identical snapshots in diff_snapshots

light-merlin-dark:
## Summary
Fast-path identical inputs in `diff_snapshots` to avoid constructing the `similar::TextDiff` diff machinery when agents compare a snapshot to itself.

## Benchmark Results
Microbench (unit test, ignored; 50,000 iterations on this machine)

- identical (before==after): 688.17 ms total
- changed: 10533.09 ms total

Speedup: ~15.3x in the identical case.

## Changes
- `cli/src/native/diff.rs`
  - `diff_snapshots(before, after)` now returns immediately when `before == after`.
  - Adds:
    - `test_diff_snapshots_identical_fast_path`
    - `bench_diff_snapshots_identical_and_changed` (ignored microbench)

## Verification
- cargo test (unit tests): PASS
- cargo fmt -- --check: PASS
- cargo clippy: PASS
- cargo test bench_diff_snapshots_identical_and_changed -- --ignored --nocapture

---

[Merlin](https://rbeckner.com) ([@EnchantedRobot](https://x.com/EnchantedRobot) on X) & Hermes (gpt-5.4-nano)

Issue #44, opened 2026-01-13
Support snapshot command return incremental snapshot info, also with saving to file option


JasonBoy:
Looking at the `agent-browser snapshot` info, it has some filters to reduce snapshot size, but is it taking the snapshot for the whole page if no options passed, which may still has large snapshot size for some pages?
Currently for the playwright-mcp, it returns incremental snapshot info to reduce size for sequential actions; I also has an option to save taken snapshot to file instead of returning to ai directly, and then let ai to search related info to further reduce context consumption.
It will be great to have those options, thanks.

syabro:
> option to save taken snapshot to file instead of returning to ai directly

Probably would help
```
agent-browser snapshot > mySnaphot.txt
```

PR #1582, opened 2026-07-20
perf: cache unchanged snapshots behind mutation barriers

jadenfix:
> **Audit outcome:** Withdrawn. The mutation barrier is document-local and does not authoritatively cover child-frame, property-only, or computed-style changes, so it can reuse stale snapshots and refs. The original proposal below is retained only as research context, not as a shipping guarantee.

## Summary

- cache an unchanged accessibility snapshot and ref map behind an authoritative page-side mutation index
- create the observer in an isolated world for every top-level, frame, and OOPIF document so page code cannot poison the barrier
- conservatively invalidate on relevant CDP lifecycle and execution-context events, option changes, and document identity changes
- cap retained cache data at 8 MiB and preserve the existing snapshot result contract

The steady-state path reuses the cached result when the isolated mutation index is unchanged, avoiding repeated full AX, DOM, and computed-style work. Changed documents still take the canonical full snapshot path.

## Issue and overlap

Related to #44, but does not close it because this is an internal unchanged-page cache rather than the requested CLI incremental-output and file contract.

PR #473 by @giulio-leone previously proposed that broader CLI diff surface, and merged PR #922 added the existing identical-result fast path. This work builds on those ideas without superseding or claiming either contribution.

## Validation

- `cargo fmt --all -- --check`
- strict Clippy across all targets and features
- full Rust suite: 967 unit tests plus 2 doctor integration tests passed
- all 86 ignored serial browser E2Es passed
- 17 focused snapshot/cache tests passed
- 1,500-node real-browser test verified the unchanged fast path with DOM and AX calls disabled, timer mutation invalidation, navigation recreation, and cached-context reuse
- adversarial E2E poisoned `Symbol`, `MutationObserver`, and `Object.defineProperty` in the main world and still passed
- frame and option-change invalidation E2Es passed

## Full fork CI

- [Cross-platform validation run](https://github.com/tempera-dev/agent-browser/actions/runs/29722826047) passed Linux Rust, Apple Silicon, Windows test compilation and focused filters, native E2E, Windows integration, and global installs on Linux, macOS, and Windows.
- The only red job is the Intel macOS parity test timing out on action `launch`. The same test and timeout reproduce on a [fresh untouched upstream `main` validation run](https://github.com/tempera-dev/agent-browser/actions/runs/29728538114).

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -23,7 +23,7 @@ use super::cdp::types::{
 };
 use super::cookies;
 use super::diff;
-use super::element::RefMap;
+use super::element::{resolve_ax_session, RefMap};
 use super::inspect_server::InspectServer;
 use super::interaction;
 use super::network::{self, DomainFilter, EventTracker};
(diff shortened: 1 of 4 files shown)


jadenfix:
Closing after a deeper safety audit.

The isolated `MutationObserver` is document-local, while a top-level snapshot recursively includes child-frame documents. Snapshot output also depends on property and computed-style state that `MutationObserver` cannot authoritatively version (for example property-only checked/onclick changes and CSS-driven cursor/layout changes). Those false negatives can return a stale ref map while reporting the cache as current, so this optimization is not safe enough to ship.

The branch remains available as research. A successor should maintain per-document AX/DOM state or use a conservative cache contract that can prove unchanged snapshot inputs before reuse.

PR #1588, opened 2026-07-20
fix(screenshot): error clearly when an explicit path's parent dir is missing

### LINKED ISSUE #1534: screenshot silently falls back to tmp dir + prints success when target directory doesn't exist
## Summary

When the `screenshot` command is given a path whose parent directory does not exist (as resolved against the daemon's cwd), agent-browser **silently saves to `~/.agent-browser/tmp/screenshots/screenshot-<epoch>.png`** and prints `✓ Screenshot saved to …` with exit code 0. No warning, no error. The only signal that anything went wrong is that the returned path does not match the path the caller requested.

For an agent-driven tool, a false success on evidence capture is the worst-case failure mode: the agent records the check as PASS and moves on, and the screenshot is quietly lost to a temp dir the agent never inspects.

## Environment

- agent-browser 0.31.1
- macOS arm64 (Darwin 25.5.0), system Chrome 150

## Reproduction

```bash
agent-browser screenshot ./does-not-exist/page.png
# ✓ Screenshot saved to /Users/<user>/.agent-browser/tmp/screenshots/screenshot-<epoch>.png
# exit 0
#
# ./does-not-exist/ was never created. The file is NOT at ./does-not-exist/page.png.
```

This compounds with the daemon-cwd decoupling (#137): a relative path is resolved against the daemon's pinned cwd, not the caller's shell cwd, so even a directory that exists in the caller's cwd can trigger the fallback if it doesn't exist in the daemon's cwd.

## Expected behavior

If the parent directory of the requested screenshot path does not exist (after path resolution), **error** with a clear message, e.g. `✗ Screenshot path parent directory does not exist: <resolved dir>`. Do not fall back to the tmp dir when an explicit path was supplied. The tmp-dir fallback introduced in #247 is reasonable for the *no-path-provided* case; it should not fire for the *path-provided-but-unwritable* case.

## Suggested fix

In `cli/src/native/screenshot.rs` (and any shared write path), before delegating to the tmp-dir fallback: if an explicit output path was provided, verify the parent directory exists (after resolving relative paths against the CLI cwd per #137). If it doesn't, return an error rather than falling back.

## Related

- #137 — relative paths resolve to daemon cwd (the resolution half of this story; fixing it alone does not fix the silent fallback)
- #247 — introduced the tmp-dir fallback for the no-path case (the fallback this issue argues should be scoped to that case only)

amitvijapur:
Fixes #1534

### Context

On current `main`, giving `screenshot` an explicit output path whose parent directory doesn't exist already exits non-zero: `save_screenshot`'s explicit-path arm writes directly and propagates the OS error, and the tmp-dir fallback only lives in the no-path arm. So the catastrophic silent-success in the report isn't reproducible at HEAD (consistent with @vishnujayvel's stale-daemon note). What's missing is a clear message and an explicit, tested guarantee that an explicit path never falls back to tmp.

### Change

In the explicit-path arm of `save_screenshot`, verify the parent directory exists before writing. If it doesn't, return `Screenshot path parent directory does not exist: <dir>` instead of the generic OS failure. The no-path arm (which creates and uses the tmp screenshot dir) is unchanged, so default behavior is preserved.

### Before / after (unit-level, explicit path with a missing parent)

- Before: `Failed to save screenshot to .../does-not-exist/shot.png: No such file or directory (os error 2)`
- After: `Screenshot path parent directory does not exist: .../does-not-exist`

### Behavior preserved

- Bare filenames (parent `""`) and the no-path/`--screenshot-dir` tmp flow are unaffected.
- The change is in the shared native `save_screenshot`, so CLI and MCP stay in parity.

### Tests

Added two unit tests in `cli/src/native/screenshot.rs`: one asserting the clear error (and that the tmp dir isn't used) when the parent is missing, one asserting an explicit path with an existing parent writes exactly there and returns that path. `cargo test` (960 passed, 0 failed), `cargo fmt --check`, and `cargo clippy` all pass.

### DIFF
--- a/cli/src/native/screenshot.rs
+++ b/cli/src/native/screenshot.rs
@@ -558,7 +558,23 @@ fn save_screenshot(
     output_dir: Option<&str>,
 ) -> Result<String, String> {
     let save_path = match explicit_path {
-        Some(path) => path.to_string(),
+        Some(path) => {
+            // An explicit output path was supplied. Its parent directory must
+            // already exist: we do not create it, and we must never silently
+            // redirect to the tmp screenshot dir. Falling back there would
+            // report a false success for an agent that relies on the returned
+            // path as evidence (see issue #1534). The tmp fallback below is
+            // reserved for the no-path case.
+            if let Some(parent) = std::path::Path::new(path).parent() {
+                if !parent.as_os_str().is_empty() && !parent.exists() {
+                    return Err(format!(
+                        "Screenshot path parent directory does not exist: {}",
+                        parent.display()
+                    ));
+                }
+            }
+            path.to_string()
+        }
         None => {
             let dir = match output_dir {
                 Some(d) => PathBuf::from(d),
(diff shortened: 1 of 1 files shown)


amitvijapur:
Closing this to keep the queue tidy, not because the bug went away.

The change is small and still applies: `screenshot` currently fails with an unclear error when an explicit path's parent directory does not exist, and this makes it say so directly. The two Vercel deploy checks on the PR report `Authorization required to deploy`, which needs someone on the team to grant, so the checks cannot go green from my side.

Happy to reopen or send it again if it would be useful. Thanks for the project.

Issue #1272, opened 2026-04-19
CDP connection fails silently when external browser restarts (new webSocketDebuggerUrl)


VionSSS:
## Bug Description

When connecting to an external Chrome browser via CDP (e.g., `agent-browser connect ws://localhost:9222`), if the browser is closed and restarted, the CDP connection fails silently. This is because Chrome generates a new `webSocketDebuggerUrl` UUID on each restart, but the agent-browser daemon does not rediscover the new URL.

## Steps to Reproduce

1. Start Chrome with remote debugging:
   ```
   chrome.exe --remote-debugging-port=9222 --remote-debugging-address=0.0.0.0
   ```
2. Connect via CDP:
   ```
   agent-browser connect ws://localhost:9222
   ```
   Connection succeeds. Chrome UUID is e.g. `aaa-111`.

3. Close Chrome and restart it with the same flags.
   New Chrome UUID is e.g. `bbb-222`.

4. Try to connect again:
   ```
   agent-browser connect ws://localhost:9222
   ```
   **Result:** Connection fails with:
   > Failed to connect via CDP to ws://localhost:9222. Make sure the app is running with --remote-debugging-port=ws://localhost:9222

## Root Cause

In `cli/src/main.rs` around line 985:
```rust
// Skip when daemon already running — it already holds the CDP connection.
if !daemon_result.already_running {
    send_command(launch_cmd, &flags.session);
}
```

When the daemon is already running (from a previous connection), the CDP launch command is **silently skipped**. The daemon continues trying to use the old `webSocketDebuggerUrl` UUID, which no longer exists after Chrome restarts.

## Expected Behavior

The `connect` command (or `--cdp` flag) should always attempt to (re)connect, even when the daemon is already running. The daemon should close the stale connection and establish a new one using the fresh `/json/version` discovery endpoint.

## Environment

- agent-browser version: 0.17.1
- OS: WSL2 on Windows 11
- Chrome: 147.0.7727.102
- Use case: Connecting to an external real Chrome browser for session/cookie preservation

VionSSS:
## Bug Description

When connecting to an external Chrome browser via CDP (e.g., `agent-browser connect ws://localhost:9222`), if the browser is closed and restarted, the CDP connection fails silently. This is because Chrome generates a new `webSocketDebuggerUrl` UUID on each restart, but the agent-browser daemon does not rediscover the new URL.

## Steps to Reproduce

1. Start Chrome with remote debugging:
   ```
[...]

Issue #1211, opened 2026-04-10
`--profile` sessions lose the active page after the first command and fall back to `about:blank`


liustack:
## Summary

On macOS, `agent-browser 0.25.3` works correctly without `--profile`, but when I use a persistent profile directory, the first `open` succeeds and the second command in the same session (`tab list` / `get url`) loses the active page and reports `about:blank`.

## Environment

- `agent-browser 0.25.3`
- macOS 15.3 (`24D2059`)
- Chrome profile version: `147.0.7727.56`

## Minimal repro

This works correctly without `--profile`:

```bash
npx agent-browser --headed close --all
npx agent-browser --headed --session repro-noprofile open https://example.com
npx agent-browser --headed --session repro-noprofile tab list
npx agent-browser --headed --session repro-noprofile get url
```

Output:

```text
✓ Example Domain
  https://example.com/
→ [0] example.com - https://example.com/
https://example.com/
```

This fails with `--profile`:

```bash
npx agent-browser --headed close --all
AGENT_BROWSER_DEBUG=1 npx agent-browser --headed --profile ~/.agent-browser/profile --session repro-profile open https://example.com
AGENT_BROWSER_DEBUG=1 npx agent-browser --headed --session repro-profile tab list
AGENT_BROWSER_DEBUG=1 npx agent-browser --headed --session repro-profile get url
```

Output:

```text
✓ Example Domain
  https://example.com/
→ [0]  - about:blank
about:blank
```

## Expected

The second and third commands should keep using the page opened by the first command:

- `tab list` should show `https://example.com/`
- `get url` should return `https://example.com/`

## Actual

After a successful `open`, the next command in the same session falls back to `about:blank`.

## Extra signal

The daemon log shows:

```text
[cdp] WebSocket Error: WebSocket protocol error: Connection reset without closing handshake
```

Also, if I inspect Chrome immediately after `open`, the browser really does contain the correct page target (`https://example.com/`). The problem appears when the next `agent-browser` command reuses the session.

mpz4life:
I can reproduce this on Linux with agent-browser 0.31.1.

Environment:
- agent-browser 0.31.1
- Linux (NewStartOS V4.4.2-ZTE)
- Chrome 138.0 / Chrome for Testing 148.0

Repro:
```bash
agent-browser --profile Default open https://example.com
agent-browser get url
# returns: about:blank
```

Also verified:
- `--profile Default --headed` → still about:blank
- without `--profile` → works correctly
[...]

PR #883, opened 2026-03-17
feat(cli): add --tab flag for tab-scoped commands

mvanhorn:
## Summary

Add a `--tab N` global flag that runs a command against a specific tab without switching the active tab. Useful for multi-tab scraping workflows where you need to interact with background tabs without losing your current context.

```bash
# Take screenshot of tab 2 without switching to it
agent-browser --tab 2 screenshot

# Get the URL of tab 1
agent-browser --tab 1 url

# Click an element in tab 0
agent-browser --tab 0 click @e1
```

Related: #853

## Evidence

| Signal | Source |
|--------|--------|
| Issue #853 | "SPA Scraping Pattern: Preserving accessibility refs with tab isolation" |
| Issue #74 | Tab management friction (tab list didn't show all tabs) |
| Gap | Must currently switch tabs to run commands, losing active tab context |
| [Reddit r/ClaudeAI](https://www.reddit.com/r/ClaudeAI/comments/1rr0pyr/) | "Whats your reliable browser setup for Claude Code?" - parallel tab management is a pain point | 3 upvotes |
| [Reddit r/AI_Agents](https://www.reddit.com/r/AI_Agents/comments/1rtlzmy/) | "Built a fully autonomous system to coordinate 100+ browser automation agents" | 8 upvotes |
| [NxCode](https://www.nxcode.io/resources/news/stagehand-vs-browser-use-vs-playwright-ai-browser-automation-2026) | "Parallel browser session management" identified as universal unsolved gap | Industry report |

## Implementation

The `--tab` flag is parsed in `flags.rs` and injected as `tabIndex` into the command JSON in `main.rs`. The daemon's `handle_action` in `actions.rs` temporarily overrides the `active_page_index` on `BrowserManager` for the duration of the command, then restores the original index afterward. This means all existing handlers work with `--tab` without modification.

### Files changed
- `cli/src/flags.rs` - parse `--tab N` flag, add to clean_args skip list
- `cli/src/main.rs` - inject `tabIndex` into command JSON
- `cli/src/native/actions.rs` - temporarily swap active page index before/after dispatch
- `cli/src/native/browser.rs` - add `active_page()` and `set_active_page()` methods
- `cli/src/commands.rs` - add `tab_index` to test default_flags
- `cli/src/output.rs` - add `--tab` to help text

## Test plan
- [x] `cargo fmt` passes
- [x] `cargo clippy` passes
- [x] `cargo test` passes (473 tests, 0 failures)
[...]

PR #1589, opened 2026-07-21
feat(session): persistent session tab binding with --pin-tab

### LINKED ISSUE #214: Skill does not use agent-browser session by default, causing issues when parallel agents are testing

### LINKED ISSUE #1265: Feature request: agent-browser tab list --json – include CDP targetId and support tab operations by targetId
Image

Could we enhance agent-browser tab list --json to return the CDP targetId for each tab? 
Additionally, it would be great if tab operations (like switching, closing, or focusing on a tab) could be performed using the targetId instead of just the index or URL.

This would improve automation reliability, especially in scenarios where tabs have dynamic URLs or the tab order changes.

### LINKED ISSUE #1530: Sessions sharing one Chrome over --cdp hijack each other's tabs (no persisted session-to-tab binding)
## Summary

When N agents share one Chrome over `--cdp <port>` with unique `--session` names, sessions hijack each other's tabs (and the user's tabs). The session's "current tab" is an in-memory index inside the daemon; nothing maps a session to a CDP target across daemon restarts or tab removals, so the pointer silently re-resolves to some other tab.

## Reproduction

Verified on a live system (v0.28.0; the relevant code is unchanged in 0.31.1). Start one Chrome with `--remote-debugging-port=9222`, then create 10 named sessions and open a distinct URL in each:

```bash
for i in $(seq 1 10); do
  agent-browser --session "agent$i" --cdp 9222 open "https://example.com/?page=$i"
done
for i in $(seq 1 10); do
  agent-browser --session "agent$i" eval "location.href"
done
```

Result: `eval location.href` returned the SAME url for all 10 sessions, and `tab list` from different sessions showed the same current-tab marker. Every session ends up driving whatever tab happens to be active.

## Root cause

`--session` isolates the daemon process, not the tab. Two code paths re-resolve the active tab silently:

1. **Daemon (re)attach.** `discover_and_attach_targets` in `cli/src/native/browser.rs` attaches to all page targets and sets `active_page_index = 0`. `Target.getTargets` returns most-recently-active first, so a fresh daemon (idle-timeout shutdown, crash, or the first command of a new session) adopts whatever tab is currently active, typically another agent's tab.
2. **Tab removal fallback.** `active_page_index_after_removal` and the clamp logic in `browser.rs` silently shift the pointer to a neighboring tab when the current tab is closed by anyone else.

The consequences compound: `open <url>` navigates the wrong tab in place, and `tab close` with no ref closes the wrong tab. There is no persisted session to target mapping anywhere; daemon session state is only `{session}.{pid,sock,stream,version}` in the socket dir. Meanwhile `PageInfo` already stores `target_id` per tab, so the raw data exists. Only the binding, its persistence, and the failure semantics are missing.

## Related issues

These look like the same defect surfacing in different code paths:

- #326 (multi-agent tab hijacking)
- #896 (parallel `--session` collisions)
- #1219 (tab close retargets the active page; closed)
- #384 (current tab is a CLI-side pointer; closed)
- #1211 (active page lost, falls back to about:blank)
- #1272 (silent state loss on external Chrome restart)
- #1265 (feature request: targetId in `tab list`; a fix for this issue naturally implements it)
- #1068 (BrowserContext isolation; complementary but out of scope here)

## Proposed fix

1. Persist a session to target binding (`{session}.target` next to the existing per-session files), written when a session creates a tab or explicitly switches, and re-selected by targetId on attach instead of index 0.
2. Fail loudly instead of falling back: when the bound tab is gone, return a structured `tab_gone` error (non-zero exit, machine-readable in `--json`) while keeping `tab list` / `tab new` / `tab switch` working for recovery.
3. Gate the strict semantics behind an opt-in `--pin-tab` flag (sticky per session) so existing scripts keep the legacy fallback; the re-attach-by-binding part is a pure bug fix and applies without the flag.
4. Expose `targetId` in the CLI surface (`tab list --json`, accept target ids as tab refs), which subsumes #1265.

I have an implementation of this ready and will open a PR referencing this issue.

Railly:
## Summary
Sessions sharing one Chrome over `--cdp` no longer hijack each other's tabs, the most reported failure in the session-isolation cluster (#1530, #326, #896, #86). `--session` isolated the daemon process but not the tab: re-attach adopted the most-recently-active tab, and event-discovered tabs stole the active slot.

## What changed
- Event-discovered targets (`Target.targetCreated`) register for `tab list` but never steal the active tab.
- Persistent session-to-tab binding: the daemon persists the bound target and re-attach selects it instead of index 0.
- `--pin-tab` strict mode: a destroyed bound tab fails loud with a `tab_gone` code instead of adopting a neighbor. JSON output includes `data.targetId` and optional sanitized `data.lastUrl`, so wrappers can recover without parsing prose. Batch exposes the same recovery object as `result`. Recover via `tab new <url>`.
- CDP target ids usable as tab refs (`tab <targetId>`), covering #883.
- Skill docs require a named session up front, so parallel agents stop colliding on the default (#214).

## Re-testing
Add `--pin-tab` to the first command for every shared-browser session when re-running the repro from #1530. Without the flag, `open` intentionally preserves the legacy behavior and navigates the shared active tab, so the original script still collides. The same rule applies to sessions attached with `--auto-connect` instead of `--cdp`.

## Refs
- Closes #1530, closes #214, closes #1265.
- Supersedes, closed with credit on merge (credited via Co-authored-by):
  #1426 (@soichisumi, no-steal fix), #1531 (@dandaka, binding + strict mode,
  plus the #1530 diagnosis), #883 (@mvanhorn, covered by tab <targetId>).
- Progresses #326, #896, #86, #1211, #1272; the shared-profile singleton is #899's territory.

## Testing
The full suite passes 1,077 tests with one pre-existing screenshot parity timeout excluded; the timeout reproduces on the exact base commit. Real-Chrome coverage includes daemon restart re-binding, safe and opaque `tab_gone` recovery metadata, CLI and batch output, and the existing-daemon `--cdp --pin-tab` regression. `cargo fmt --check` and `clippy` are clean.


Railly:
Scope note: this PR is only the tab-focus layer. Follow-ups, in order:

1. Storage isolation per context: fold the community implementations into
   a maintainer PR with credit, plus a regression test for the ordering
   bug caught in review. Refs: #1068, #1340, #1117.
2. Shared-profile singleton collision: #899 covers it.
3. Per-connection daemon state redesign, once the above land: #326.

Then isolated tabs with shared auth (#1352) is a small follow-up.

nathanvale:
Built this branch (fb37749) and ran the repro from #1530 against a throwaway headless Chrome 151 over `--cdp` on macOS, with a fresh socket dir per binary. For a baseline, the same script against 0.31.2 from npm: all 10 sessions read back the same `location.href`, bug reproduces exactly as filed.

On this branch: with `--pin-tab`, 10/10 sessions keep their own tab. Killed all 10 daemons and read back again — every session re-attached to its own tab, even after I made a different tab active first via raw CDP. The unflagged path holds up too: a session that created its tab with `tab new` came back to it after a daemon kill, and external close still falls back silently like before, so the gating behaves as described.

One thing worth a line in the description or docs: the literal repro script from #1530 still collides on this branch, because `open` navigates the shared active tab and never writes a binding. That's consistent with the design, but someone re-testing the issue verbatim will conclude the fix doesn't work. "Re-run the repro with `--pin-tab`" would save that confusion.

Small ask from the machine-consumer side. Closing the pinned tab externally gives:

```json
{"success":false,"data":null,"error":"tab_gone: bound tab is gone (target 95B191E43A0A2CFE23F8CF933CB17604, last url http://localhost:8907/?page=pin2). Run `agent-browser tab new <url>` to bind a new tab, or `agent-browser tab list` to pick an existing one","code":"tab_gone"}
```

`code` is great, but the dead targetId and last url only exist inside the message string. We drive agent-browser from a wrapper that keys everything to CDP target ids, and for automatic recovery (match the dead target against our own binding record, reopen with `tab new <lastUrl>`) we'd end up regexing prose. Could those be structured fields next to `code`?

Happy to re-test any follow-up, this is the main thing keeping us from running parallel sessions against one Chrome.

ShacktarMalacky (approved):

ctate (approved):

PR #1614, opened 2026-07-27
fix: stream input coordinates use true frame dimensions

### LINKED ISSUE #1613: Stream mouse input lands off target: frame metadata carries the configured viewport instead of real CDP device dimensions
*generated with Kimi Code and verified by a human*

## Summary

In streaming mode, mouse clicks sent over the session WebSocket (from the dashboard viewport or any stream client) land off target whenever the browser's real content area differs from the configured viewport:

1. The daemon's stream server substitutes its **configured** viewport (default 1280x720, or the last `set viewport` value) into the screencast frame metadata, overwriting the real CDP `Page.screencastFrame` values for `deviceWidth`/`deviceHeight` (`cli/src/native/stream/cdp_loop.rs`).
2. The dashboard scales pointer coordinates by the `viewportWidth`/`viewportHeight` from the `status` message (`packages/dashboard/src/components/viewport.tsx` `toViewport`), not by the frame's true DIP dimensions.

The real content area routinely differs from the configured viewport: window chrome, scrollbar appearance, device emulation, or a manual window resize all change it. A browser launched at `--window-size=1280,720` ends up with a 1280x633 content area, so the dashboard stretches every Y coordinate by 720/633 ≈ 1.14: clicking the point visually at (320, 158) on the streamed page dispatches the CDP event at (320, 180). The error grows linearly with distance from the top, and X is affected whenever the widths differ.

`Input.dispatchMouseEvent` expects coordinates in the page's DIP (CSS pixel) space, which is exactly what the real CDP frame metadata reports. The protocol just doesn't forward it.

## Scope (verified against 0.33.0)

| Path | Affected | Why |
| --- | --- | --- |
| Dashboard mouse click / move / wheel | Yes | Scaled by status viewport instead of true frame dims |
| Third-party WS clients mapping by `status.viewportWidth/Height` or frame `metadata.deviceWidth/Height` | Yes | Both carry the phantom configured viewport |
| Device presets with deviceScaleFactor > 1 (e.g. iPhone 15) | Yes | Same root cause; also intersects #632 |
| CLI `click`/`mouse` commands | No | Dispatched in real CSS coordinates, no stream scaling |
| Keyboard input over the stream | No | Not coordinate-based (but see #1453, #1380) |

## Environment

- agent-browser **0.33.0**
- Linux, Google Chrome, default `chrome` engine, default headless launch

## Reproduction

Standalone, no dashboard required:

```js
// click-offset-repro.mjs — node click-offset-repro.mjs (agent-browser on PATH)
import { execFileSync } from 'node:child_process';
const SESSION = 'click-offset-repro';
const ab = (...a) => execFileSync('agent-browser', ['--session', SESSION, ...a], { encoding: 'utf8' });
const evj = (js) => JSON.parse(ab('eval', js, '--json')).data.result;

const page = 'data:text/html,' + encodeURIComponent(
  '<script>document.addEventListener("click",e=>{document.title=`CLICK ${e.clientX},${e.clientY}`})</script>');
ab('open', page);
const port = JSON.parse(ab('stream', 'status', '--json')).data.port;
const truth = evj('JSON.stringify({w:innerWidth,h:innerHeight})');

const ws = new WebSocket(`ws://127.0.0.1:${port}`);
await new Promise((res, rej) => { ws.onopen = res; ws.onerror = () => rej(new Error('ws failed')); });
const sleep = (ms) => new Promise(r => setTimeout(r, ms));

let status, meta;
ws.onmessage = (ev) => {
  const m = JSON.parse(ev.data);
  if (m.type === 'status') status ??= m;
  if (m.type === 'frame') meta ??= m.metadata;
};
await sleep(2000);

// Mimic the dashboard: a click at 25%/25% of the displayed frame maps to
// 25% of the advertised viewport.
const x = Math.round(0.25 * status.viewportWidth);
const y = Math.round(0.25 * status.viewportHeight);
for (const [eventType, button, clickCount] of [['mouseMoved','none',0],['mousePressed','left',1],['mouseReleased','left',0]]) {
  ws.send(JSON.stringify({ type: 'input_mouse', eventType, x, y, button, clickCount, modifiers: 0 }));
}
await sleep(1000);

const t = JSON.parse(truth);
console.log('real content area :', t.w + 'x' + t.h);
console.log('status viewport   :', status.viewportWidth + 'x' + status.viewportHeight);
console.log('frame metadata    :', meta.deviceWidth + 'x' + meta.deviceHeight);
console.log('intended target   :', Math.round(0.25 * t.w) + ',' + Math.round(0.25 * t.h));
console.log('page received     :', ab('get', 'title').trim());
ws.close();
```

**Observed (0.33.0, verbatim output of the script above):**

```
real content area : 1280x633
status viewport   : 1280x720
frame metadata    : 1280x720        <- fabricated: CDP reports 1280x633
intended target   : 320,158
page received     : CLICK 320,180   <- 22px too low (scaled by 720/633)
```

**Expected:** frame metadata matches the real content area (1280x633), and a click at 25%/25% of the displayed frame lands at (320,158).

## Suggested fix

Forward the real CDP `deviceWidth`/`deviceHeight` in the stream server's frame messages (fall back to the configured viewport only when CDP omits them), and have the dashboard map pointer input into the per-frame DIP space, keeping the status viewport as what it is: the *configured* size.

I have this working on a fork. Verified with the same repro: the daemon then reports `frame metadata: 1280x633`, and a client mapping by frame metadata (as the patched dashboard does) receives `CLICK 320,158`, exactly on target. 

Related: #632 (configurable screencast resolution is only safe for input mapping once the metadata is truthful).

kevingatera:
## What

In streaming mode, mouse clicks in the dashboard (or any WebSocket stream client) land off target whenever the browser's real content area differs from the configured viewport (#1613). A 1280x720 window produces a 1280x633 content area, so the dashboard stretched every Y coordinate by 720/633 and clicks landed proportionally too low.

## Fix

- `cdp_loop.rs` forwards the real CDP `Page.screencastFrame` metadata `deviceWidth`/`deviceHeight` instead of substituting the configured viewport, falling back only when CDP omits the values
- the dashboard tracks the per-frame device dimensions and maps pointer input into that DIP space, which is exactly what `Input.dispatchMouseEvent` expects; it falls back to the configured viewport before the first frame arrives
- the status viewport stays what it is: the configured size, still used for the size badge and Fit/preset actions
- README and the streaming docs now state that input coordinates map to frame metadata, not the status viewport
- the e2e viewport test previously asserted the fabricated values; it now skips not-yet-settled frames and compares against real (float) CDP values

## Verification

- live test with agent-browser driving its own dashboard: clicking the point visually at (320, 158) on the streamed page delivered (319, 181) before, (319, 159) after
- protocol-level repro from the issue: `frame metadata` reports 1280x633 after the fix, and a client mapping by metadata receives `CLICK 320,158`, exactly on target
- `cargo test` passes, `e2e_stream_frame_metadata_respects_custom_viewport` passes, `cargo fmt --check` clean

Fixes #1613.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1452,6 +1452,8 @@ Connect to `ws://localhost:9223` to receive frames and send input:
 
 **Send mouse events:**
 
+Coordinates are CSS pixels in the page's device-independent pixel (DIP) space. Map positions from the displayed frame using the `metadata.deviceWidth` and `metadata.deviceHeight` of the latest frame, not the status message's `viewportWidth`/`viewportHeight` (the configured viewport can differ from the actual content area).
+
 ```json
 {
   "type": "input_mouse",
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -5487,18 +5487,11 @@ async fn e2e_stream_frame_metadata_respects_custom_viewport() {
             serde_json::from_str(message.to_text().expect("text message should be readable"))
                 .expect("stream payload should be valid JSON");
         if parsed.get("type") == Some(&json!("frame")) {
-            let meta = &parsed["metadata"];
-            assert_eq!(
-                meta["deviceWidth"], 800,
-                "frame metadata deviceWidth should match custom viewport, got: {}",
-                meta
-            );
-            assert_eq!(
-                meta["deviceHeight"], 600,
-                "frame metadata deviceHeight should match custom viewport, got: {}",
-                meta
-            );
-
+            // Early frames may arrive before Chrome fully applies the viewport
+            // resize, with stale image and metadata dimensions. Skip those and
+            // only assert once the settled frame arrives, since the metadata
+            // now reflects the real CDP values rather than the configured
+            // viewport.
             let data_str = parsed
                 .get("data")
                 .and_then(|v| v.as_str())
@@ -5512,6 +5505,21 @@ async fn e2e_stream_frame_metadata_respects_custom_viewport() {
             if img_w != 800 || img_h != 600 {
                 continue;
             }
+
+            let meta = &parsed["metadata"];
+            // Real CDP metadata arrives as floats (e.g. 800.0).
+            assert_eq!(
+                meta["deviceWidth"].as_f64(),
+                Some(800.0),
+                "frame metadata deviceWidth should match custom viewport, got: {}",
+                meta
+            );
+            assert_eq!(
+                meta["deviceHeight"].as_f64(),
+                Some(600.0),
+                "frame metadata deviceHeight should match custom viewport, got: {}",
+                meta
+            );
 
             found_frame = true;
             break;
--- a/cli/src/native/stream/cdp_loop.rs
+++ b/cli/src/native/stream/cdp_loop.rs
@@ -150,14 +150,19 @@ pub(super) async fn cdp_event_loop(
 
                                         if let Some(data) = evt.params.get("data").and_then(|v| v.as_str()) {
                                             let meta = evt.params.get("metadata");
+                                            // Forward the real CDP frame metadata. Clients map input
+                                            // coordinates into this DIP space, so substituting the
+                                            // configured viewport here would misplace clicks whenever
+                                            // the actual content area differs from it (window chrome,
+                                            // scrollbars, device emulation, manual resizes).
                                             let msg = json!({
                                                 "type": "frame",
                                                 "data": data,
                                                 "metadata": {
                                                     "offsetTop": meta.and_then(|m| m.get("offsetTop")).and_then(|v| v.as_f64()).unwrap_or(0.0),
                                                     "pageScaleFactor": meta.and_then(|m| m.get("pageScaleFactor")).and_then(|v| v.as_f64()).unwrap_or(1.0),
-                                                    "deviceWidth": vw,
-                                                    "deviceHeight": vh,
+                                                    "deviceWidth": meta.and_then(|m| m.get("deviceWidth")).and_then(|v| v.as_f64()).unwrap_or(vw as f64),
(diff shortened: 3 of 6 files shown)


richard-fairthorne:
Validation report: running this PR's patch on top of the **v0.37.1** tag (applies cleanly; macOS arm64, built from source) as my daily driver, and it resolves the misalignment end to end.

**Before (stock 0.37.1):** frame metadata fabricated `deviceWidth/Height` from the configured viewport (1280x720) while the real content area was 1280x577, and the dashboard scaled clicks by 720/577. Measured by clicking known canvas positions and reading `e.clientX/Y` in the streamed page:

| aim | dispatched (stock) |
| --- | --- |
| (640, 150) | (640, 186) |
| (640, 500) | (640, 624) |
| (200, 60) | (200, 24) |

**After (this PR on v0.37.1, same default un-Fit-ted session):** frame metadata reports the real 1280x577, and:

| aim | dispatched (with PR) |
| --- | --- |
| (640, 150) | (640, 149) |
| (640, 500) | (640, 500) |
| (300, 700) | (299, 699) |

For anyone hitting this on stock builds, this one-liner in the dashboard host browser's DevTools console shows the mismatch on any affected session (badge = status viewport, bitmap = what the canvas actually shows):

```js
(() => { const c = document.querySelector('canvas'); const badge = [...document.querySelectorAll('span')].map(e=>e.textContent).find(t=>/^\d+ x \d+$/.test(t)); const [bw,bh] = badge.split(' x ').map(Number); console.log('badge:', badge, '| bitmap:', c.width+'x'+c.height, '| Y factor:', (bh/c.height).toFixed(3), '| ALIGNED:', bw===c.width && bh===c.height); })()
```

Output on a default session before this PR: `badge: 1280 x 720 | bitmap: 1280x577 | Y factor: 1.248 | ALIGNED: false`. With this PR the badge still shows the configured viewport (by design), but clicks map by the per-frame device dims and land exactly.

Happy to re-run these checks against any revision of the PR.

PR #1618, opened 2026-07-28
fix: dashboard stream input latency, keyboard shortcuts, and clipboard bridge

### LINKED ISSUE #1615: Dashboard live view: clicks take seconds to register, and copy/paste keyboard shortcuts don't work
## What happens

Two input problems in the dashboard's live view:

1. Clicks often take several seconds to do anything, especially right after moving the mouse across the page. Typing into a field is fast, but clicking a button or link right after moving the mouse feels stuck. The faster and longer you move the mouse before clicking, the longer the wait.
2. Keyboard shortcuts that work in any normal browser don't reach the page. Ctrl+C and Ctrl+X copy nothing, Ctrl+V pastes nothing, and other Ctrl+letter shortcuts (select all, undo, and so on) do nothing either.

## Steps to reproduce

1. Start the dashboard and open a session.
2. Move the mouse back and forth over the page for a few seconds, then immediately click a link or button. The click only takes effect after a noticeable delay.
3. Focus a text field in the page, select some text, press Ctrl+C, click into another field, press Ctrl+V. Nothing is pasted.

## Expected

Clicks register right away, and the usual editing shortcuts work like they do in a regular browser.

Tested on Linux with the Chrome engine, current main (0.33.0).

kevingatera:
## Bug

Reported in #1615. In the dashboard's live view, clicks could take seconds to register after moving the mouse, and standard editing shortcuts (copy, paste, select all, ...) never reached the page. Three compounding causes:

- Every input event waited for Chrome's reply before the next message was read, and the reply carries no data. A mouse sweep's `mouseMoved` flood (60-125 events/s) serialized a following click behind seconds of backlog.
- Letter virtual-key codes were derived with `charCodeAt` on the lowercase key, which yields numpad VKs instead of the uppercase ASCII codes Chrome matches editing shortcuts against, so every Ctrl+letter shortcut was a no-op. Combos also went out with `text`, which Chrome treats as typing rather than a command, and the native side forwarded absent `key`/`code`/`text` fields as explicit null, which makes Chrome reject the whole `dispatchKeyEvent`.
- There was no clipboard path between the local machine and the remote page at all.

## Fix

- Dispatch stream input without awaiting CDP responses (socket writes stay ordered, so event order is preserved) and coalesce forwarded mouse moves to ~25/s; only the latest position matters.
- Derive VK codes from `KeyboardEvent.code` (uppercase ASCII for letters, US OEM table for punctuation, same as Puppeteer); send shortcut combos as `rawKeyDown` without text; omit absent optional fields on the native side.
- Bridge the clipboard: Ctrl+C/X copies remotely and mirrors the remote clipboard back into the local one; Ctrl+V reads the local clipboard and inserts its text at the focus, falling back to the raw paste keystroke where the local clipboard API is unavailable (insecure LAN origins, Firefox without `readText`). `clipboard-read`/`clipboard-write` permissions are granted at screencast start. Slow clipboard reads run in a spawned task so they can't freeze that client's screencast frames or input.

## Verification

- New (ignored) e2e `e2e_stream_input_dispatch_and_clipboard_bridge` drives copy/paste through the real stream path; unit test for the new input message types. Suite 1028/1028, `cargo fmt`/`clippy` clean, dashboard `tsc` + `next build` clean.
- Live dashboard: 300-event mouse sweep then click went from ~5s to ~35ms; Ctrl+A/C/V round-tripped text between the local clipboard and a textarea in the page, including unicode and a 100k-char paste.
- Adversarial review pass reproduced and fixed two follow-ups on this branch: slow clipboard reads head-of-line blocking frames/input (7e75ffc), and unguarded `navigator.clipboard` killing paste on insecure origins (521fdfe).

Fixes #1615.

### DIFF
--- a/cli/src/native/e2e_tests.rs
+++ b/cli/src/native/e2e_tests.rs
@@ -5535,6 +5535,197 @@ async fn e2e_stream_frame_metadata_respects_custom_viewport() {
     let _ = std::fs::remove_dir_all(&socket_dir);
 }
 
+// ---------------------------------------------------------------------------
+// Stream: dashboard input dispatch and clipboard bridge
+// ---------------------------------------------------------------------------
+
+#[tokio::test]
+#[ignore]
+async fn e2e_stream_input_dispatch_and_clipboard_bridge() {
+    use futures_util::SinkExt;
+
+    let guard = EnvGuard::new(&["AGENT_BROWSER_SOCKET_DIR", "AGENT_BROWSER_SESSION"]);
+    let socket_dir = std::env::temp_dir().join(format!(
+        "agent-browser-e2e-stream-input-{}-{}",
+        std::process::id(),
+        std::time::SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .expect("system clock should be after unix epoch")
+            .as_nanos()
+    ));
+    std::fs::create_dir_all(&socket_dir).expect("socket dir should be created");
+    guard.set(
+        "AGENT_BROWSER_SOCKET_DIR",
+        socket_dir.to_str().expect("socket dir should be utf-8"),
+    );
+    guard.set("AGENT_BROWSER_SESSION", "e2e-stream-input");
+
+    let mut state = DaemonState::new();
+
+    let resp = execute_command(
+        &json!({ "id": "1", "action": "stream_enable", "port": 0 }),
+        &mut state,
+    )
+    .await;
+    assert_success(&resp);
+    let port = get_data(&resp)["port"]
+        .as_u64()
+        .expect("stream enable should report the bound port");
+
+    let (mut ws, _) = tokio_tungstenite::connect_async(format!("ws://127.0.0.1:{port}"))
+        .await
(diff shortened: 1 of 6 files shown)


KachurPro:
Confirming the clipboard part of this PR from an independent reproduction on 0.33.2/0.34.0: paste can never work in the current viewport because the capture-phase `keydown` handler calls `preventDefault()` on every key while the canvas is focused, so the browser never fires a `paste` event at all — there is nothing for the page to receive.

As an interim fix we intercept Ctrl/Cmd+V in an earlier capture listener (registered before the viewport mounts), let the native `paste` event fire, and replay `clipboardData` text char-by-char over the same stream socket as `input_keyboard` keyDown/keyUp frames with `text` set and `windowsVirtualKeyCode: 0` (Enter frames for newlines). That works reliably, including multiline and non-ASCII text, so the protocol needs no changes for basic paste — but a proper clipboard bridge like this PR is the better long-term answer, and the latency/shortcut fixes here address real pain we also observed. +1.

PR #1623, opened 2026-07-29
fix(cli): reject unknown command options

### LINKED ISSUE #1622: Unknown --flags are consumed as positionals:  writes a file named --full-page
## Summary

Unknown `--flags` are not rejected. They are consumed as positional arguments, so a typo silently becomes a selector or an output path instead of an error.

## Reproduction

```
$ agent-browser open "data:text/html,<h1>shot test</h1>"

$ agent-browser screenshot body --full-page
✓ Screenshot saved to --full-page

$ ls
--full-page

$ file -- --full-page
--full-page: PNG image data, 756 x 469, 8-bit/color RGB, non-interlaced
```

`--full-page` is not a flag; the real one is `--full` / `-f`. Rather than saying so, `screenshot` took it as the `[path]` positional and wrote a PNG whose name starts with two dashes, which then needs `rm -- --full-page` or `rm ./--full-page` to remove.

With one positional it lands in the other slot instead:

```
$ agent-browser screenshot --full-page
✗ Element not found. Verify the selector is correct and the element exists in the DOM.
```

That message sends you looking at the page for an element named `--full-page`.

Unknown flags are also accepted silently on other commands:

```
$ agent-browser open --totally-bogus-flag "data:text/html,<h1>x</h1>"
✓
  data:text/html,<h1>x</h1>
```

Tested on macOS 26 (arm64), agent-browser 0.27.0.

## Why it matters

I found this because a repository working tree had an untracked 56 KB file named `--full-page` sitting in its root, committed by nobody and explained by nothing. An agent had written a screenshot there, been told `✓ Screenshot saved`, and moved on. Nothing in the output said the flag was wrong, so the mistake persisted as repo litter rather than surfacing at the point it was made.

For an agent-facing CLI this is the expensive kind of failure: the success checkmark actively confirms the wrong belief, so the agent does not retry with the correct flag. `--full` and `--full-page` are exactly the pair a model will guess between, since Playwright and Puppeteer both call it `fullPage`.

## Suggested fix

Reject any argument starting with `-` that is not a known flag for the command, and say what was passed. Something like:

```
✗ Unknown option `--full-page` for `screenshot`. Did you mean `--full`?
```

An explicit `--` separator can keep the escape hatch for the genuinely rare path that starts with a dash.

If the parser cannot distinguish flags from positionals generally, the narrower version still removes most of the cost: refuse a `[path]` or `[selector]` positional that begins with `--`, since neither is a plausible real value.

---

Filed by an AI agent (Claude); every command and output above was run on the machine described.

KXHXK:
Closes #1622

## Summary

- reject unknown dash-prefixed options for `open`, `goto`, `navigate`, and `screenshot`
- preserve an explicit `--` separator for dash-prefixed navigation targets and screenshot paths
- add regression coverage for the reported `--full-page` cases and the navigation typo case

## Root cause

Global argument cleanup intentionally leaves command-specific arguments in place. The navigation parser then skipped every `--`-prefixed argument, while the screenshot parser treated every option other than `--full`/`-f` as positional input. That allowed typos to be silently ignored or used as selectors and output paths.

## Validation

- added parser unit tests for unknown options before and after screenshot positionals
- added parser unit tests for navigation rejection and explicit `--` escaping
- `git diff --check`
- Rust build, formatting, and test execution are left to CI because a complete local Rust toolchain was unavailable

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -372,9 +372,26 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
             // scripts before the first real navigation (see `batch`).
             // `goto` and `navigate` still require a URL since those verbs
             // imply the navigation itself.
-            let first_url = rest.iter().find(|a| !a.starts_with("--"));
+            let mut first_url = None;
+            let mut options_ended = false;
+            for arg in &rest {
+                if !options_ended && *arg == "--" {
+                    options_ended = true;
+                } else if !options_ended && arg.starts_with('-') {
+                    return Err(ParseError::InvalidValue {
+                        message: format!("Unknown option '{}' for {}", arg, cmd),
+                        usage: if cmd == "open" {
+                            "open [url]"
+                        } else {
+                            "goto <url>"
+                        },
+                    });
+                } else if first_url.is_none() {
+                    first_url = Some(*arg);
+                }
+            }
             let url = match first_url {
-                Some(u) => *u,
+                Some(u) => u,
                 None if cmd == "open" => {
                     return Ok(json!({ "id": id, "action": "launch", "headless": !flags.headed }));
                 }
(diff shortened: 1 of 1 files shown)


KXHXK:
Refreshed this PR onto current main and resolved the screenshot parser conflict in 2e39fe6. The fix preserves the new --if-changed/--threshold options and URL-less open behavior. MCP navigation/screenshot argument builders now escape dash-prefixed literal inputs, with a CLI/MCP parity regression; help and documentation are synchronized. Local Windows validation: 314 command-parser tests and 71 MCP-related tests passed (cargo test with --no-default-features); cargo fmt --check and git diff --check passed. Browser E2E tests were not run. The updated head is ready for CI and review.

PR #1638, opened 2026-08-01
Support staged TOTP challenges from credential providers

gabeosx:
Credential providers can currently supply a username and password for `auth login`, but many sign-in flows ask for a time-based one-time password on a second page. This adds a staged challenge flow so agent-browser can request that code only after the OTP field appears.

The provider marks the initial credential response with a `stagedChallenge`. After agent-browser submits the username and password, it waits for the OTP field and sends a separate `credential.resolve.challenge` request to the same provider. The response is validated for freshness and filled into the page, but it is deliberately not submitted. The caller remains responsible for the final confirmation action.

Providers may include an opaque `challengeRef` in the initial response. agent-browser returns it unchanged in the challenge request, allowing a stateless provider process to resolve the code for the same vault item without exposing provider-specific identifiers to agent-browser. The challenge request also includes the current page origin so providers can enforce their own origin binding.

This PR includes:

- the new `credential.challenge` capability and challenge request/response types
- optional TOTP selector and `challengeRef` metadata on credential responses
- `--otp-selector` support for `auth login`
- expiry validation and fill-without-submit behavior
- plugin policy enforcement for challenge resolution
- CLI and MCP parity
- updated schemas, help text, documentation, and embedded skill references
- focused protocol, parser, policy, MCP, and browser tests

A compatible reference provider is available at [`gabeosx/agent-browser-plugin-bitwarden`](https://github.com/gabeosx/agent-browser-plugin-bitwarden). It resolves username, password, login URI, and staged TOTP values through the Bitwarden CLI while keeping Bitwarden-specific vault access outside agent-browser core.

Validation included formatting and Clippy checks, focused Rust tests, a browser E2E covering a two-page login with a six-digit OTP field, and an end-to-end login using an external credential provider. The staged step returned `challenge_filled` with `loggedIn: false`, and authentication completed only after a separate confirmation action.


gabeosx:
For maintainers who want to exercise the provider side of this protocol, I published a compatible reference implementation: [`agent-browser-plugin-bitwarden`](https://github.com/gabeosx/agent-browser-plugin-bitwarden).

It declares `credential.read` and `credential.challenge`, keeps Bitwarden CLI and vault-specific logic outside agent-browser core, passes an opaque `challengeRef`, verifies the live page origin, and resolves the current TOTP only when agent-browser sends `credential.resolve.challenge`.

PR #1026, opened 2026-03-25
Add --ca-cert flag for HTTPS interception proxy support

ctate:
Adds support for trusting specific CA certificates when using HTTPS-intercepting proxies (corporate egress proxies, security appliances like Zscaler/Netskope, development tools like mitmproxy).

## Problem
agent-browser fails with `ERR_CERT_AUTHORITY_INVALID` behind HTTPS-intercepting proxies that use custom CAs. The current `--ignore-https-errors` workaround disables ALL certificate validation, creating security risks.

## Changes
- Added `--ca-cert <path>` CLI flag and `AGENT_BROWSER_CA_CERT` environment variable
- Added `caCert` field to JSON configuration
- Implemented CA certificate validation via Chromium's `--ignore-certificate-errors-spki-list` flag
- Added early validation to check CA certificate file exists
- Added comprehensive test coverage for flag parsing and configuration merging
- Updated documentation in README and configuration pages

## Implementation Details
- Computes SHA-256 hash of the CA certificate's Subject Public Key Info (SPKI)
- Passes the hash to Chromium via launch args for targeted certificate trust
- Validates certificate file exists before daemon spawn to provide early feedback
- Maintains security by only trusting the specified CA, not disabling all validation

Fixes #1022

PR #1669, opened 2026-08-08
feat(cli): trust proxy CAs through Chromium NSS

### LINKED ISSUE #1022: No way to trust a specific CA certificate for HTTPS interception proxies
## Problem

agent-browser fails with `ERR_CERT_AUTHORITY_INVALID` when running behind an HTTPS-intercepting proxy (e.g., corporate egress proxies, security appliances like Zscaler/Netskope, or development tools like mitmproxy). These proxies use a custom CA to re-sign traffic, and the browser doesn't trust it.

### Why standard CA trust mechanisms don't work

- **`update-ca-certificates` / system CA store**: Chromium launched by agent-browser uses a temporary profile on each run, so it ignores the OS-level OpenSSL certificate store.
- **NSS database (`$HOME/.pki/nssdb`)**: Same issue — the temporary profile doesn't inherit user-level NSS trust.
- **`--args` pass-through**: Chromium doesn't expose a `--ca-cert-path` flag. The closest mechanism (`--ignore-certificate-errors-spki-list`) requires computing a hash from the certificate first, which users can't easily do.

### Current workaround

`--ignore-https-errors` (or `AGENT_BROWSER_IGNORE_HTTPS_ERRORS=1`) works, but it disables **all** certificate validation via CDP's `Security.setIgnoreCertificateErrors`. This is a security concern for automated agents that may follow links or navigate to arbitrary URLs — any MITM would be silently accepted, not just the known proxy.

### Who this affects

- Teams running agent-browser in corporate environments with DLP/egress proxies
- CI/CD pipelines behind security appliances
- Development setups using mitmproxy, Charles, or similar tools
- Cloud workspaces (Codespaces, devcontainers) with egress filtering

### Desired behavior

A way to tell agent-browser "trust this specific CA certificate" — equivalent to adding it to the browser's trust store — without disabling all certificate validation. Chromium supports this via `--ignore-certificate-errors-spki-list=<hash>`, which trusts specific certificate fingerprints rather than ignoring all errors. agent-browser could accept a path to a CA cert file, compute the SPKI hash internally, and pass it to Chromium at launch.

### Proposed interface

| Method | Example |
|---|---|
| CLI flag | `--ca-cert /etc/egress-proxy/ca/ca.crt` |
| Env var | `AGENT_BROWSER_CA_CERT=/etc/egress-proxy/ca/ca.crt` |
| Config file | `{ "caCert": "/etc/egress-proxy/ca/ca.crt" }` |

We have a working implementation with unit and e2e tests ready to submit as a PR at [`crunchloop/agent-browser@feat/ca-cert-spki-pinning`](https://github.com/crunchloop/agent-browser/tree/feat/ca-cert-spki-pinning).

Railly:
Recreates the useful parts of #1026 with real CA trust instead of a Chromium certificate-error exception. Original work by @ctate and @bilby91 is preserved through commit co-authorship.

## Problem

`agent-browser` fails with `ERR_CERT_AUTHORITY_INVALID` behind HTTPS-intercepting proxies that re-sign traffic with a private CA. `--ignore-https-errors` disables certificate verification broadly, which is a different contract.

The earlier implementation used `--ignore-certificate-errors-spki-list`. That primitive is not equivalent to installing a CA: it can bypass hostname and certificate verification for matching keys, and it can still reject a separately keyed leaf signed by the supplied CA when the server omits that CA from the presented chain.

## Change

`--ca-cert <path>` now:

- accepts a PEM bundle or one DER certificate through the shared `ca_bundle` loader;
- creates a private UUID-scoped HOME for each locally launched Linux Chromium process;
- initializes an NSS Shared DB with `certutil` and imports the normalized certificates as trusted CAs;
- installs the package providing `certutil` through first-party CLI, Eve, sandbox, environment, and benchmark setup paths;
- scopes `HOME` and `XDG_DATA_HOME` only to the Chromium child;
- preserves normal hostname, validity-period, and unrelated-authority verification;
- keeps the daemon alive and stores the effective CA in daemon state;
- treats omission on later commands as preservation, so daemon PID, browser target, URL, page, and trust remain intact;
- treats equivalent normalized certificate content as the same CA, even when supplied from another path or with reordered or duplicate bundle entries;
- relaunches only Chromium when certificate content changes or trust is explicitly cleared with `--no-ca-cert`;
- prepares replacement NSS trust before closing the active browser and commits sticky state only after launch succeeds;
- removes the private profile and NSS HOME after the owned Chrome process tree is reaped.

The option is also available through `AGENT_BROWSER_CA_CERT`, `caCert` in config and MCP, and the Eve extension. Explicit clear is available through `AGENT_BROWSER_CLEAR_CA_CERT`, `clearCaCert`, and `--no-ca-cert`.

CLI, daemon, MCP, and Eve reject simultaneous set and clear.

## Session behavior

```bash
agent-browser --ca-cert ca.pem open https://internal
agent-browser snapshot
```

The second command inherits the effective CA and reuses the live Chromium session. Repeating the same certificate content also reuses it. Selecting different certificate content or explicitly clearing the CA safely replaces Chromium without restarting the daemon.

CA state uses three transitions:

- set: use the supplied CA;
- omit: retain the current CA;
- clear: remove the retained CA.

The transition is preserved across local, provider, CDP, auto-connect, and parsed `connect` launch envelopes. A retained CA blocks an incompatible external launch with an actionable `--no-ca-cert` remedy. Explicit clear permits the transition and commits only after launch success.

## Scope

Installing a CA is supported only for locally launched Chromium on Linux.

`--ca-cert` rejects:

- macOS and Windows;
- `--profile`;
- `--cdp` and `--auto-connect`;
- providers;
- Lightpanda;
- combination with `--ignore-https-errors`;
- combination with `--no-ca-cert`.

`--no-ca-cert` remains valid when transitioning to provider, CDP, auto-connect, iOS, or Safari launches so retained local CA state does not leak into the new launch family.

## Verification

The Linux ARM64 Chromium matrix covered:

| Case | Result |
| --- | --- |
| Supplied CA, correct hostname, CA omitted from served chain | accepted |
| CA omitted from configuration | rejected |
| Wrong hostname | rejected |
| Unrelated CA | rejected |
| Expired certificate | rejected |
| Not-yet-valid certificate | rejected |
| Concurrent daemons with different CAs | isolated, both accepted |
| Missing or failing `certutil` | actionable error, no profile leak |
| Normal close, Chrome crash, daemon SIGTERM | NSS HOME removed |
| Daemon SIGKILL | expected bounded orphan, since no destructor can run |

Exact-head local gates:

- 1,144 Rust tests passed, 101 ignored;
- 2 doctor CLI integration tests passed;
- `cargo clippy --all-targets -- -D warnings`;
- `cargo fmt`;
- Eve typecheck and 42 tests;
- sandbox package 18 tests;
- version sync with no diff;
- Linux ARM64 Docker build;
- clean Debian Bookworm install of `libnss3-tools` with `/usr/bin/certutil`;
- clean Amazon Linux 2023 install of `nss-tools` with `/usr/bin/certutil`;
- force-red coverage for parsed `connect` clear propagation;
- exact-head Review Gate coverage with independent Fable 5 and Grok 4.6 challenge.

Fixes #1022. Supersedes the closed #1026 implementation while preserving contributor credit.

## Separate follow-up

This PR changes Chromium's trust store. The CLI's own outbound HTTPS uses a separate Rust trust store. #1670 remains the stacked follow-up for that independent connection boundary.


modi95:
Excited for this!

ctate (approved):

PR #1670, opened 2026-08-08
feat(cli): trust private CAs in CLI TLS

Railly:
Stacked on #1669. Reported by Notion for Vercel Sandbox.

## Problem

An intercepting proxy can break TLS in two independent processes:

1. Chromium rejects certificates for pages it loads. #1669 handles this for local Chromium on Linux with isolated NSS trust.
2. The Rust CLI rejects certificates on its own outbound connections before or outside the browser. This PR handles that trust store.

The CLI previously used compiled-in Mozilla roots only, so private proxy CAs installed by a sandbox, MDM, or enterprise image were invisible.

## Contract

- `--ca-cert <path>` always adds the supplied PEM bundle or DER certificate to CLI-owned TLS trust.
- For a local Chromium launch on Linux, the same option also configures isolated browser trust through #1669 and requires `certutil`.
- With `--cdp`, `--auto-connect`, or a provider, `--ca-cert` affects CLI TLS only. It does not attempt to modify an attached browser.
- For CLI-only actions such as `read`, `--ca-cert` does not trigger a browser launch.
- `--use-system-ca` affects CLI TLS only and selects the operating system trust store instead of the compiled-in roots.
- `SSL_CERT_FILE` is accepted as an ambient CA-bundle fallback. If it is stale or unusable, the CLI warns and keeps usable roots instead of breaking unrelated commands.

Certificate and hostname verification remain enabled. These options change trust roots; they do not bypass TLS verification.

## Change

Adds a shared trust layer in `cli/src/tls.rs` and routes every outbound client through it:

- CDP WebSocket connections and discovery
- provider APIs
- `read`
- Chrome installation and upgrade checks
- doctor network probes
- AI Gateway chat

The configuration is available through CLI flags, environment variables, config files, MCP, Eve, and the published schemas. CLI help, README, proxy docs, configuration docs, commands docs, and skill docs describe the same split contract.

## Mode matrix

| Mode | `--ca-cert` effect |
|---|---|
| Local Chromium on Linux | CLI TLS plus isolated Chromium NSS trust |
| Local Chromium on macOS or Windows | CLI TLS works; browser-side trust is rejected with an actionable error |
| `--cdp` or `--auto-connect` | CLI TLS only |
| Provider | CLI TLS only |
| CLI-only action such as `read` | CLI TLS only; no browser launch |

`--use-system-ca` is CLI-only in every mode.

## Verification

- `cargo fmt -- --check`
- `cargo clippy -- -D warnings`
- `cargo clippy --all-targets --all-features -- -D warnings`
- Rust serial suite: 1,148 passed, 101 ignored, 0 failed
- Doctor integration: 2 passed
- Eve typecheck and 40 tests passed
- `git diff --check`

Built-artifact dogfood:

- `read https://example.com` with `--ca-cert` succeeded with `browserLaunched:false`.
- CDP with `--ca-cert` reached the remote connection attempt instead of failing local-browser validation.
- Provider mode with `--ca-cert` reached provider credential validation.
- A local Chromium launch on macOS rejected only the unsupported browser-trust half.

## Stack lifecycle

This PR remains based on #1669 while the shared `--ca-cert` plumbing and Linux Chromium trust land there. After #1669 merges, this PR should be retargeted to `main`.


ctate:
**cli/src/main.rs:243**

Exclude browserless actions such as `read` when deciding whether CA configuration requires a launch command. As written, `--ca-cert` turns a direct HTTP read into an NSS/Chromium setup operation.

ctate:
**cli/src/connection.rs:599-607**

Include the effective CA bundle and system-store selection in the daemon configuration fingerprint, or pass trust settings per command. Otherwise a reused session permanently keeps the trust roots from its first invocation.

SugeethK:
Independent confirmation of the CLI-side half of this, from macOS behind Zscaler TLS inspection. Adding it because the environment differs from the Vercel Sandbox and Netskope reports already on the thread, and because one of the review notes below matches something I verified empirically today.

**Environment:** agent-browser 0.35.2 (npm, darwin-arm64), macOS 26.6.2, Zscaler ZIA SSL inspection. Corporate root CA present and trusted in the macOS System keychain, so Safari, Chrome, Apple `curl`, and `git` all validate the intercepted chain without complaint.

**Current release behavior**

Both the installer and `read` fail on any TLS-inspected host:

```
$ agent-browser install
Installing Chrome...
✗ Failed to fetch version info: error sending request for url (https://googlechromelabs.github.io/chrome-for-testing/last-known-good-versions-with-downloads.json): client error (Connect): invalid peer certificate: UnknownIssuer

$ agent-browser read https://example.com
✗ Read request failed: error sending request for url (https://example.com/): client error (Connect): invalid peer certificate: UnknownIssuer
```

Confirming the diagnosis in this PR against the shipped 0.35.2 binary — no hits for `rustls_native_certs`, `SecTrustEvaluate`, `SSL_CERT_FILE`, or `SSL_CERT_DIR`, so there is currently no local way to supply a trust root to the CLI on macOS:

```
$ strings agent-browser-darwin-arm64 | grep -c rustls              # 43
$ strings agent-browser-darwin-arm64 | grep -c rustls_native_certs # 0
$ strings agent-browser-darwin-arm64 | grep -c SecTrustEvaluate    # 0
$ strings agent-browser-darwin-arm64 | grep -c SSL_CERT_FILE       # 0
```

What the existing knobs do on macOS today, for the record:

| Attempt | Result |
|---|---|
| `--ca-cert` / `AGENT_BROWSER_CA_CERT` / `"caCert"` in config | `✗ --ca-cert is currently supported only on Linux` |
| `"ignoreHttpsErrors": true` | no effect on `install` or `read` — identical `UnknownIssuer` |
| `SSL_CERT_FILE`, `SSL_CERT_DIR`, `NODE_EXTRA_CA_CERTS` | ignored |

**Re: the `read` review note (`cli/src/main.rs:243`)**

Worth flagging that this is the highest-value case in the PR from an enterprise-network standpoint, not an edge case. Browser-driven commands already work behind interception without any of this, because Chromium reads the OS trust store:

```
$ agent-browser open https://googlechromelabs.github.io/chrome-for-testing/
✓ Chrome for Testing availability
$ agent-browser snapshot
- heading "Chrome for Testing availability" [level=1, ref=e1]
```

So `read` and `install` are the only two commands actually broken by interception, and `read` is the one with no workaround at all — `install` can be sidestepped with `executablePath` pointed at an existing Chrome. If `--ca-cert` on a `read` triggers Chromium/NSS setup, the fix ends up not covering the case that needs it most. The split contract in the PR description reads correctly on this; it's just worth confirming the reviewer's note is load-bearing rather than cosmetic.

**Re: the daemon fingerprint note (`cli/src/connection.rs:599-607`)**

Also agree this matters in practice. Sessions here are long-lived, and a root cert rotation mid-session would otherwise silently keep the old trust roots until someone thinks to kill the daemon.

**One request on scope:** `--use-system-ca` reading the OS trust store is the option that makes this work with zero configuration on a managed fleet, where the CA is already deployed to the keychain by MDM. `--ca-cert` requires every developer to locate and reference a PEM by hand. If the two land separately, `--use-system-ca` is the one that removes the support burden. Node took this route in 22.15 and it works well — `NODE_OPTIONS=--use-system-ca` is what I ended up using to fix every other Node tool on this machine.

Happy to test a preview build on this setup if that's useful — real Zscaler interception on macOS with the CA in the System keychain is easy for me to reproduce on demand.

cli/src/connection.rs
```diff
@@ -513,6 +515,15 @@ fn apply_daemon_env(cmd: &mut Command, session: &str, opts: &DaemonOptions) {
     if opts.ignore_https_errors {
         cmd.env("AGENT_BROWSER_IGNORE_HTTPS_ERRORS", "1");
     }
+    if let Some(ca) = opts.ca_cert {
+        cmd.env("AGENT_BROWSER_CA_CERT", ca);
+    } else {
+        cmd.env_remove("AGENT_BROWSER_CA_CERT");
```
ctate, line 521: The session’s TLS selection is lost whenever the daemon restarts for another config change. Could you carry the effective trust options into the replacement daemon or next command? `--ca-cert ... read` followed by `--debug read` currently ends with `UnknownIssuer`.

ctate (changes requested):

PR #1649, opened 2026-08-04
fix: reset reference (ref) numbers on `diff snapshot`

Angelmmiguel:
`diff snapshot` doesn't reset the ref map before taking the new snapshot. The daemon's monotonic ref counter kept climbing across commands. This caused the command to always show differences. You can find an example on #1394 , although I could also reproduce on my side.

Closes #1394 

## Reproduce bug

Running `agent-browser snapshot > base.txt` followed by `agent-browser diff snapshot --baseline base.txt` reported every line as changed, with the offset growing on each subsequent run (`e1 -> e84 -> e167`). `diff url` had the same gap on its first snapshot: it cleared before the second one only, so the two sides were numbered from different bases.

## Implementation

The fix only required to clear the references before taking the new snapshot in the `diff snapshot` command. This reproduces the same behavior of `snapshot`. I tried to minimize the changes in the codebase to keep the PR clean.

## Tests

- Unit coverage for the `RefMap::clear()` contract, 
- E2E tests for a self-diff reporting no changes across repeated runs. The existing `e2e_diff_snapshot` only checked that something changed, which was always true.

## Manual testing

To confirm the fix, I built a development released and tested it manually. You can find the dev release on my fork: https://github.com/endorhq/agent-browser/releases/tag/v0.33.2-fix-dev-v1 . That release contains this and a minor feature related to the `-i / --interactive` argument for `diff snapshot`.

Here you can see how the command only shows the differences now:

```bash
agent-browser diff snapshot -i -c --baseline diff.txt
```

```diff
--- before
+++ after
@@ -9,9 +9,9 @@
 - combobox "Selecciona el tipo de billete. Solo ida" [expanded=false, ref=e72]: Solo ida
 - button "1 pasajero" [ref=e73]
 - combobox "Selecciona la clase de asiento. Turista" [expanded=false, ref=e74]: Turista
-- combobox "¿Desde dónde?" [expanded=false, ref=e75]: Ámsterdam
+- combobox "¿Desde dónde?" [expanded=false, ref=e75]: Madrid
 - button "Intercambiar el origen y el destino." [ref=e35]
-- combobox "¿A dónde quieres ir?" [expanded=false, ref=e82]: Madrid
+- combobox "¿A dónde quieres ir?" [expanded=false, ref=e82]: Ámsterdam
 - textbox "Salida" [ref=e83]: lun, 10 ago
 - button "Buscar" [ref=e28]
[...]

PR #1672, opened 2026-08-08
fix: reset refs before diff snapshots

### LINKED ISSUE #1394: diff always only displays additions
Image

iroiro147:
## Summary

- Reset snapshot reference mappings before taking a diff snapshot.
- Add an end-to-end regression assertion that an unchanged page produces an empty diff.

Fixes #1394

## Verification

local verification not run; relying on upstream CI

The fresh worktree has no installed Node/Rust dependencies, so the local e2e suite could not be run.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -5746,6 +5746,8 @@ async fn handle_diff_snapshot(cmd: &Value, state: &mut DaemonState) -> Result<Va
         selector,
         ..SnapshotOptions::default()
     };
+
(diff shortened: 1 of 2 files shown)


Angelmmiguel:
Related: #1649

iroiro147:
Confirmed #1649 is the broader superseding implementation: it applies the same ref_map.clear() fix before handle_diff_snapshot, and also covers diff_url plus regression tests. Closing #1672 as a duplicate to reduce review noise. Thanks for the pointer.

PR #1675, opened 2026-08-10
feat: discover Helium during auto-connect

mkromann:
## Summary

- discover Helium on macOS by checking its default user-data directory for DevToolsActivePort
- update CLI, schema, README, docs, and the bundled core skill to describe supported Chromium browsers
- add a macOS regression test for the Helium path

This is a current-main successor to #1063 and credits that PR for identifying the Helium user-data path. Unlike #1063, this change includes only the verified current Helium bundle path, net.imput.helium.

No MCP change is needed because this extends the existing auto-connect behavior without adding a new flag or command.

## Checks

- cargo test --manifest-path cli/Cargo.toml (1062 passed, 97 ignored)
- targeted Helium and CDP resolution tests
- pnpm -C docs build
- git diff --check

Strict Clippy currently stops on a pre-existing nonminimal_bool warning at cli/src/native/cdp/chrome.rs:337, outside this diff.


mkromann:
Closing at the fork owner's request; this will remain a personal fork change.

PR #1682, opened 2026-08-11
fix(stream): emit active main-frame URL updates

### LINKED ISSUE #1677: Please emit Page.navigatedWithinDocument events from CDP
## Version

- agent-browser `0.33.2`
- Also present on current main as of `2026-08-10`

## Bug

The native WebSocket browser stream does not emit a URL update after
same-document navigation.

The streamed frames show the new page, but stream consumers keep the old URL.  We use the emitted url to show our users which url the browser is on.

## Root cause

`cli/src/native/stream/cdp_loop.rs` emits `{"type":"url"}` only when it receives
`Page.frameNavigated`.

Chrome sends `Page.navigatedWithinDocument` for navigation caused by:

- `history.pushState`
- `history.replaceState`
- client-side routers
- anchor navigation

The stream does not handle this event.

## Reproduction

1. Open a site with client-side navigation:

   agent-browser --session url-stream open https://cofounder.co/

2. Enable the stream and connect a WebSocket client to the returned URL:

   agent-browser --session url-stream stream enable --json

3. Click the `Start` navigation link:

   agent-browser --session url-stream find role link click --name "Start"

4. Confirm that the browser URL changed:

   agent-browser --session url-stream get url

The browser reports:

https://cofounder.co/how-to/start

The stream continues to send frames for the new page, but it does not send a `type: "url"` message for the route change.

## Expected behavior

The stream sends a URL message for every main-frame URL change, including same-document navigation.

## Actual behavior

The stream sends URL messages only for full document navigation. Client-side navigation can change the page and browser URL without updating stream consumers.

## Suggested fix

Handle `Page.navigatedWithinDocument` in the native CDP event loop.

The handler should:

1. Confirm that the event belongs to the main frame.
2. Update the cached active-tab URL.
3. Emit the existing `{"type":"url"}` stream message.

Please add regression coverage for:

- History API navigation
- anchor navigation
- full document navigation
- same-document navigation in a child frame, which must not update the main URL

Railly:
## Summary

- emit URL updates for active main-frame navigations in the stream
- cover full navigations, History API changes, and fragment updates in Chrome
- ignore child-frame and background-session events while rebinding the active session and cached tab atomically
- support Lightpanda full-navigation URL updates with bounded seeding, retry, and shutdown cleanup
- document stream URL behavior and add real-browser regression coverage

Fixes #1677

## Verification

- cargo fmt -- --check
- git diff --check
- cargo test -- --test-threads=1: 1,122 passed, 106 ignored
- Chrome navigation stress: 100 iterations with seed 1677
- Chrome navigation stress: 20 iterations each with seeds 1677, 42, and 9001
- Lightpanda 0.3.6 stream E2E: 10 consecutive passes
- all three Lightpanda E2Es together
- cargo clippy --all-targets -- -D warnings
- force-red checks confirmed for Chrome and Lightpanda regressions

## Lightpanda note

Lightpanda currently rejects opening a second page with TargetAlreadyLoaded, consistent with lightpanda-io/browser#1962. Its full-navigation path is covered by a real Lightpanda E2E; background-session filtering is covered by the deterministic CDP harness.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -2791,17 +2791,10 @@ pub async fn execute_command(cmd: &Value, state: &mut DaemonState) -> Value {
 
         if let Some(ref mgr) = state.browser {
             let tabs = mgr.tab_list();
-            if matches!(
-                action,
-                "tab_new" | "tab_switch" | "tab_close" | "open" | "navigate"
-            ) {
-                let session_id = mgr.active_session_id().ok().map(|s| s.to_string());
-                server
-                    .bind_cdp_session_and_broadcast_tabs(session_id, &tabs)
-                    .await;
-            } else {
-                server.broadcast_tabs(&tabs).await;
-            }
+            let session_id = mgr.active_session_id().ok().map(|s| s.to_string());
+            server
+                .bind_cdp_session_and_broadcast_tabs(session_id, &tabs)
+                .await;
         }
     }
 
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -857,22 +857,21 @@ impl DaemonState {
                 .browser
                 .as_ref()
                 .and_then(|m| m.active_session_id().ok().map(|s| s.to_string()));
-            server.set_cdp_session_id(session_id).await;
-
-            // Broadcast connection status change to WebSocket clients
             let connected = self.browser.is_some();
             let sc = server.is_screencasting().await;
             let (vw, vh) = server.viewport().await;
+            if let Some(ref mgr) = self.browser {
+                server
+                    .bind_cdp_session_and_broadcast_tabs(session_id, &mgr.tab_list())
+                    .await;
+            } else {
+                server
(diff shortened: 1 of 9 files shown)


cli/src/native/actions.rs
```diff
@@ -2710,17 +2709,17 @@ pub async fn execute_command(cmd: &Value, state: &mut DaemonState) -> Value {
...
                 "tab_new" | "tab_switch" | "tab_close" | "open" | "navigate"
             ) {
                 let session_id = mgr.active_session_id().ok().map(|s| s.to_string());
-                server.set_cdp_session_id(session_id).await;
-                server.notify_client_changed();
+                server
+                    .bind_cdp_session_and_broadcast_tabs(session_id, &tabs)
+                    .await;
+            } else {
+                server.broadcast_tabs(&tabs).await;
```
ctate: Could we rebind whenever draining CDP events changes the manager's active page, rather than only for this action-name allowlist? An externally opened tab can become active during an unrelated command while the stream keeps listening to the previous session.

PR #1683, opened 2026-08-11
feat: isolate shared Chrome sessions with BrowserContexts

### LINKED ISSUE #1068: [Feature Request] CDP BrowserContext support for cookie-isolated parallel sessions
## Problem

When multiple agents connect to the same Chrome instance via `--auto-connect` or `connect <port>`, all sessions share the default BrowserContext. This means cookies, localStorage, and sessionStorage leak between sessions.

The `--session` flag provides tab-namespace isolation (each session tracks its own active tab), but does **not** create a CDP `Target.createBrowserContext` — so there is no storage isolation.

This is the root cause behind #86, #326, and #896 when agents use `--auto-connect` mode. PR #899 addresses the daemon-mode case (separate `--profile` per session), but `--auto-connect` has no equivalent fix.

## Use Case

I run multiple AI agents against one real Chrome instance (launched with `--remote-debugging-port=9223`). Each agent handles a different service:

- Agent A: university admin panel (`*.monash.edu`)
- Agent B: personal cloud admin (`*.google.com`)
- Agent C: infrastructure (`*.tailscale.com`)

Today this works because the domains don't overlap — browser same-origin policy provides natural isolation. But if two agents need to interact with the **same domain** using different credentials (e.g., two Google accounts, or staging vs production on the same host), cookies collide and the second login overwrites the first.

## Proposal

Expose CDP `Target.createBrowserContext` via a new `--context <name>` flag (or `--incognito`). When present, agent-browser would:

1. Call `Target.createBrowserContext({ disposeOnDetach: true })` on the browser-level CDP session
2. Create new tabs within that context via `Target.createTarget({ browserContextId: ... })`
3. Scope all session commands to pages within that context
4. Dispose the context on `close` (or rely on `disposeOnDetach`)

### Proposed CLI surface

```bash
# Create an isolated context connected to running Chrome
agent-browser --session work --context work connect 9223
agent-browser --session work open https://admin.example.com

# Separate isolated context, same Chrome, different cookies
agent-browser --session personal --context personal connect 9223
agent-browser --session personal open https://admin.example.com

# Cookies are fully isolated between contexts
```

### Why `createBrowserContext` over separate profiles

| Approach | Cookie isolation | Window count | Chrome processes | Startup time |
|----------|-----------------|--------------|-----------------|--------------|
| Separate `--profile` per session | Full | N windows | N processes | Slow (Chrome launch per session) |
| CDP `createBrowserContext` | Full | **1 window** | **1 process** | Fast (context creation is ~5ms) |

CDP BrowserContexts are the same mechanism Playwright and Puppeteer use for parallel isolated sessions. Tabs from different contexts can share one Chrome window — the context is a storage boundary, not a windowing boundary.

## What This Enables

- True parallel multi-agent automation against one Chrome window
- Cookie-isolated sessions without window proliferation
- `disposeOnDetach` auto-cleans contexts when agents disconnect (no orphaned state)
- Parity with Playwright's `browser.newContext()` for users migrating from Playwright-based setups

## Environment

- agent-browser v0.23.0
- Chrome v146.0.7680.165 (real Chrome, `--remote-debugging-port`)
- macOS Darwin 25.3.0

## Related Issues

- #86 — Support for Multi Session
- #326 — Multiple agents hijacking each other
- #896 — `--session` daemons sharing same `--profile`
- #899 — Fix: scope profiles to session IDs (daemon mode only)
- #1028 — Session close does not kill daemon (BrowserContext suggested as option 2)

Happy to provide more details or help with testing.

Railly:
## Summary

- add `--isolate-context` and `AGENT_BROWSER_ISOLATE_CONTEXT` for cookie, storage, cache, and target isolation when sessions share Chrome over CDP
- persist a versioned, fail-closed BrowserContext binding across daemon restarts and recreate it safely after Chrome restarts
- scope target discovery, auto-attach, tab/window/state operations, recording child contexts, shutdown, and explicit close to the owned context boundary
- document the flag across CLI help, schemas, MCP, README, web docs, and skill data

## Why

Named sessions isolate daemon state and tab selection, but sessions attached to the same Chrome still share the default BrowserContext. Two agents using different credentials on the same origin can overwrite each other’s cookies and storage.

This implements the session-owned BrowserContext shape proposed in #1068 while preserving existing non-isolated behavior. It incorporates and extends the work explored in #1117 and #1340.

## Lifecycle

- daemon restart, version restart, idle shutdown, and signal shutdown preserve the persisted primary context
- internal shutdown disposes non-persisted child contexts before autosave
- explicit `close` disposes child and primary contexts owned by the session, without closing shared Chrome or foreign targets
- Chrome restart creates a clean replacement context and reapplies configured restore state
- enabling isolation on an already-open non-isolated connection fails with reconnect guidance and does not mutate sticky state

## Verification

- full serial suite: 1,117 passed, 0 failed, 104 ignored
- doctor tests: 2 passed
- all three isolated Chrome E2Es passed together
- docs build passed
- force-red verified late-isolation mutation and child-context shutdown cleanup
- seven additional Test Strength mutations failed as intended
- Solution Gate and Review Gate completed before push

## Real-world stress coverage

Added an ignored integration test that runs the compiled CLI through two isolated daemons sharing one real Chrome. It alternates tabs and windows, creates cross-origin page popups, mutates cookies/localStorage/sessionStorage/IndexedDB/Cache API, saves state concurrently, restarts one daemon while the other stays active, and verifies close removes only owned targets and bindings.

- 144 total rounds passed across fresh-Chrome campaigns
- one 100-round repeated campaign passed without intermittent failures
- three force-red mutations independently failed: unscoped tab creation, foreign-target adoption, and skipped close disposal
- replay: `AGENT_BROWSER_ISOLATION_STRESS_ROUNDS=20 cargo test --manifest-path cli/Cargo.toml --test isolate_context_cli shared_chrome_real_world_isolation_stress -- --ignored --nocapture --test-threads=1`

Fixes #1068

Co-authored-by: @nathanvale
Co-authored-by: @mvanhorn
Co-authored-by: @pedro-saraiva88


Railly:
Real-world stress update:

- Added `cli/tests/isolate_context_cli.rs`, an ignored integration test that drives the compiled CLI through two isolated daemons sharing one real Chrome.
- Covers concurrent `tab new` / `window new`, cross-origin `window.open` popups, cookies, localStorage, sessionStorage, IndexedDB, Cache API, parallel state saves, SIGTERM/reconnect of one daemon, and explicit close cleanup.
- 144 total rounds passed. A five-campaign run completed 100 rounds without an intermittent failure.
- Force-red verified the test detects three separate regressions: missing `browserContextId` on `tab new`, accepting foreign targets, and omitting owned-context disposal on close.
- Full serial suite remains green: 1,117 passed, 0 failed, 104 ignored; doctor tests 2/2; fmt and clippy pass.

Replay:
```bash
AGENT_BROWSER_ISOLATION_STRESS_ROUNDS=20 cargo test --manifest-path cli/Cargo.toml --test isolate_context_cli shared_chrome_real_world_isolation_stress -- --ignored --nocapture --test-threads=1
```

ctate:
**cli/src/native/browser.rs:2336-2340**

Include the session's isolated context ID when configuring downloads. Omitting it applies the behavior only to Chrome's default context, so isolated sessions cannot complete the `download` command.

cli/src/native/actions.rs
```diff
@@ -2346,6 +2395,20 @@ pub async fn execute_command(cmd: &Value, state: &mut DaemonState) -> Value {
...
+                .browser
+                .as_ref()
+                .is_some_and(|manager| !manager.is_isolated())
+        {
+            return error_response(
+                &id,
+                "--isolate-context requires reconnecting with --cdp, --auto-connect, or connect <port|url>",
+            );
+        }
+        state.isolate_context = true;
```
ctate: Commit the sticky isolation state only after a compatible isolated connection succeeds. This assignment runs before launch validation, so one invalid request changes later flagless commands.

ctate:
Isolated sessions still install Fetch/auth handling on foreign targets when credentials are configured. I reproduced a default-context page authenticating with the isolated agent’s `alice:secret` credentials. Could you gate the auto-attach and Fetch handlers by owned BrowserContext while still resuming foreign targets?

ctate (changes requested):

PR #1686, opened 2026-08-12
fix(state): keep autosave target hidden

emmanuelchucks:
## Problem

State save collects cross-origin storage through a disposable `about:blank` CDP target. `Target.createTarget` currently receives only the URL, so headed Chrome can show or focus a blank scratch page during manual or automatic state capture.

## Change

Create the storage scratch target with both `background: true` and `hidden: true`. There is no visible fallback: if the browser does not support a hidden target, state capture reports the existing CDP failure instead of opening a page.

A focused unit test locks the exact target payload.

## Verification

- `git diff --check` passes on current `main`.
- The same runtime change was built from `v0.33.2` with Rust 1.97.1 and exercised against headed Chrome on macOS: the scratch target was created and closed, cross-origin localStorage restored into a fresh profile, and the foreground app remained unchanged through state save.
- Current-main Rust checks were not run locally because this machine no longer has a Rust toolchain; CI owns current compilation, formatting, clippy, and unit-test verification.

### DIFF
--- a/cli/src/native/state.rs
+++ b/cli/src/native/state.rs
@@ -9,8 +9,8 @@ use std::path::PathBuf;
 
 use super::cdp::client::CdpClient;
 use super::cdp::types::{
-    AttachToTargetParams, AttachToTargetResult, CloseTargetParams, CreateTargetParams,
-    CreateTargetResult, EvaluateParams,
+    AttachToTargetParams, AttachToTargetResult, CloseTargetParams, CreateTargetResult,
+    EvaluateParams,
 };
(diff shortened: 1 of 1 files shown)


emmanuelchucks:
Closing after a stricter maintainer-fit review. The added unit test only asserted a helper literal rather than the state-save behavior, and I had not validated the experimental `hidden` field across supported CDP paths. In the current save path, scratch-target errors can also be swallowed, so unsupported-field failure could silently omit cross-origin storage. I do not want to leave maintainers with an inadequately validated patch. Sorry for the review noise.

Issue #1688, opened 2026-08-13
`state save` opens a visible temporary tab and brings headed Chrome to the foreground


emmanuelchucks:
## Summary

When `state save` needs to collect localStorage from a previously visited
origin, it opens a temporary `about:blank` tab.

In headed Chrome on macOS, that tab appears in the tab strip and brings Chrome
in front of the application the user is working in. Periodic autosave uses the
same code path and causes the same interruption.

## Environment

- agent-browser v0.34.0
- macOS 26.5.2
- Chrome 151

I reproduced this against both the v0.34.0 source and the published npm package.

## Reproduction

Use a headed session and write localStorage on two origins:

```bash
agent-browser \
  --headed \
  --profile /tmp/agent-browser-state-focus-profile \
  --session state-focus \
  open https://example.com

agent-browser --session state-focus \
  storage local set origin-a value-a

agent-browser --session state-focus \
  open https://httpbin.org/html

agent-browser --session state-focus \
  storage local set origin-b value-b
```

Bring another application to the foreground, then save state:

```bash
agent-browser --session state-focus \
  state save /tmp/agent-browser-state-focus.json
```

## Actual behavior

While state is being saved:

- Chrome comes to the foreground.
- A temporary `about:blank` tab appears.
- The temporary tab navigates to the previously visited origin and is then
  closed.

The saved state includes both origins, but saving it interrupts the user.

## Expected behavior

State saving should:

- keep its temporary page out of the tab strip;
- leave the foreground application and visible Chrome tab unchanged;
- continue to save localStorage from all visited origins.

## Cause

`collect_storage_via_temp_target` in `cli/src/native/state.rs` creates the
temporary page with:

```json
{ "url": "about:blank" }
```

Chrome's `Target.createTarget` command supports:

```json
{
  "url": "about:blank",
  "background": true,
  "hidden": true
}
```

A hidden target remains available through CDP but is not shown in the tab
strip.

The `hidden` field is marked experimental. There is also a compatibility issue
to resolve: the current `save_state` code ignores errors from creating or using
the temporary target. If a browser rejects `hidden`, `state save` can therefore
report success while omitting localStorage from the other origins.

## Tested locally

I tested a local change that:

- uses `hidden: true` and `background: true` only for this temporary target;
- leaves ordinary tab creation unchanged;
- returns an error instead of replacing the state file with incomplete data
  when the temporary target cannot be created.

With that change:

- cmux remained in the foreground during manual and periodic saves;
[...]

Issue #1247, opened 2026-04-16
Background Mode to prevent stealing focus when connected to already-running Chrome instance


yangyaofei:
## Problem

When connected to an existing Chrome via `--cdp` or `--auto-connect`, `tab new` and `tab switch` commands activate the target tab via CDP's `Target.createTarget` (which activates by default) and `Page.bringToFront`. This pulls the user's current tab away, making it impossible for an AI agent to operate in the background without disrupting the user.

Additionally, the `navigate` (`open`) command never called `Page.bringToFront`, so opened pages were not brought to the foreground even in normal (non-background) mode.

## Proposed Solution

Add a `--background` flag (env: `AGENT_BROWSER_BACKGROUND`) that:

1. New tabs (tab_new): Sets the CDP background: true parameter on Target.createTarget, so new tabs open in the background without activating
2. Tab switches (tab_switch): Skips Page.bringToFront so the user's current view is not disturbed
3. Navigation (navigate): Skips Page.bringToFront after navigation
4. Auto-connect: Respects the flag when creating the initial fresh tab, so the user's existing tabs are not disturbed on attach

The flag can be set globally (daemon-level) and also passed per-command, allowing agents to opt into background behavior for specific operations.

mikirejf:
PR #1695 fixes this without a new flag. It passes `background: true` to every `Target.createTarget` and removes the implicit `Page.bringToFront` on tab switch and auto-connect. I confirmed it on macOS with a shared headed Chrome over `--cdp`. Details are in the PR thread. The branch needs a small rebase onto current `main`.

yangyaofei:
@mikirejf hope it can be merged. For this long time waiting, I decided to create a CDP proxy to rewrite any cdp message I want to archive this...

And I hope I am wrong about no mantainer will merge that PR, 🥹

raine:
This is really annoying

AlexHerranr:
Rebased #1695 onto current `main` in #1880, with the `handle_recording_start` hunk dropped as @mikirejf described, plus the piece that was missing: `bringtofront` existed in the daemon but no CLI command reached it, so once the implicit raise is gone there is no way to raise the window. Measurements for Linux/X11 are in the PR.

[...]

PR #1695, opened 2026-08-15
Create tabs in the background; drop implicit bringToFront on switch and auto-connect

ashot:
## Summary

Create every tab in the background (`Target.createTarget` with `background: true`) and remove the two implicit `Page.bringToFront` calls (tab switch, auto-connect fresh tab). The explicit `bringtofront` command is unchanged, so raising the window stays available as an opt-in.

## Why

A `Target.createTarget` without `background: true` is a foreground create. On a headed Chrome on macOS that raises the whole browser over whatever app the user is in. The daemon does this every time it pins a fresh tab at attach, and it follows with `Page.bringToFront` on every tab switch and on auto-connect.

We attach many daemon sessions to one shared headed Chrome (`--cdp`, `--pin-tab`, one session per coding agent). Measured on macOS (Darwin 25): sampling the frontmost app at 150ms while agents work, Chrome took OS focus every few seconds with no user action. Isolated, both calls steal focus 3 out of 3 times each; the same `createTarget` with `background: true` steals it 0 out of 3 times.

Related: #1247 (background mode request), #1688 (`state save` foregrounds Chrome — its temp target is one of the sites fixed here), #1427 (adds `tab new --background`; this PR instead changes the default, because the daemon's own internal creates are the ones users cannot reach with a flag).

## What

- `CreateTargetParams` gains `background: bool`, set to `true` at all four typed call sites (connect-time create, pin-tab attach, `tab_new`, the `state save` temp target) and the two `json!` sites (browser-context creates).
- `switch_tab` no longer sends `Page.bringToFront`. CDP drives background tabs fine; the switch still re-binds the session and enables domains.
- `open_fresh_tab_for_auto_connect` no longer sends `Page.bringToFront`.

No new flags, no new env vars. `bringtofront` still raises deliberately.

## Test plan

- `cargo test --release --bin agent-browser`: 1099 passed. One failure in our environment, `same_origin_command_post_relays_without_wildcard_cors`, is unrelated: it builds its socket path from `CARGO_MANIFEST_DIR`, and our checkout sits deep enough that the path exceeds the macOS `SUN_LEN` limit. It fails the same way on unpatched source at that path.
- Against a real headed Chrome over `--cdp` on macOS: attach with no binding, `tab new`, and `tab <id>` all complete without Chrome taking OS focus (frontmost app sampled at 150ms throughout); pages load, snapshots and screenshots work from the backgrounded tabs.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -2878,11 +2878,6 @@ async fn open_fresh_tab_for_auto_connect(state: &mut DaemonState) -> Result<(), 
         return Ok(());
     };
     mgr.tab_new(None, None).await?;
-    let session_id = mgr.active_session_id()?.to_string();
-    let _ = mgr
-        .client
-        .send_command("Page.bringToFront", None, Some(&session_id))
-        .await;
     // Best effort: a failed write is retried after the next command.
     let _ = maybe_persist_tab_binding(state);
     Ok(())
@@ -6686,7 +6681,7 @@ async fn handle_recording_start(cmd: &Value, state: &mut DaemonState) -> Result<
             .client
             .send_command_typed(
                 "Target.createTarget",
-                &json!({ "url": "about:blank", "browserContextId": context_id }),
+                &json!({ "url": "about:blank", "browserContextId": context_id, "background": true }),
                 None,
             )
             .await?;
@@ -9416,7 +9411,7 @@ async fn handle_window_new(cmd: &Value, state: &mut DaemonState) -> Result<Value
             .client
             .send_command_typed(
                 "Target.createTarget",
-                &json!({ "url": "about:blank", "browserContextId": context_id }),
+                &json!({ "url": "about:blank", "browserContextId": context_id, "background": true }),
                 None,
             )
             .await?;
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -663,6 +663,7 @@ impl BrowserManager {
                     "Target.createTarget",
                     &CreateTargetParams {
                         url: "about:blank".to_string(),
+                        background: true,
                     },
                     None,
                 )
@@ -1361,6 +1362,7 @@ impl BrowserManager {
                 "Target.createTarget",
                 &CreateTargetParams {
                     url: "about:blank".to_string(),
+                    background: true,
                 },
                 None,
             )
@@ -1526,6 +1528,7 @@ impl BrowserManager {
                 "Target.createTarget",
                 &CreateTargetParams {
                     url: target_url.to_string(),
+                    background: true,
                 },
                 None,
             )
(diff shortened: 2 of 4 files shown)


eyalev:
**Independent confirmation on Linux.** Same failure mode and same fix verified on headed desktop Chrome 146 (Ubuntu/X11), agent-browser 0.34.0, multiple `--session --pin-tab` daemons sharing one authenticated Chrome over `--cdp`:

- `Target.createTarget` **without** `background: true` foregrounded the new tab every run; **with** it, 0 steals (focus checked via `document.hasFocus()`/`visibilityState` across all page targets, plus `xdotool getactivewindow` as OS-level ground truth). The CDP docs mark `background` as Android-only — it demonstrably works on desktop Chrome on both Linux and macOS (this PR's measurements).
- The daemon's connect-time `about:blank` create is indeed one of the steal sites: N sessions attaching over `--cdp` produced exactly N foreground blips and N leaked `about:blank` tabs.
- With 3 pinned sessions running 90 concurrent commands against the shared Chrome, the only steals observed were the create sites this PR patches; everything after creation stayed `visibilityState: "hidden"` with zero cross-session drift.

mikirejf:
Confirmed on macOS 15 (Darwin 24.6) with Chrome for Testing 152, AeroSpace as the window manager, and agent-browser 0.37.1 built from `main` plus this branch.

Before the patch, `tab new` and `tab <label>` pulled Chrome for Testing in front of the current app on every run. I checked the focused app with `aerospace list-windows --focused` after each command. `open`, `snapshot`, `screenshot`, `click`, `get url`, and `tab close` did not raise it. Raw CDP behaved the same way. `Target.createTarget` without `background`, `Target.activateTarget`, and `Page.bringToFront` each raise the app. `createTarget` with `background: true`, followed by navigate and screenshot on that target, does not, and the screenshot of the hidden tab renders correctly.

After the patch, the same sequence never moved focus. 1268 of 1268 unit tests pass.

The branch no longer applies cleanly to `main` (72007a6). `handle_recording_start` was refactored and no longer calls `Target.createTarget`, so that hunk drops out. The other hunks apply unchanged.

AlexHerranr:
Rebased this onto `main` (aff6125) in #1880, keeping the change as yours: the `handle_recording_start` hunk drops out exactly as @mikirejf found, the rest applies unchanged.

I added one thing on top, because the rebase surfaced it: with the implicit raise gone there is no way to raise the window from the CLI. `handle_bringtofront` is in the daemon dispatcher but no command reaches it, so `agent-browser bringtofront` answers "Unknown command". #1880 wires it through the parser, `--help` and the MCP list, which the screenshot path needs: capturing a tab that is not the foreground tab of its window times out in `Page.captureScreenshot`. Measurements on Linux/X11 are in that PR.

Happy to close mine if you would rather push the rebase here.

PR #1698, opened 2026-08-16
fix(cli): errors plain-text output reads the field the producer emits

### LINKED ISSUE #1692: `errors` plain-text output drops every error message
## Summary

`agent-browser errors` prints one `✗` per captured page error, but the message after each marker is empty. The same command with `--json` returns the full error text and stack.

This is separate from #1645: that issue covers clearing the error buffer; this issue covers rendering errors already present in the buffer.

## Reproduction

Reproduced with agent-browser 0.33.2 and 0.34.0 on macOS.

```bash
agent-browser --session errors-output open about:blank
agent-browser --session errors-output eval 'setTimeout(() => { throw new Error("DETACHED-REJECTION"); }, 0); "returned-normally"'
agent-browser --session errors-output wait 200
agent-browser --session errors-output errors
```

Actual output:

```text
✗
```

The JSON output contains the message:

```bash
agent-browser --session errors-output errors --json
```

```json
{"errors":[{"text":"Error: DETACHED-REJECTION\n    at <anonymous>:1:26"}]}
```

Expected plain-text output:

```text
✗ Error: DETACHED-REJECTION
    at <anonymous>:1:26
```

## Root cause

`EventTracker::get_errors_json()` emits each error under the `text` field in `cli/src/native/network.rs`. The plain-text formatter in `cli/src/output.rs` reads `message` and falls back to an empty string:

```rust
let msg = err.get("message").and_then(|v| v.as_str()).unwrap_or("");
```

Reading `text` instead restores the captured message. The same field mismatch is present on current main at `548b159`.

vinhnguyenthanhdn:
Closes #1692.

## The defect

`agent-browser errors` printed the red indicator with an empty message for every error:

```
✗ 
✗ 
```

`--json` was unaffected, which is why the payload looks correct while the default output shows nothing.

The renderer read `message` from each entry:

```rust
let msg = err.get("message").and_then(|v| v.as_str()).unwrap_or("");
```

while `NetworkState::get_errors_json` (`cli/src/native/network.rs`) emits `text`:

```rust
json!({ "text": e.text, "url": e.url, "line": e.line, "column": e.column })
```

The lookup misses and `unwrap_or("")` turns the miss into an empty string, so the failure is silent rather than loud. That is the whole bug.

## The change

Moved the rendering into `format_errors_text`, matching the `format_storage_text` / `format_a11y_text` / `format_vitals_text` helpers already in this module, so the output can be asserted on directly instead of through `println!`. Then read `text`.

Behaviour is otherwise unchanged: an absent or non-array `errors` field still falls through to the next branch, and an empty array still prints nothing and returns.

## Verification

Two tests added. The first one was checked against the old field name before the fix went in — with `message` it fails exactly the way the issue describes:

```
thread 'output::tests::test_format_errors_text_renders_the_message_of_each_error' panicked at src/output.rs:4044:9:
first error message missing from "✗ \n✗ "
```

With `text` it passes. Full local run on macOS 15 (aarch64), `stable` 1.97.1:

```
cargo test --manifest-path cli/Cargo.toml
test result: ok. 1113 passed; 0 failed; 101 ignored

cargo fmt --manifest-path cli/Cargo.toml -- --check     # clean
cargo clippy --manifest-path cli/Cargo.toml -- -D warnings   # clean
```

## Scope

One file. No new dependency, no flag, no change to the JSON surface, so the MCP parity rule in `AGENTS.md` has nothing to sync here — the payload shape was already correct and only the plain-text renderer was reading it wrong. I have left `CHANGELOG.md` alone since `AGENTS.md` says the maintainer owns the changelog voice; happy to add an entry in whatever form you prefer.

### DIFF
--- a/cli/src/output.rs
+++ b/cli/src/output.rs
@@ -99,6 +99,22 @@ fn format_storage_value(value: &serde_json::Value) -> String {
         .unwrap_or_else(|| serde_json::to_string(value).unwrap_or_default())
 }
 
+fn format_errors_text(data: &serde_json::Value) -> Option<String> {
+    let errors = data.get("errors")?.as_array()?;
+
+    let lines = errors
+        .iter()
+        .map(|err| {
+            // The producer emits `text` (see `NetworkState::get_errors_json`); reading
+            // `message` here printed the indicator with an empty message for every error.
+            let msg = err.get("text").and_then(|v| v.as_str()).unwrap_or("");
+            format!("{} {}", color::error_indicator(), msg)
+        })
+        .collect::<Vec<_>>();
+
+    Some(lines.join("\n"))
+}
+
 fn format_storage_text(data: &serde_json::Value) -> Option<String> {
     if let Some(entries) = data.get("data").and_then(|v| v.as_object()) {
         if entries.is_empty() {
@@ -791,10 +807,9 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
             return;
         }
         // Errors
-        if let Some(errors) = data.get("errors").and_then(|v| v.as_array()) {
-            for err in errors {
-                let msg = err.get("message").and_then(|v| v.as_str()).unwrap_or("");
-                println!("{} {}", color::error_indicator(), msg);
+        if let Some(rendered) = format_errors_text(data) {
+            if !rendered.is_empty() {
+                println!("{}", rendered);
             }
             return;
         }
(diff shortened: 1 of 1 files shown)


vinhnguyenthanhdn:
One follow-up after a week: if this small producer/consumer field mismatch is still in scope, I can rebase or respond quickly to any requested change. If the CLI error path has moved since this was opened, I am also happy to close it rather than keep a stale patch around.

PR #1736, opened 2026-08-30
fix(select): match normalized option labels

### LINKED ISSUE #1700: select cannot match native <select> options whose label contains a non-breaking space (U+00A0)
## Summary

`select` cannot drive a native `<select>` whose `<option>` labels contain a
non-breaking space (`&nbsp;` / U+00A0). The exact string an agent reads from
`snapshot -i` cannot be matched back, because the snapshot **strips** U+00A0
from the option's accessible name while `select` compares against the raw
`textContent` (which still contains U+00A0). The two never line up, so the
command fails on the very name it just displayed. On real forms this is common
(province/state pickers with opaque `value` codes), and it silently breaks any
dependent logic — e.g. an AJAX-populated second dropdown that only fires on the
province `change`.

## Environment

- agent-browser 0.34.0

## Reproduction

A province `<select>` with opaque values and `&nbsp;` labels:

```html
<select id="prov">
  <option value="" selected>Provincia</option>
  <option value="B">Buenos&nbsp;Aires</option>
  <option value="C">Capital&nbsp;Federal</option>
</select>
```

In `snapshot -i` the option's accessible name is rendered as **`CapitalFederal`**
— the `&nbsp;` is dropped, so the two words are joined. An agent naturally
copies that string and runs:

```
agent-browser select @<ref> "CapitalFederal"
```

Result, for the various strings a caller might reasonably try:

| `select @<ref> "X"` with X = | option selected? |
|---|---|
| `CapitalFederal` (exactly what the snapshot shows) | **no** |
| `Capital Federal` (normal, single ASCII space) | **no** |
| `Capital\u00A0Federal` (exact label, with nbsp) | yes |
| `C` (the opaque `value`) | yes |

So the string the tree displays, and the human-obvious `Capital Federal`, are
both unmatchable; only the raw nbsp label or the opaque code work — neither of
which the agent can see in the snapshot.

## Root cause

Two individually reasonable choices combine into a mismatch:

1. **Snapshot rendering strips U+00A0 from names.** `cli/src/native/snapshot.rs`
   lists U+00A0 in `INVISIBLE_CHARS` and renders names via
   `display_name.replace(INVISIBLE_CHARS, "")`, so `Capital\u00A0Federal` is
   shown as `CapitalFederal`.
2. **`select` matching is whitespace-sensitive.**
   `cli/src/native/interaction.rs::select_option` matches
   `vals.includes(opt.value) || vals.includes(opt.textContent.trim())`.
   `textContent` still contains the U+00A0, so neither the displayed
   `CapitalFederal` nor a normal-space `Capital Federal` matches.

(The loud `No option matched ...; available: [...]` error added in #1432 is a
real improvement over the earlier silent no-op, but it doesn't help here: the
option shown in the tree still isn't selectable, so the agent just gets a hard
error for the name it was given.)

## Proposed fix

Make option matching whitespace-insensitive and accessible-name aware: match
each option's `value`, `label` and `textContent` using both a
whitespace-collapsed form (any run of whitespace, **including** U+00A0, → a
single space) and an invisible-stripped form (mirrors what the snapshot
renders), with a case-insensitive pass only as a last resort. Keep the loud
error when a value genuinely matches nothing. This makes the exact string shown
in `snapshot -i` selectable on any site, without changing the command surface,
and both the CLI and MCP `select` paths inherit it (they share
`select_option`).

I have a PR ready that implements this plus an e2e test (select-by-displayed
name, select-by-normal-space label, select-by-value, and the loud failure on a
genuinely unmatched value).

Railly:
Fixes #1700.

Recreates and tightens the solution from #1701 by @mauriantolin.

## What changed

- Render non-breaking spaces as ordinary spaces in snapshots instead of joining words.
- Match native select options by exact case-sensitive value, label, or text first.
- Fall back only to whitespace-normalized visible labels.
- Reject normalized ambiguity without changing browser state.
- Keep total and partial multi-value misses non-destructive.
- Preserve multi-select support and the existing bubbling `change` event.
- Align CLI help, README, docs, and the packaged core skill.

Unlike #1701, this does not add case-insensitive matching and does not silently select multiple options that collide after normalization.

## Verification

- `cargo test --profile ci --manifest-path cli/Cargo.toml`: 1,156 passed, 106 ignored
- Full native Chrome E2E suite: 105 passed
- Focused E2E repeated four times
- Force-red mutations for snapshot NBSP handling, normalized selection, and non-destructive misses
- Version sync and launcher tests passed

Co-authored-by: Mauricio Antolin <suscripciones@mauricioantolin.com>

### DIFF
--- a/cli/src/native/interaction.rs
+++ b/cli/src/native/interaction.rs
@@ -442,18 +442,41 @@ pub async fn select_option(
     // selects a misspelled option otherwise sees "Done", and only discovers
     // the page state is wrong after more commands. List what was available.
     let js = r#"function(vals) {
+            const normalize = (value) => String(value ?? '')
+                .replace(/[\u200B\u200C\u200D\u2060\uFEFF]/g, '')
+                .replace(/\s+/g, ' ')
+                .trim();
             const options = Array.from(this.options);
-            let matched = 0;
-            for (const opt of options) {
-                opt.selected = vals.includes(opt.value) || vals.includes(opt.textContent.trim());
-                if (opt.selected) matched += 1;
+            const wanted = new Set();
+            for (const value of vals) {
+                let matches = options.filter((opt) =>
+                    value === opt.value ||
+                    value === opt.label.trim() ||
+                    value === opt.textContent.trim()
+                );
+                if (matches.length === 0) {
+                    const normalizedValue = normalize(value);
+                    matches = options.filter((opt) =>
+                        normalize(opt.label) === normalizedValue ||
+                        normalize(opt.textContent) === normalizedValue
+                    );
+                    if (matches.length > 1) {
+                        return { error: 'Multiple options matched ' + JSON.stringify(value) + ' after whitespace normalization' };
+                    }
+                }
+                if (matches.length === 0) {
+                    const available = options.map(o => o.value + ' ("' + normalize(o.label) + '")').join(', ');
+                    return { error: 'No option matched ' + JSON.stringify(vals) + '. Available options: ' + available };
+                }
+                for (const opt of matches) wanted.add(opt);
             }
-            if (matched === 0) {
-                const available = options.map(o => o.value + ' ("' + o.textContent.trim() + '")').join(', ');
+            if (wanted.size === 0) {
+                const available = options.map(o => o.value + ' ("' + normalize(o.label) + '")').join(', ');
                 return { error: 'No option matched ' + JSON.stringify(vals) + '. Available options: ' + available };
             }
+            for (const opt of options) opt.selected = wanted.has(opt);
             this.dispatchEvent(new Event('change', { bubbles: true }));
-            return { matched };
+            return { matched: wanted.size };
         }"#
     .to_string();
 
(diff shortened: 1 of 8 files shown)


cli/src/native/interaction.rs
```diff
@@ -442,18 +442,41 @@ pub async fn select_option(
...
+                let matches = options.filter((opt) =>
+                    value === opt.value ||
+                    value === opt.label.trim() ||
+                    value === opt.textContent.trim()
+                );
+                if (matches.length === 0) {
+                    const normalizedValue = normalize(value);
+                    matches = options.filter((opt) =>
+                        normalize(opt.label) === normalizedValue ||
+                        normalize(opt.textContent) === normalizedValue
```
ctate: The fallback also checks `textContent` when a `label` attribute overrides it, so hidden text from another option can make the snapshot's displayed label impossible to select. Please use the normalized `opt.label` for fallback matching; it already falls back to option text when no label attribute exists.

ctate (changes requested):
.

ctate (approved):

PR #1739, opened 2026-08-30
fix(cdp): harden transport parsing and shutdown

### LINKED ISSUE #1666: snapshot hangs forever when the AX tree contains a lone UTF-16 surrogate (flag emoji): CDP response silently dropped in reader loop
**Version**: agent-browser 0.33.2 (homebrew, darwin-arm64), managed Chrome for Testing 150.0.7871.24, headless=new, macOS 15.5 (arm64). Also reproduces on 0.31.1.

## Summary

Any page whose accessibility tree contains a **lone UTF-16 surrogate** in an `InlineTextBox` name makes `snapshot` hang forever. `serde_json` cannot parse the CDP response, and the reader loop at `cli/src/native/cdp/client.rs:168-173` silently `continue`s past it, so the caller's oneshot receiver never resolves.

The single most common real-world trigger is a **flag emoji** (🇺🇸, 🇬🇧, …) — every country-picker component built on regional indicator pairs is affected. In our case it was a phone country selector inside a modal; every dialog containing it became un-snapshottable.

## Root cause

A flag emoji is two code points (`U+1F1FA U+1F1F8`) = four UTF-16 units (`D83C DDFA D83C DDF8`). Chrome's `InlineTextBox` layout splits text by UTF-16 unit and can cut this sequence in half, emitting an AX node whose entire name is one unpaired high surrogate:

```json
{"nodeId":"-1000000831","ignored":false,
 "role":{"type":"internalRole","value":"InlineTextBox"},
 "name":{"type":"computedString","value":"\ud83c"},
 "properties":[],"parentId":"2785","childIds":[]}
```

`"\ud83c"` is valid JSON per RFC 8259 (which does not require well-formed UTF-16), and Chrome emits it happily. But Rust's `String` is UTF-8 and cannot represent an unpaired surrogate, so `serde_json` rejects it:

```rust
let msg = r#"{"id":1,"result":{"nodes":[{"name":{"value":"\ud83c"}}]}}"#;
serde_json::from_str::<serde_json::Value>(msg)
// Err: unexpected end of hex escape at line 1 column 52
```

`cli/src/native/cdp/client.rs:168-173`:

```rust
let parsed: CdpMessage = match serde_json::from_str(&msg) {
    Ok(m) => m,
    // Expected for inspect proxy messages with negative IDs
    // (CdpMessage.id is u64); handled via raw broadcast above.
    Err(_) => continue,      // <-- the getFullAXTree response is dropped here
};
```

The `continue` was written for inspect-proxy messages with negative ids, but it swallows **every** unparseable frame — including command responses. `pending` still holds the entry, the oneshot sender is never fired, and `send_command_typed` waits forever. The 30s per-command timeout in `send_command` does not help because the whole session daemon is now stuck behind it, and the CLI's own timeout never fires either — the process must be SIGKILLed externally (`pkill -f` is not enough; `ps | awk '{print $2}' | xargs kill -9` is required).

This is **not** #1660: the renderer is perfectly healthy here. Raw CDP against the exact same page returns in **43 ms**.

## Reproduction

Chrome itself is fine — only agent-browser hangs. Side-by-side on the identical live page and identical open dialog:

| client | result |
|---|---|
| raw CDP `Accessibility.getFullAXTree` over the DevTools WebSocket | **OK in 43 ms**, 1946 nodes, 753 KB |
| `agent-browser snapshot -i -d 2` | **hangs indefinitely** (>25 s, killed externally) |

The AX tree is structurally sound — I walked all 1946 nodes: `cycles: 0, orphans: 0`. Exactly **one** node contains the lone surrogate: the `InlineTextBox` shown above.

Evidence that the surrogate is the trigger, on the live page, fully reversible:

```js
// hangs
document.querySelector('#react-select-2-single-value').innerText  // "🇺🇸"

// replace the emoji with ASCII, change nothing else -> snapshot returns in 0.07s
document.querySelector('#react-select-2-single-value span').textContent = 'US'

// put the flag back -> hangs again
document.querySelector('#react-select-2-single-value span').textContent = '\u{1F1FA}\u{1F1F8}'
```

Ruled out by experiment: DOM size (a 758-node page hangs while a 1013-node one is fine), `aria-hidden` count, Atlaskit modal focus-trap / `data-focus-guard`, font family, and Google Places widgets. Adding `aria-hidden="true"` to the emoji node also fixes it — it keeps the node out of the AX tree entirely.

Note that a static HTML file containing the same markup does **not** reproduce: whether Chrome splits the `InlineTextBox` mid-surrogate depends on actual layout/line-breaking, so it needs a real page. This is why the bug looks intermittent and content-dependent.

## Impact

- Silently drops **any** CDP response Rust cannot parse, not just this one. Any command can hang forever with no error.
- Flag emoji are extremely common in country/phone/language pickers (`react-select` + regional indicators is a standard pattern), so this hits a lot of real apps.
- User-generated content with emoji (names, comments, chat) can trigger it anywhere, unpredictably.
- Recovery requires an external SIGKILL. After that the session's active tab often ends up on `about:blank` and `close --all` itself times out — the dead CDP connection is never cleaned up.

## Suggested fix

Two independent changes; the first alone fixes this bug, the second prevents the whole class:

**1. Decode lone surrogates leniently.** Sanitize unpaired surrogate escapes to `U+FFFD` before `from_str`, or parse with a lenient decoder. A replacement char in an `InlineTextBox` name is harmless — these nodes are layout artifacts and are not user-facing refs.

**2. Never silently drop a frame that carries an `id`.** Before `continue`, do a cheap `Value` probe for an `id` field; if present, resolve that pending entry with an error so the caller fails fast instead of hanging:

```rust
let parsed: CdpMessage = match serde_json::from_str(&msg) {
    Ok(m) => m,
    Err(e) => {
        // Don't strand a caller: if this frame carries a command id,
        // fail that request instead of dropping it silently.
        if let Some(id) = extract_id_loosely(&msg) {
            if let Some(tx) = pending_clone.lock().await.remove(&id) {
                let _ = tx.send(CdpMessage::parse_error(id, e.to_string()));
            }
        }
        continue;
    }
};
```

Happy to send a PR for either or both.

### LINKED ISSUE #1726: External CDP connections are never closed, so every liveness-probe redial strands a live socket
## Summary

On an external CDP connection (`--cdp`, `--auto-connect`), the daemon never closes its WebSocket. Because `close` also sits on the *reconnect* path, a slow or briefly stalled endpoint makes the daemon strand one live, still-subscribed socket per redial, for the life of the daemon.

## Where

`BrowserManager::close` (`cli/src/native/browser.rs`) sends `Browser.close` only when the daemon launched the browser itself, which is correct. For an external connection it then falls through and returns `Ok(())` without touching the socket:

```rust
pub async fn close(&mut self) -> Result<(), String> {
    if self.browser_process.is_some() {
        // Only send Browser.close when we launched the browser ourselves.
        // For external connections (--auto-connect, --cdp) we just disconnect
        // without shutting down the user's browser.
        let _ = self.client.send_command_no_params("Browser.close", None).await;
    }
    if let Some(mut process) = self.browser_process.take() { /* ... */ }
    Ok(())
}
```

The comment says "we just disconnect", but nothing disconnects. Dropping the `CdpClient` does not do it either: `_reader_handle` and `_keepalive_handle` are detached `JoinHandle`s, so dropping the handles leaves both tasks running, and the reader task owns the read half of the split stream. The TCP connection therefore stays open, still subscribed to every event, and the keepalive task keeps pinging it. `CdpClient` has no `Drop` impl and no `close`/`disconnect` method.

Against a real Chrome this is invisible: the only disconnect path is `Browser.close`, and Chrome then closes the socket from its side.

## Why it compounds

`is_connection_alive` probes `Browser.getVersion` with a hardcoded 3s timeout before every command (`handle_command` in `cli/src/native/actions.rs`). A failed probe sets `needs_launch`, which runs `close_current_browser` and then `auto_launch` — a fresh socket. So every timed-out probe leaks one connection.

Against a CDP multiplexer that fans one browser out to several clients, each stranded socket receives the full event stream again over the same link, which makes the next probe likelier to time out, which strands another socket. We hit that spiral against our own CDP proxy in production: sockets accumulated every one to three minutes over seventeen minutes until every id-routed round trip exceeded its client deadline, and the browser lane stopped answering while plain HTTP on the same host stayed healthy.

## Reproduction

A mock CDP endpoint that answers normally, then stalls `Browser.getVersion` past 3s:

```
$ agent-browser --cdp ws://127.0.0.1:PORT/devtools/browser/mock get url
about:blank
# flip the mock into stall mode
$ agent-browser --cdp ws://127.0.0.1:PORT/devtools/browser/mock get url
[agent-browser] relaunched browser
```

Server-side socket census, `agent-browser 0.34.0`, `bin/agent-browser-linux-x64` as shipped on npm (sha256 `69eadf5d8d6003a06a5cd2f914ebb261c7754fe1335a9190122c334e91909789`):

```
sockets opened: 3
  #1 open=True  commands=12
  #2 open=True  commands=16
  #3 open=True  commands=16
```

One stalled probe leaves three live connections. The same census after the fix in the PR below leaves only the newest:

```
sockets opened: 3
  #1 open=False commands=12
  #2 open=False commands=16
  #3 open=True  commands=16
```

Same behaviour on `main` (0.35.0): `close` and the `CdpClient` internals are unchanged there.

## Fix

PR: give `CdpClient` a `close` that sends a Close frame and stops both background tasks, and have `BrowserManager::close` call it on every path.

Railly:
## Summary

- repair lone UTF-16 surrogate escapes before typed CDP parsing while preserving valid pairs and ordinary Unicode
- fail only the matching positive top-level pending command when a malformed response still cannot be parsed
- make CDP shutdown bounded and idempotent, release pending callers, reject post-close writes, and force TCP EOF through shared plain or Rustls transports
- disconnect external CDP transports without sending `Browser.close` to browsers agent-browser does not own

Closes #1666.
Closes #1726.

## Testing

- `cargo test --locked --manifest-path cli/Cargo.toml`: 1,164 unit tests and 2 integration tests passed
- 8 focused CDP tests passed five consecutive runs
- `e2e_launch_navigate_evaluate_close` passed against real Chrome
- fix-absent mutations proved lone-surrogate repair and `shutdown(Both)` are load-bearing
- Microsoft Windows Server 2025 with Rust 1.98.0: all 8 CDP tests and `test_close_external_manager_disconnects_without_closing_browser` passed through a cuse scenario

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -1286,6 +1286,8 @@ impl BrowserManager {
                 .await;
         }
 
+        self.client.close().await;
+
         if let Some(mut process) = self.browser_process.take() {
             let timeout = std::time::Duration::from_secs(5);
             let _ = tokio::task::spawn_blocking(move || {
@@ -3012,6 +3014,64 @@ mod tests {
             bound_target_gone: None,
             headless: true,
         }
+    }
+
+    #[tokio::test]
+    async fn test_close_external_manager_disconnects_without_closing_browser() {
+        use futures_util::StreamExt;
+        use tokio::io::AsyncReadExt;
+        use tokio_tungstenite::tungstenite::Message;
+
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+        let server = tokio::spawn(async move {
+            let (stream, _) = listener.accept().await.unwrap();
+            let mut ws = tokio_tungstenite::accept_async(stream).await.unwrap();
+            let mut browser_close_seen = false;
+
+            loop {
+                match ws.next().await {
+                    Some(Ok(Message::Text(text))) => {
+                        let command: Value = serde_json::from_str(&text).unwrap();
+                        browser_close_seen |= command["method"] == "Browser.close";
+                    }
+                    Some(Ok(Message::Close(_))) => break,
+                    Some(Ok(_)) => {}
+                    Some(Err(_)) | None => break,
+                }
+            }
+
+            let mut byte = [0u8; 1];
+            let count = tokio::time::timeout(Duration::from_secs(1), ws.get_mut().read(&mut byte))
+                .await
+                .expect("external manager close should shut down TCP")
+                .expect("external manager TCP read should succeed");
+            (browser_close_seen, count)
+        });
+
+        let client = CdpClient::connect(&format!("ws://{}", addr)).await.unwrap();
+        let mut manager = BrowserManager {
+            client: Arc::new(client),
+            browser_process: None,
+            ws_url: format!("ws://{}", addr),
+            pages: Vec::new(),
+            active_page_index: 0,
+            default_timeout_ms: 25_000,
+            download_path: None,
+            ignore_https_errors: false,
+            visited_origins: HashSet::new(),
+            next_tab_id: 1,
+            direct_page: false,
+            pin_tab: false,
+            bound_target_id: None,
+            bound_target_gone: None,
+            headless: true,
+        };
+        manager.close().await.unwrap();
+
+        let (browser_close_seen, count) = server.await.unwrap();
+        assert!(!browser_close_seen);
+        assert_eq!(count, 0);
     }
 
     const TARGET_A: &str = "AAAA0000BBBB1111CCCC2222DDDD3333";
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -447,6 +447,7 @@ fn tab_gone_error(target_id: &str, last_url: &str) -> String {
 const LIGHTPANDA_CDP_CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
 const LIGHTPANDA_CDP_CONNECT_POLL_INTERVAL: Duration = Duration::from_millis(100);
 const LIGHTPANDA_TARGET_INIT_TIMEOUT: Duration = Duration::from_secs(10);
+const FAILED_INITIALIZATION_CLOSE_TIMEOUT: Duration = Duration::from_secs(1);
 
 impl BrowserManager {
     /// True when a *default* idle timeout must not close this browser:
@@ -517,7 +518,7 @@ impl BrowserManager {
             }
         };
 
-        let manager = if engine == "lightpanda" {
+        let mut manager = if engine == "lightpanda" {
             initialize_lightpanda_manager(ws_url, process).await?
         } else {
             let client = Arc::new(CdpClient::connect(&ws_url).await?);
@@ -538,11 +539,24 @@ impl BrowserManager {
                 bound_target_gone: None,
                 headless,
             };
-            manager.discover_and_attach_targets().await?;
+            if let Err(error) = manager.discover_and_attach_targets().await {
+                let _ = manager
+                    .close_with_timeout(Some(FAILED_INITIALIZATION_CLOSE_TIMEOUT))
+                    .await;
+                return Err(error);
+            }
             manager
         };
 
-        let session_id = manager.active_session_id()?.to_string();
+        let session_id = match manager.active_session_id() {
+            Ok(session_id) => session_id.to_string(),
+            Err(error) => {
+                let _ = manager
+                    .close_with_timeout(Some(FAILED_INITIALIZATION_CLOSE_TIMEOUT))
+                    .await;
+                return Err(error);
+            }
+        };
 
         if ignore_https_errors {
             let _ = manager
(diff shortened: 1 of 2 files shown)


cli/src/native/browser.rs
```diff
@@ -1286,6 +1286,8 @@ impl BrowserManager {
                 .await;
         }
 
+        self.client.close().await;
```
ctate, line 1326: Please also route failed manager initialization through this cleanup. After opening a client, target discovery and domain setup can return an error before the manager is handed to the daemon, so this close call is never reached.

ctate (approved):

PR #1760, opened 2026-09-01
feat: advertise WebMCP tools after navigation

Railly:
## Summary

- advertise available WebMCP tools after `open <url>` and `navigate`
- include `data.webmcp.toolCount` in JSON output
- respect allowed domains and `--no-webmcp`

## Example

```console
$ agent-browser open https://example.com
✓ Example
  https://example.com/
WebMCP tools are available on this page (experimental)
Run `agent-browser webmcp list` to view them
```

Validated with the real Chrome WebMCP fixture.

## Tests

- `cargo fmt -- --check`
- `cargo clippy`
- `cargo test`
- `cargo test e2e -- --ignored --test-threads=1`


cli/src/native/actions.rs
```diff
@@ -11914,6 +11929,58 @@ fn success_response(id: &str, data: Value) -> Value {
...
+        .as_ref()
+        .and_then(|browser| browser.active_session_id().ok())
+    else {
+        return;
+    };
+    let Ok(tools) = state.webmcp.tools(session_id) else {
+        return;
+    };
+    let domain_filter = state.domain_filter.read().await;
+    attach_webmcp_availability_from_tools(resp, action, &tools, domain_filter.as_ref());
```
ctate: Could navigation wait briefly for the page's initial WebMCP registrations, or emit an availability update when a tool is added after navigation?

ctate (approved):

ShacktarMalacky (approved):

ShacktarMalacky (approved):

PR #1770, opened 2026-09-03
Support reliable selection for native and ARIA dropdowns

### LINKED ISSUE #1105: bug: agent-browser select fails on custom ARIA combobox widgets
## Problem

  `agent-browser select` on custom ARIA combobox widgets (Radix UI, Headless UI, etc.)
  fails with
  `Element not found`, while native `<select>` elements work correctly.

  ## Steps to Reproduce

  **Custom ARIA combobox — fails:**
  ```bash
  agent-browser open "https://ybqcgqwz8b.coze.site"
  agent-browser snapshot
  # → combobox "Position *" [ref=e3]: Select Position

  agent-browser select e3 "Product Manager"
  # → ✗ Element "e3" not found or not visible.
  ```

  **Native `<select>` — works:**
  ```bash
  agent-browser open "https://w7nzypr53r.coze.site"
  agent-browser snapshot
  # → combobox "Target Position" [ref=e2]
  #     - option "Frontend Developer" [ref=e4]
  #     - option "Backend Developer" [ref=e5]
  #     ...

  agent-browser select e2 "Backend Developer"
  # → ✓ Done
  ```

  ## Root Cause

  Chrome's accessibility tree reports `role="combobox"` for **both** native
  `<select>` elements
  and custom ARIA widgets. The existing `select` implementation only handles
  native `<select>`,
  which exposes its options directly in the AX tree. Custom combobox widgets
  (Radix UI, etc.)
  render options dynamically — only after the trigger is clicked — so
  `Element not found` is
  returned before the dropdown is ever opened.

  ## Expected Behavior

 agent-browser select should support custom ARIA combobox widgets（widely used in modern frontend frameworks） the same way it      
  supports native <select> elements — selecting the specified option by text. 

  Labels: bug

ctate:
## Summary

- Extend `select` to support native `<select>` elements and standard ARIA `combobox` and `listbox` controls.
- Match options by exact value, label, or whitespace-normalized accessible text, then use trusted browser input, dynamic option polling, and state verification to complete custom-widget selections reliably.
- Preserve native multi-select behavior, define final-set semantics for explicitly multiselectable ARIA listboxes, handle active-descendant Enter activation and dialogs, and return actionable errors for unsupported or unverifiable controls.
- Keep the CLI, MCP, Eve integration, documentation, and agent guidance aligned with the expanded selection behavior, with regression coverage for native, ARIA, filtering, dynamic, and dialog-driven controls.

Closes #1105


Railly (changes requested):
- **False success**: selection verification accepts highlighted but uncommitted ARIA options and detached native selects after rerender, while the live value remains unchanged.
- **Unbounded retries**: active-descendant activation repeatedly presses Enter beyond the configured timeout.
- **Valid selections fail**: replacing an existing filter on macOS leaves stale text (Red → ReBlue), and value-to-label selections (ca → Canada) time out when the popup unmounts.

PR #1778, opened 2026-09-03
[cli] Record: require ffmpeg up front, validate output extension

### LINKED ISSUE #1772: record: ffmpeg dependency is undocumented, unchecked by doctor, and output extension is not validated
\`agent-browser record start\` hard-depends on \`ffmpeg\` on PATH for every output format (\`cli/src/native/recording.rs\`, \`build_ffmpeg_command\` spawns \`ffmpeg\` and pipes JPEG frames into it; \`.webm\` picks \`libvpx\`, anything else picks \`libx264\`). Nothing tells the user this up front, and a bad output path fails late and opaquely.

Concretely:

1. The ffmpeg requirement is not documented anywhere. \`grep -rin ffmpeg docs/src/app/recording/page.mdx skill-data/core/references/video-recording.md skill-data/core/SKILL.md README.md\` returns nothing. The only mention is the runtime error string.
2. \`agent-browser doctor\` has no ffmpeg probe (\`grep -ri ffmpeg cli/src/doctor\` is empty), so a fresh machine finds out at the first \`record start\`.
3. The output extension is never validated. \`.webm\` is the only documented format; \`.mp4\` works via libx264 but is undocumented. A path with no recognised extension makes ffmpeg fail because it cannot pick a muxer, and the error only surfaces at \`record stop\`, truncated to the first 300 chars of stderr (\`recording.rs:394\`), which is ffmpeg's version banner rather than the actual cause. An ffmpeg build without libvpx or libx264 fails the same late, opaque way. By then the take is already lost.

## Repro

\`\`\`sh
# fresh machine, no ffmpeg
agent-browser record start ./take.webm
# Error: ffmpeg not found or failed to execute: No such file or directory (os error 2). Install ffmpeg to enable recording.

# with ffmpeg installed, no extension
agent-browser record start ./take
agent-browser record stop
# Error: ffmpeg failed: ffmpeg version 7.x Copyright (c) 2000-2025 the FFmpeg developers
#   built with ... (300-char head of the banner; the real "Unable to find a suitable output format" line is cut off)
\`\`\`

## Proposed fix

- (a) Document the ffmpeg requirement (\`brew install ffmpeg\`, \`apt install ffmpeg\`) in \`docs/src/app/recording/page.mdx\`, \`skill-data/core/references/video-recording.md\` / \`SKILL.md\`, and \`README.md\`.
- (b) Add a \`doctor\` probe: ffmpeg on PATH, and via \`ffmpeg -encoders\` whether \`libvpx\` and \`libx264\` are compiled in.
- (c) Validate the output extension in the CLI parser up front (same place \`--fps\` is validated in \`cli/src/commands.rs\` on #1763), reject unknown extensions with a clear message, and document that \`.mp4\` is supported.
- (d) When ffmpeg fails, surface the tail of stderr instead of the head; the useful line is always last.

Related: #1763 (\`--fps\` on \`record start\`) is in flight and touches the same \`record\` parser. This should land after it or be stacked on the \`recording-fps-30-60\` branch to avoid conflicts.

jamesvclements:
`record` needs ffmpeg for every output format, but nothing said so until the first take failed, and a path with no extension only failed at `record stop` with the first 300 chars of stderr, which is the version banner. This makes the requirement visible up front and the failure legible. Builds on #1763.

## Before / after

`record start` with an extensionless path, `main` vs this branch:

```
$ agent-browser record start /tmp/take
✓ Recording started: /tmp/take (30 fps)
$ agent-browser record stop
✗ ffmpeg failed: ffmpeg version 8.1 Copyright (c) 2000-2026 the FFmpeg developers
  built with Apple clang version 21.0.0 (clang-2100.0.123.102)
  ...
```

```
$ agent-browser record start /tmp/take
Invalid output path: '/tmp/take' has no extension (use .webm or .mp4)
Usage: agent-browser record start <output.webm|output.mp4> [url] [--fps <n>]
```

`doctor` has no Recording section on `main`. On this branch, with ffmpeg installed and then with it off PATH:

```
Recording
  pass  ffmpeg version 8.1 at /opt/homebrew/bin/ffmpeg
  pass  libvpx (.webm) and libx264 (.mp4) encoders available
```

```
Recording
  warn  ffmpeg not found on PATH (only needed for `record`)
        fix: brew install ffmpeg
```

## Details

**Extension check.** The parser rejects only paths whose file name has no extension (`take`, `dir.v2/take`, `.hidden`), before the daemon is contacted. That's the one case that was already broken: ffmpeg picks the muxer from the extension and had nothing to go on. Anything with an extension is handed to ffmpeg exactly as before, so `.mkv`, `.mov` and `.avi` outputs that worked pre-PR keep working (H.264 in that container). `.webm` and `.mp4` are documented as the supported formats because those are the two the encoder settings are tuned for. `recording_start` in the daemon runs the same check so raw socket callers get the same error, and `record restart` checks the new path before stopping the in-flight take. `build_ffmpeg_command` now matches `.webm` case-insensitively: on `main` an uppercase `.WEBM` path picked libx264, which the WebM muxer rejects, so the take failed at stop with the same banner.

**stderr tail.** ffmpeg prints the cause last, so the error keeps the last 400 characters, cut at a line boundary, instead of the first 300. `-hide_banner` keeps the banner out of stderr entirely. The spawn failure message ("ffmpeg not found ... Install ffmpeg") is unchanged.

**doctor.** New `Recording` section with `recording.ffmpeg` (PATH lookup and version line) and `recording.ffmpeg_encoders` (`ffmpeg -hide_banner -encoders` parsed for `libvpx` and `libx264`). Everything is Pass or Warn, never Fail, because recording is the only feature that needs ffmpeg and a machine without it is otherwise healthy; a Fail would flip doctor's exit code for people who never record. Fix hints are `brew install ffmpeg` on macOS, `apt install ffmpeg` on Linux, and say it's only needed for `record`. The probe is split into process calls and a pure `push_checks`, so the tests cover every probe outcome without an ffmpeg install, plus `/bin/sh` scripted runs for the command wrapper (same style as `chrome.rs`).

**Docs.** Requirements section on the recording docs page and the skill reference, one line each in `SKILL.md`, `commands.md`, README, `record --help`, and the MCP `path` schema descriptions for `record_start`/`record_restart`. They name `.webm` (VP8) and `.mp4` (H.264) as the supported formats and say other extensions go to ffmpeg as-is with H.264 video.

Verified on macOS with ffmpeg 8.1 (Homebrew, libvpx + libx264): `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (1257 passed), `cargo test doctor` (34 passed), and `cargo test e2e_recording -- --ignored --test-threads=1` (3 passed) all clean. With release builds of `main` and this branch side by side: `record start /tmp/take` and `/tmp/dir.v2/take` exit 1 from the parser with the message above, where `main` started and then printed the banner at stop; `.avi`, `.MKV`, `.WEBM` and `.mp4` takes all save on this branch and ffprobe shows `h264`/`avi`, `h264`/`matroska,webm`, `vp8`/`matroska,webm` and `h264`/`mov,mp4`, where `main` failed the `.WEBM` take with the banner; `record --help` mentions the ffmpeg requirement; `doctor` shows the two passes above, the warn with `fix: brew install ffmpeg` under `PATH=/usr/bin:/bin` (exit 0), and `pass ... (version unknown)` plus `warn Could not list ffmpeg encoders` against a stub ffmpeg; a stub ffmpeg that prints a 600-char banner and exits 1 yields `ffmpeg failed: [out#0/webm @ 0x600] Unknown encoder 'libvpx' / Error opening output file /tmp/x.webm.` instead of the banner. Not re-run in this pass: raw socket `recording_start`/`recording_restart` with an extensionless path.

Closes #1772

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,7 @@ agent-browser trace start             # Start recording trace
 agent-browser trace stop [path]       # Stop and save trace
 agent-browser profiler start          # Start Chrome DevTools profiling
 agent-browser profiler stop [path]    # Stop and save profile (.json)
-agent-browser record start ./demo.webm           # Start video recording at 30 fps
+agent-browser record start ./demo.webm           # Start video recording at 30 fps (.webm or .mp4; needs ffmpeg on PATH)
 agent-browser record start ./demo.webm --fps 60  # 60 fps for motion-heavy takes (1-60 allowed)
 agent-browser record stop                        # Stop and save the video
 agent-browser record restart ./take2.webm        # Stop the current recording, start a new one
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -1677,7 +1677,7 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
                     "recording_start",
                     &rest[1..],
                     "record start",
-                    "record start <output.webm> [url] [--fps <n>]",
+                    "record start <output.webm|output.mp4> [url] [--fps <n>]",
                 ),
                 Some("stop") => Ok(json!({ "id": id, "action": "recording_stop" })),
                 Some("restart") => parse_record_take(
@@ -1685,7 +1685,7 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
                     "recording_restart",
                     &rest[1..],
                     "record restart",
-                    "record restart <output.webm> [url] [--fps <n>]",
+                    "record restart <output.webm|output.mp4> [url] [--fps <n>]",
                 ),
                 Some(sub) => Err(ParseError::UnknownSubcommand {
                     subcommand: sub.to_string(),
@@ -2324,7 +2324,9 @@ fn parse_read(rest: &[&str], id: &str, flags: &Flags) -> Result<Value, ParseErro
 /// Parse the arguments shared by `record start` and `record restart`:
 /// `<path> [url] [--fps <n>]`.
 ///
-/// `rest` excludes the subcommand. Recording defaults to
+/// `rest` excludes the subcommand. `path` needs an extension so ffmpeg can
+/// pick a container (`.webm` and `.mp4` are the tuned ones). Recording
+/// defaults to
 /// `recording::DEFAULT_FPS`; `--fps` accepts anything up to
 /// `recording::MAX_FPS`, with 60 reserved for motion-heavy takes.
 fn parse_record_take(
@@ -2391,6 +2393,11 @@ fn parse_record_take(
         context: context.to_string(),
         usage,
     })?;
+
+    // ffmpeg picks the container from the extension and only fails at
+    // `record stop`, so reject extensionless paths before the daemon is asked.
+    crate::native::recording::validate_output_path(path)
+        .map_err(|message| ParseError::InvalidValue { message, usage })?;
 
     let mut cmd = json!({ "id": id, "action": action, "path": path });
     if let Some(u) = url {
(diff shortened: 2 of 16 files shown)


Railly:
**cli/src/native/recording.rs:356-367**

Could record start synchronously verify that ffmpeg can be launched and roll back the active state when that probe fails?

Railly:
**cli/src/native/actions.rs:7130**

Can output-path validation move before creating or activating the recording context?

jamesvclements:
Both done in fe04937.

> Could record start synchronously verify that ffmpeg can be launched and roll back the active state when that probe fails?

ffmpeg is now spawned in `start_recording_task` before the capture session is attached, rather than inside the background task. If the spawn fails, the state is rolled back and `record start` returns the "ffmpeg not found ... Install ffmpeg" error itself; if attaching the capture session fails afterwards, ffmpeg is killed and its empty output file removed. The task receives the running child. `e2e_recording_fails_fast_without_ffmpeg` runs `recording_start` with an empty `PATH`, asserts the failure names ffmpeg and leaves the state inactive, then restores `PATH` and asserts the next start records normally.

> Can output-path validation move before creating or activating the recording context?

Yes, it now sits next to the fps check at the top of `handle_recording_start` (and `handle_video_start`), before the context, tab, or navigation. `e2e_recording_rejects_extensionless_path_before_context` asserts the rejection leaves one tab and no file.

1258 unit tests, 4 recording e2e tests green.

Issue #1713, opened 2026-08-23
No per-command deadline: a wedged session target hangs snapshot/read/open indefinitely and harness-side timeouts cannot reap the detached daemon + Chrome tree


mecworks:
## Summary

Once a session's browser target wedges (page that never finishes rendering, a frozen tab, or a dead target), **a single client command blocks the daemon socket indefinitely** — there is no deadline anywhere on the client↔daemon round trip. Commands such as `snapshot`, `read` (active tab), and even `open` hang for minutes to tens of minutes; the only recovery is killing the detached daemon process tree from outside the CLI.

This makes agent-browser unusable inside agent harnesses (Chatbox, MCP wrappers, editors) that run commands under a sandbox timeout: the harness kills its own child (the CLI process) after N seconds, but the **detached daemon and its Chrome tree survive**, still hold the session socket, and the *next* command reattaches to the same wedged daemon and hangs again. The user sees an indefinite "Preparing Code Execution" / frozen UI loop with no in-product way out.

## Environment

- agent-browser **0.34.0** (windows-x64, `agent-browser-win32-x64.exe`), Chrome 152.0.7977.54
- Windows 11 (desktop), session name `qwen3-validate-9b123d7cf5e7`
- Integrated via Chatbox.ai's bash sandbox (120 s per-op timeout); the failure is on the agent-browser side — see "Observed evidence"

## Reproduction

1. `agent-browser --session S open "https://www.bing.com/search?q=..."` (or any heavy / slow-to-settle page)
2. Let the page land in a state where the AX tree / DOM extraction never completes (never-rendering state, frozen tab, discarded target)
3. Run `agent-browser --session S snapshot` (or `read` with no URL)
4. The command blocks. No `--timeout` exists for the daemon round trip (`read --timeout` applies only to explicit-URL HTTP fetches; `open`/`snapshot`/`click`/`wait` have no command deadline). The only escape is:
   ```bat
   taskkill /F /T /PID <pid-from-~/.agent-browser/S.pid>
   ```
   Note the **`/T` is required**: killing the daemon alone leaves orphaned `chrome.exe` children holding the wedged socket, and the next command wedges again.

## Observed evidence (one afternoon, single machine)

10 consecutive wedges in the same session; the harness's 120 s timeout was exceeded by 1.5× to 30× and — critically — killing the timed-out CLI child did **not** reap the daemon or its Chrome children, so every subsequent command re-wedged:

| op | command | started | finished | duration | exit |
[...]

Issue #1308, opened 2026-04-29
open hangs pre-handshake on Windows 11 (0-byte stdout, 16+ min) — distinct from #1138


MarbTrek:
Confirming a related-but-distinct failure mode on agent-browser 0.26.0 (Windows 11) — same project tree as the [MarbTrek comment on #1138](https://github.com/vercel-labs/agent-browser/issues/1138) but a different surface.

## Symptom

```
npx agent-browser --session trade --headed --color-scheme dark open http://localhost:3004
```

- 0-byte stdout for 16+ minutes
- No error emitted to stderr or to any log
- No Stop-phase log entry in `~/.agent-browser/cleanup-timeouts.log` (the file is never created — the close step is never reached)
- `tasklist` shows no live `agent-browser` process; `chrome.exe` instances exist but none are launched by the daemon

## Reproduced

- 2 sessions, ~5 hours apart, same Windows 11 host (10.0.26200)
- agent-browser 0.26.0 installed via `npm install -g agent-browser`
- Local-headless launch path (no AgentCore, no `--native`)
- Target URL is a local SolidStart dev server at `http://localhost:3004`

## Why this looks distinct from #1138

#1138 reproduces as `EOF while parsing a value at line 1 column 0` after roughly 10 commands within an existing daemon session — the daemon responds normally for the first ~10 invocations, then starts returning EOF on subsequent calls until `agent-browser close --all` recycles it. The MarbTrek comment on that thread describes the "cumulative-N" variant.

What I'm seeing here is the **first** `open` call against a freshly started session. The daemon never produces a first response — the hang is at handshake/IPC initialization, before any successful command has happened. Different surface, likely different root cause:

- `#1138` "EOF after ~N commands" → suggests resource leak / state bloat / connection pool exhaustion
- This issue "pre-handshake hang on first open" → suggests handshake or IPC initialization failure

PR #1139's read-timeout bump (30s → 120s) is on-the-money for the launch case but wouldn't surface a distinct error here either, since the symptom is "no response ever" rather than "response after long delay".

## Workaround

Migrating off agent-browser to Playwright MCP for the localhost UI-verify use case in this consumer ecosystem. agent-browser stays installed for the narrow cases that need shared-Chrome vault access (logged-in third-party services in the user's personal Chrome profile).

## Offer

[...]

PR #1781, opened 2026-09-04
fix: detached daemon no longer wedges MCP server and captured output on Windows

### LINKED ISSUE #1407: Bug: [Windows] CLI hangs when PowerShell captures output after close-then-open sequence

holny:
## Summary

Fixes #1407. Related: #1308, #1713, and #1641 (the same inheritance mechanism, fixed for doctor's Chrome version detection).

On Windows, the detached daemon spawned on cold start inherited the CLI child's stdout/stderr pipe write ends. Any parent that reads that output to EOF then blocked forever:

- The MCP server's `run_cli` spawns `agent-browser --json open <url>` with piped stdout/stderr and waits on `read_to_end`. The child spawns the daemon, does its work, and exits normally, so the child-exit deadline never fires, but EOF never arrives because the long-lived daemon keeps the pipe open. The server wedge also stops the stdio loop from reading further requests, so every subsequent tool call times out until the MCP client is restarted. This reproduces the classic "works after warm-up, breaks again after every restart" pattern on Windows MCP harnesses.
- The same wedge hits plain PowerShell output capture (`$output = agent-browser open ...`), as reported in #1407.

## Changes

- `connection.rs`: new `prevent_stdio_handle_inheritance()` clears `HANDLE_FLAG_INHERIT` on the CLI's own stdout/stderr before spawning the detached daemon, so nothing inherits the pipe write ends. The dashboard spawn in `main.rs` gets the same treatment.
- `mcp.rs`: `run_cli` now drains child output for a bounded window (2s) after the child exits and returns whatever arrived instead of blocking indefinitely on `read_to_end`. Reader threads publish chunks as they arrive so a blocked EOF cannot hide already-emitted output. This also bounds the kill path, which previously joined the same threads.

I'm aware of #1408, which fixes the same root cause by reimplementing daemon spawn with a raw `CreateProcessW(bInheritHandles = FALSE)` call. This PR takes a lighter approach via `SetHandleInformation` on three handles and additionally hardens `run_cli`, which #1408 does not touch. Happy to consolidate with that PR whichever way the maintainers prefer.

## Evidence

Process captures from a Windows 11 host driving agent-browser through an MCP client, posted in #1407:

- During a wedged first call: `agent-browser.exe mcp` plus the argument-less detached daemon present; the `--json open` child already gone; client never receives a response.
- After client restart with the daemon still alive: the `--json open` child completes and the same call succeeds immediately against the warm daemon.

## Testing

- New unit test `run_command_returns_partial_output_when_a_grandchild_holds_the_pipe`: a backgrounded `sleep` inherits the output pipe and outlives the shell; asserts the call returns the partial output instead of blocking for the sleep's lifetime (fails on the old `read_to_end`-and-join implementation).
- `cargo test`: 1233 passed, 0 failed.
- `cargo clippy --all-targets` and `cargo fmt --check` clean; `cargo check --target x86_64-pc-windows-gnu` passes.

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -789,6 +789,29 @@ fn stop_existing_daemon_for_restart(session: &str) {
     }
 }
 
+/// On Windows a spawned child inherits every inheritable handle we hold. Our
+/// stdout/stderr may be pipes that a parent reads to EOF (the MCP server's
+/// `run_cli`, or a PowerShell capture). The detached daemon we are about to
+/// spawn is long-lived; if it keeps those write ends open the parent blocks
+/// forever waiting for EOF that never arrives. Mark our own stdio handles
+/// non-inheritable so the daemon cannot hold the pipes open.
+#[cfg(windows)]
+pub(crate) fn prevent_stdio_handle_inheritance() {
+    use std::os::windows::io::AsRawHandle;
+    use windows_sys::Win32::Foundation::{SetHandleInformation, HANDLE_FLAG_INHERIT};
+
+    unsafe {
+        for handle in [
+            std::io::stdout().as_raw_handle(),
+            std::io::stderr().as_raw_handle(),
+        ] {
+            if !handle.is_null() {
+                let _ = SetHandleInformation(handle as isize, HANDLE_FLAG_INHERIT, 0);
+            }
+        }
+    }
+}
+
 pub fn ensure_daemon(session: &str, opts: &DaemonOptions) -> Result<DaemonResult, String> {
     let mut restarted = false;
 
@@ -897,6 +920,8 @@ pub fn ensure_daemon(session: &str, opts: &DaemonOptions) -> Result<DaemonResult
     #[cfg(windows)]
     {
         use std::os::windows::process::CommandExt;
+
+        prevent_stdio_handle_inheritance();
 
         let mut cmd = Command::new(&exe_path);
         cmd.env("AGENT_BROWSER_DAEMON", "1");
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1105,6 +1105,7 @@ fn run_dashboard_start(port: u16, allowed_origins: Vec<String>, json_mode: bool)
         use std::os::windows::process::CommandExt;
         const CREATE_NEW_PROCESS_GROUP: u32 = 0x00000200;
         const DETACHED_PROCESS: u32 = 0x00000008;
+        connection::prevent_stdio_handle_inheritance();
         cmd.creation_flags(CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS);
     }
 
--- a/cli/src/mcp.rs
+++ b/cli/src/mcp.rs
@@ -11,8 +11,10 @@ use serde_json::{json, Value};
 use std::collections::BTreeSet;
 use std::env;
 use std::fs;
-use std::io::{self, BufRead, Read, Write};
+use std::io::{self, BufRead, Write};
 use std::process::{Command, Stdio};
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::{Duration, Instant};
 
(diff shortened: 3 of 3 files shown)


3067997259-design:
+1 with repro data from a surface not yet in this thread: the **plain CLI driving an external browser over `--cdp`** (no MCP layer involved), where the caller is an agent harness capturing output in PowerShell.

**Environment**: agent-browser 0.36.0 (`npm i -g`, native exe), Windows 11 Pro 10.0.26200, PowerShell 5.1 + Git Bash. Target is an Electron app with `--remote-debugging-port`, driven as `agent-browser --session S --cdp 9250 ...`.

**Measured** (same machine, same live target, cold daemon per run):

| invocation | stdout | result |
|---|---|---|
| `tab` (cold daemon) | pipe — `2>&1 \| Out-String` / `\| Select-Object` in PowerShell, `\| head` in bash | hangs 11–17 min, ~0 CPU, survives SIGTERM; completes instantly when the daemon is force-killed |
| `tab` (cold daemon) | console | 1.8 s |

While "hung": the daemon is fully up (`.port` sidecar written, control + stream ports listening, CDP connection to the target ESTABLISHED) and raw CDP probes against the same browser/page endpoints answer in ~15 ms — so the target is healthy and the caller is purely waiting on pipe EOF after the CLI has already exited. Force-killing the wedged daemon releases the inherited write-ends and the queued command returns correct output immediately. That signature matches the handle-inheritance mechanism analyzed in #1407 exactly.

One deployment detail worth flagging: the npm **postinstall rewrites the `.cmd`/`.ps1` shims to invoke the exe directly**, bypassing `bin/agent-browser.js` — so every PowerShell capture path (i.e. every agent harness) hits the cold-spawn hang even for users who assume the node wrapper is in the path.

**Interim mitigation** (until this PR ships): we deployed the wrapper approach @Kevofehr described in #1407 — shims route through a small node launcher that spawns the exe with node-owned pipes when stdout/stderr are captured (`isTTY` check), forwards chunks, and exits after a 500 ms output quiet window once the child exits (5 s hard cap). One implementation note for anyone copying this: the launcher must react to the child's `exit` event, not `close` — `close` also waits for the stdio streams to end, and the daemon grandchild holds exactly those write-ends, so a `close`-based launcher hangs on the same EOF trap it exists to break. With the launcher in place, cold captured `tab` returns in ~2–6 s and warm-daemon calls in ~2 s.

Happy to run this PR's branch on this machine against both the MCP surface and the plain `--cdp` CLI surface if that helps verification.

holny:
@ctate gentle nudge on this one — it's been a week. #1407 now has three independent confirmations, including the plain-CLI `--cdp` repro above from someone not using MCP at all. Branch is 12 commits behind main but merges clean; happy to rebase if that helps move it along.

holny:
@ctate just bumping this — three independent confirmations on the issue so far, and the branch still merges clean against main. Happy to rebase or split the diff if that makes review easier.

PR #1788, opened 2026-09-04
fix: give navigation-lifecycle timeouts their own agent guidance

### LINKED ISSUE #1767: fix(cli): navigate load-timeout error message pushes agents into retry loops
## Problem

When `open`/`navigate` times out waiting for `Page.loadEventFired`, the raw error ("Timeout waiting for Page.loadEventFired") is rewritten by `to_ai_friendly_error` (`cli/src/native/browser.rs`) into:

> "Operation timed out. **The page may still be loading** or the element may not exist."

For AI-agent callers (the primary audience of the MCP server) this phrasing is actively misleading: it asserts the page "may still be loading" and references an "element" that no navigation was waiting for. The natural agent response is to retry `open` or issue another `wait --load` — burning another full 25s timeout per loop. Observed in practice: agents loop for many minutes on pages whose `load` event will never fire (SSO redirect chains, HTTP auth prompts, permanently pending subresources), and to the user the session looks hung.

## Root cause

A single catch-all branch handles every timeout the same way:

```rust
if lower.contains("timeout") {
    return "Operation timed out. The page may still be loading or the element may not exist."
}
```

It can't distinguish an **element wait** timeout (where the current text fits) from a **navigation lifecycle** timeout, where the page is often already interactive (`document.readyState === 'interactive'`, DOM usable) and the productive next step is to inspect current state — not blind retry.

## Proposal

Match the lifecycle error shapes ("Timeout waiting for Page.loadEventFired", `Page.domContentEventFired`, `networkidle`) before the generic `"timeout"` branch, and emit a navigate-specific, actionable message, e.g.:

> "Navigation did not reach the 'load' event within 25s. The page may already be usable — call `url` or `snapshot` to inspect current state before retrying. Possible causes: slow SSO redirects, an HTTP auth or client-certificate prompt, or permanently pending subresources."

Optionally include the probed `document.readyState` in the message — cheap and highly diagnostic ("readyState=interactive" tells the agent the DOM is ready and it can proceed).

(Related: #1765 proposes `--wait-until` so callers can pick a weaker strategy; this issue is about what the error says when the wait still times out.)

Happy to implement with tests.

holny:
## Summary

Fixes #1767.

Timeouts waiting for `Page.loadEventFired`, `Page.domContentEventFired`, or `networkidle` were flattened by `to_ai_friendly_error` into the element-oriented text "Operation timed out. The page may still be loading or the element may not exist." For agent callers, that wording asserts an element problem where no element was involved, and the natural reaction is to retry `open` or re-run `wait --load`, each burning the full timeout. On pages whose load event will never fire (SSO redirect chains, HTTP auth prompts, permanently pending subresources) agents loop for minutes and the session looks hung.

`to_ai_friendly_error` now classifies those navigation-lifecycle timeouts separately and returns guidance that states the load event never fired, names the common causes, and points at inspecting the page that is already there (`get url`, `snapshot`) or `wait --load` with a larger timeout instead of re-running the navigation. Element-wait timeouts keep the existing message.

## Testing

- New unit test `test_to_ai_friendly_error_navigation_lifecycle_timeout` covers all three lifecycle errors and asserts the old element-flavored text no longer appears.
- `cargo test`: 1233 passed, 0 failed.

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -216,6 +216,19 @@ pub fn to_ai_friendly_error(error: &str) -> String {
         return "Another element is covering the target element. Try scrolling or closing overlays."
             .to_string();
     }
+    // Navigation-lifecycle timeouts mean the load/DOMContentLoaded event (or
+    // the network-quiet state) never arrived within the timeout. That is a
+    // page condition (auth prompts, hanging subresources, endless redirects),
+    // not a locator problem: element-oriented advice sends agents into
+    // open/wait retry loops that each burn the full timeout. Steer them
+    // toward inspecting the page that is already there instead.
+    if lower.contains("loadeventfired")
+        || lower.contains("domcontenteventfired")
+        || lower.contains("networkidle")
+    {
+        return "The page did not reach the requested navigation state before the timeout: the load event never fired. Common causes are HTTP auth prompts, hanging subresources, or endless redirects. The page may still be usable: check where it landed with get url or snapshot. To keep waiting, run wait --load (or another navigation wait) with a larger timeout instead of re-running the navigation."
+            .to_string();
+    }
     if lower.contains("timeout") {
         return "Operation timed out. The page may still be loading or the element may not exist."
             .to_string();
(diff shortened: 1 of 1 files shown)


holny:
@ctate small one — just reworks the nav-timeout message to point at the actual lifecycle causes. Still merges clean; let me know if you'd rather see it as docs instead of an error-string change.

PR #1789, opened 2026-09-04
fix: stop sending nativeVirtualKeyCode to Input.dispatchKeyEvent

### LINKED ISSUE #1775: press <key> starts a permanent ~8,000/sec stream of phantom keydown events (code: Quote, keyCode: 222)
A single `agent-browser press <key>` leaves the page receiving a permanent stream of phantom `keydown` events at roughly **8,000 per second**. The intended keypress itself is delivered correctly; the phantom stream is separate, starts immediately after, and never stops until the page is navigated.

This makes `press` unusable for verifying any keyboard-driven UI: a handler bound to `keydown` is invoked ~8,000 times a second, and a handler that reacts to *any* key (rather than matching a specific one) fires continuously.

## Reproduction

No app or server needed — `about:blank` is enough.

```bash
agent-browser open about:blank
agent-browser eval 'window.__c=0; addEventListener("keydown",()=>window.__c++,true); "armed"'
agent-browser press ArrowRight
agent-browser eval 'new Promise(r=>{const a=window.__c;setTimeout(()=>r(JSON.stringify({delta:window.__c-a})),1000)})'
```

Last command prints:

```json
{"delta":8036}
```

Expected `{"delta":0}` — the press already happened before the sampling window opened.

Repeated reads keep climbing indefinitely (`259` → `520` → `829` → `1139` …) with no further commands issued.

## What the phantom events look like

Capturing the first three `keydown` events after one `press ArrowRight`:

```json
[
  {"key":"ArrowRight",   "code":"ArrowRight","keyCode":39, "repeat":false,"trusted":true},
  {"key":"Unidentified", "code":"Quote",     "keyCode":222,"repeat":false,"trusted":true},
  {"key":"Unidentified", "code":"Quote",     "keyCode":222,"repeat":false,"trusted":true}
]
```

- The first event is the requested key and is correct, and exactly one matching `keyup` arrives — so `press` itself works.
- Every subsequent event is `code: "Quote"` / `keyCode: 222` / `key: "Unidentified"`, with `repeat: false` and `isTrusted: true`.
- Because `repeat` is `false` and the real key was released, this does not look like OS/CDP auto-repeat but like a loop dispatching a malformed key event whose unset key resolves to `Quote`.

## Scope

| | |
|---|---|
| Triggering keys | `ArrowRight`, `ArrowLeft`, `Enter`, `Escape`, `a` all trigger it |
| Non-triggering | `Tab` does not |
| Stops on | navigation only (`open` clears it; the flood is page-scoped, not daemon-scoped) |
| Does not stop on | pressing another key — a subsequent `press Tab` leaves the stream running |
| Unaffected | `keyboard type <text>` does not trigger it |

## Environment

- agent-browser 0.36.0 (Homebrew)
- Chrome for Testing 152.0.7977.75, headless (default), launched by `agent-browser install`
- macOS 26.5.1, arm64

## Possibly related

#1499 and #972 both touch `keyDown` event construction (the `text` field), and #1453 concerns CDP `commands` being dropped on keyboard input. I could not find an existing issue describing this phantom-key stream, but the `keyCode: 222` default suggests it may share a root cause with how key events are built.

holny:
## Summary

Fixes #1775.

A single `press <key>` left the page receiving a permanent stream of phantom keydown events at roughly 8000/sec (code "Quote", keyCode 222, key "Unidentified", repeat false, isTrusted true) that only stopped on navigation. The requested key itself was delivered correctly, so `press` appeared to work while corrupting any keyboard-driven UI on the page. The affected Chrome process also spins at 100% CPU for as long as the flood runs.

## Root cause, empirically bisected

Reproduced on about:blank, then dispatched every parameter combination over a raw CDP connection to the same page and measured the keydown counter delta one second after each press (about:blank navigation between variants to reset any stuck state):

| dispatchKeyEvent params | 1s keydown delta |
| --- | --- |
| keyDown, code + windowsVirtualKeyCode + nativeVirtualKeyCode | 14284 (FLOOD) |
| rawKeyDown, same set | 19606 (FLOOD) |
| keyDown, windowsVirtualKeyCode + nativeVirtualKeyCode, no code | 11976 (FLOOD) |
| keyDown, code, no VK codes | 0 |
| keyDown, windowsVirtualKeyCode + code, no nativeVirtualKeyCode | 0 |
| key only | 0 |

Every flood variant carries `nativeVirtualKeyCode`; every clean variant omits it. Headless Chrome synthesizes new trusted keydown events in a tight loop for as long as that field is present (the phantom events carry code "Quote" / VK 222 because Blink derives the DOM code from the malformed native event rather than our requested key). Playwright's keyboard implementation likewise sends only `windowsVirtualKeyCode`, so all dispatch sites here now leave `nativeVirtualKeyCode` unset.

## Testing

- Reproduction from the issue now reports `{"delta":0}` with the fix and `{"delta":5790}` without it; the real keypress still arrives exactly once.
- `cargo test`: 1232 passed, 0 failed (one pre-existing flake, `test_execute_unknown_command`, passed on rerun).

### DIFF
--- a/cli/src/native/interaction.rs
+++ b/cli/src/native/interaction.rs
@@ -258,7 +258,7 @@ pub async fn type_text_into_active_context(
                         text: text_str.clone(),
                         unmodified_text: text_str,
                         windows_virtual_key_code: Some(key_code),
-                        native_virtual_key_code: Some(key_code),
+                        native_virtual_key_code: None,
                         modifiers: None,
                     },
                     Some(session_id),
@@ -275,7 +275,7 @@ pub async fn type_text_into_active_context(
                         text: None,
                         unmodified_text: None,
                         windows_virtual_key_code: Some(key_code),
-                        native_virtual_key_code: Some(key_code),
+                        native_virtual_key_code: None,
                         modifiers: None,
                     },
                     Some(session_id),
@@ -313,6 +313,12 @@ pub async fn press_key(client: &CdpClient, session_id: &str, key: &str) -> Resul
 /// Modifier values follow the CDP `Input.dispatchKeyEvent` spec:
 /// 1 = Alt, 2 = Control, 4 = Meta (Cmd), 8 = Shift.
 ///
+/// `nativeVirtualKeyCode` must stay unset. Sending it alongside
+/// `windowsVirtualKeyCode` makes headless Chrome enter an endless synthetic
+/// keydown loop (~8000 events/sec, code "Quote" / VK 222) that only stops on
+/// navigation; see issue #1775. Playwright likewise sends only
+/// `windowsVirtualKeyCode`.
+///
 /// Callers that need a platform-appropriate modifier (e.g. Cmd on macOS,
 /// Ctrl elsewhere) must choose the value themselves -- see `cfg!(target_os)`.
 pub async fn press_key_with_modifiers(
@@ -342,7 +348,7 @@ pub async fn press_key_with_modifiers(
                 text: text.clone(),
                 unmodified_text: text.clone(),
                 windows_virtual_key_code: Some(key_code),
-                native_virtual_key_code: Some(key_code),
+                native_virtual_key_code: None,
                 modifiers,
             },
             Some(session_id),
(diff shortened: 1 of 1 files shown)


ShacktarMalacky (approved):

holny:
@ctate gentle nudge — mausferd hit the same phantom keydown flood on macOS arm64 (0.37.1), so it's not Windows-specific, and the branch still merges clean against main. Happy to adjust if the 4-site parameter removal needs a different shape.

PR #1792, opened 2026-09-05
fix: reconnect to the running browser instead of relaunching on CDP socket loss

simonellefsen:
## The problem

A dead CDP socket is not a dead browser, but `handle_launch` treats it as one. In the relaunch decision, `!mgr.is_connection_alive()` sits in the same `||` chain as a changed launch hash and an exited process:

```rust
is_external != was_external
    || hash_changed
    || storage_state_requires_clean_launch
    || mgr.has_process_exited()
    || !mgr.is_connection_alive().await
```

So a dropped WebSocket takes the same path as a genuinely unusable browser: `close_current_browser`, then launch a new one. That discards the profile — cookies, storage, and any logged-in session — to rebuild a browser that was never broken.

Chrome drops that socket on its own. On long-lived idle sessions it resets the DevTools WebSocket while the process keeps running normally:

```
[cdp] WebSocket Error: WebSocket protocol error: Connection reset without closing handshake
```

I hit this nightly on a headed session running Chrome 152. The next morning's first command destroys a perfectly good browser, and any login in that profile has to be redone by hand.

Everything needed to resume is still there. Captured on a session in exactly this state — daemon and Chrome both still alive, socket already reset:

```
# daemon's CDP socket: dead
agent-bro 41791  TCP 127.0.0.1:50604->127.0.0.1:50603 (CLOSED)

# Chrome: still listening on the same port
Google    41792  TCP 127.0.0.1:50603 (LISTEN)

$ curl -s http://127.0.0.1:50603/json/version
Browser : Chrome/152.0.7977.76
wsUrl   : ws://127.0.0.1:50603/devtools/browser/ea6f0643-cb32-4031-86bf-4c8f80a21d15
```

The browser is up, the endpoint answers, and `ws_url` is still valid. Only our end of the connection is gone.

`auto_save_restore_state` does not cover this. Saving reads state over CDP, which is precisely what is unavailable when the socket is dead.

## The change

Split "unusable" from "unreachable". A changed configuration or an exited process still relaunches immediately, and the cheap checks still run before the expensive probe. A failed liveness probe on an otherwise-valid browser now tries `BrowserManager::reconnect()` first, and relaunches only if that fails.

`reconnect()` opens a fresh WebSocket to the same `ws_url`, re-attaches all targets, restores the bound/active tab by target id through the existing `restore_target_binding`, and replays the CDP-applied emulation that a new connection does not inherit.

Two details worth flagging for review:

- `discover_and_attach_targets` **appends** to `pages`, so `reconnect` clears it first — reusing it without clearing would duplicate the whole tab list on every reconnect.
- Replaying emulation required retaining `user_agent`, `color_scheme` and the external-connection headers on the manager. They were applied at launch and then forgotten, so without this a reconnected session would have silently differed from the one it replaced.

Direct-page connections (provider CDP proxies) are refused rather than half-reattached, since their WebSocket *is* the page session; callers fall back to a relaunch for those.

## Testing

`cargo build --release`, `cargo fmt --check` clean, `cargo clippy --all-targets` introduces no new warnings (the two it reports are pre-existing, in `cdp/client.rs` and an unrelated test). **1233 tests pass, 0 failures.**

Added coverage for the direct-page refusal.

**The reconnect path itself has no automated test**, and I want to be upfront about that: the existing `test_manager` scaffolding backs the manager with a socket that accepts the connection and then stays silent, so it cannot answer the `Target.*` commands re-attach issues. Happy to add a responding CDP mock if you'd like that before merging.

Verified manually instead:

- against a browser whose socket had been reset, the DevTools endpoint still served its browser target and the session was recoverable (output above)
- the normal reuse path is unchanged — same browser pid across repeated commands on a scratch session

## One caveat on the cause

I can demonstrate the mechanism and the fix. I have **not** proven *why* Chrome resets the socket. The onset correlates with Chrome 151 → 152 on this machine, but that is correlation, not a root cause. The fix stands regardless: agent-browser should not destroy a healthy browser and its profile because a socket dropped, whatever caused the drop.

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -2203,6 +2203,25 @@ async fn close_active_provider_session(state: &mut DaemonState) {
     state.active_provider_connection = false;
 }
 
+/// Reset the daemon state that belongs to a specific CDP *connection* rather
+/// than to the browser itself.
+///
+/// Everything here is invalidated when the socket goes away, whether because the
+/// browser was closed or because the connection dropped and was re-established:
+/// `Target.setAutoAttach` is a per-connection command (and the flag gates
+/// re-sending it), and the session ids in the iframe maps, the WebMCP registry
+/// and the screencast flag all belong to the sessions the old socket owned.
+///
+/// Deliberately excludes `launch_hash`: that describes the browser's
+/// configuration, which a reconnect does not change.
+fn reset_cdp_connection_state(state: &mut DaemonState) {
+    state.network_auto_attach_installed = false;
+    state.iframe_sessions.clear();
+    state.active_iframe_sessions.clear();
+    state.webmcp.clear_all();
+    state.screencasting = false;
+}
+
 pub(crate) async fn close_current_browser(state: &mut DaemonState) -> Result<(), String> {
     let close_error = if let Some(mut mgr) = state.browser.take() {
         mgr.close().await.err()
@@ -2212,11 +2231,7 @@ pub(crate) async fn close_current_browser(state: &mut DaemonState) -> Result<(),
 
     close_active_provider_session(state).await;
     state.launch_hash = None;
-    state.network_auto_attach_installed = false;
-    state.iframe_sessions.clear();
-    state.active_iframe_sessions.clear();
-    state.webmcp.clear_all();
-    state.screencasting = false;
+    reset_cdp_connection_state(state);
     state.reset_input_state();
     state.update_stream_client().await;
 
@@ -4578,6 +4593,16 @@ async fn handle_launch(cmd: &Value, state: &mut DaemonState) -> Result<Value, St
             // allow-domains and answers proxy auth, so skipping this would drop
             // a security control while still reporting a healthy reused browser.
             // Rebuild the same wiring the attach branches below install.
+            //
+            // Clearing the connection-scoped state first is what makes
+            // install_network_controls_or_close effective: it re-sends
+            // Target.setAutoAttach only when network_auto_attach_installed is
+            // false, and auto-attach is a per-connection command that did not
+            // survive the drop. Leaving the flag set would silently skip it, so
+            // popups and new windows opened after a reconnect would never be
+            // attached — and therefore never filtered by allow-domains or
+            // covered by proxy auth.
+            reset_cdp_connection_state(state);
             let has_proxy_auth = store_proxy_credentials(state, &launch_options).await;
             state.reset_input_state();
             state.subscribe_to_browser_events();
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -3502,6 +3502,22 @@ async fn install_active_network_controls(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


simonellefsen:
Good catch — this was a real defect, and I've fixed it in c9c791c.

Confirmed the mechanism: `reconnect()` swaps `BrowserManager::client` for a new `Arc<CdpClient>`, but everything `handle_launch` installs after an attach captures the *old* one —

- `subscribe_to_browser_events` → `self.event_rx = Some(browser.client.subscribe())`
- `start_fetch_handler` / `start_dialog_handler` → tasks spawned against that client
- `update_stream_client` → `Arc::clone(&m.client)` handed to the stream server

and the `reused: true` early-return skipped rebuilding any of it, so they'd all have gone on talking to a dead socket. The Fetch handler is the one that matters most: it enforces allow-domains and answers `Fetch.authRequired`, so the reconnect would have silently dropped a network control while reporting a healthy reused browser. That's a worse failure than the relaunch I was trying to avoid.

The reused branch now replays the same wiring the attach branches install when a reconnect occurred, plus `apply_launch_init_scripts` — `Page.addScriptToEvaluateOnNewDocument` registrations die with the old session ids too, which I'd also missed. Proxy-credential storage is factored into a `store_proxy_credentials` helper so the reconnect path and the attach branches share it instead of duplicating the block.

`cargo build --release` and `cargo fmt --check` clean; 1233 tests pass, 0 failures.

Still worth a careful look from a maintainer: I've matched the attach branches' sequence, but you'll know better than I do whether anything else in `DaemonState` is bound to the CDP client and needs rebuilding here.

simonellefsen:
Correct again — fixed in d42da28.

`Target.setAutoAttach` is a per-connection command, but re-sending it is gated on `network_auto_attach_installed`, and that flag is only cleared by `close_current_browser` — which a reconnect never calls. So it stayed `true`, `install_active_network_controls` skipped `enable_browser_auto_attach_pub`, and auto-attach was never re-established on the new socket. Popups and new windows opened after a reconnect would have been attached to nothing, and therefore covered by neither allow-domains filtering nor proxy auth.

Following the same reasoning through the rest of `close_current_browser` turned up more state in the same category, all keyed by the old connection's session ids and all being carried over: `iframe_sessions`, `active_iframe_sessions`, the WebMCP registry, and `screencasting`.

Rather than patch the one flag, I extracted `reset_cdp_connection_state` for the state that belongs to a *CDP connection* rather than to the browser, and call it from both `close_current_browser` and the reconnect path so the two can't drift apart as more connection-scoped state gets added. `launch_hash` is deliberately excluded — it describes the browser's configuration, which a reconnect doesn't change.

Build and `cargo fmt --check` clean; 1233 tests pass, 0 failures.

Both findings were on the same theme — I was treating the reconnect as if it only swapped a socket, when it invalidates every session id and per-connection command in the daemon. Worth a maintainer's eye on whether `reset_cdp_connection_state` now covers the full set; that's the part I'm least able to verify from outside the project.

PR #1793, opened 2026-09-06
fix: try every loopback host when connecting to a launched Chrome

### LINKED ISSUE #1791: Auto-launch fails on WSL2: daemon-spawned Chrome with --remote-debugging-port=0 LISTENs but refuses all CDP connects (fixed port works)
## Bug: auto-launch fails on WSL2 — daemon-spawned Chrome LISTENs on ephemeral port but refuses all CDP connects

`agent-browser open` / `doctor` fail with:

```
✗ Auto-launch failed: CDP WebSocket connect failed: IO error: Connection refused (os error 111)
```

Same failure on **0.35.1 and 0.36.0** (upgraded to retest; 0.36.0 still spawns with `--remote-debugging-port=0`, confirmed via strace).

### Environment

- agent-browser 0.35.1, 0.36.0 (npm global)
- Google Chrome for Testing 152.0.7977.75 (bundled, `~/.agent-browser/browsers/`)
- WSL 2.7.12.0, kernel 6.18.33.2-microsoft-standard-WSL2, WSLg 1.0.73.2, Ubuntu 24.04.4 LTS

### Repro

```bash
agent-browser close --all
agent-browser open about:blank
# ✗ Auto-launch failed: CDP WebSocket connect failed: IO error: Connection refused (os error 111)
```

### What I found

Strace on the daemon shows it spawns Chrome with **`--remote-debugging-port=0`** (plus `--headless=new`, `--user-data-dir=/tmp/agent-brows*`, etc.). Reproducing that exact spawn manually:

| Case | Result |
|---|---|
| Chrome `--remote-debugging-port=0`, curl `/json/version` on the assigned port | **refused, 20/20 attempts** |
| Chrome `--remote-debugging-port=9222` (fixed), same flags | **HTTP 200** |
| Chrome `--remote-debugging-port=40655` (fixed high port), same flags | **HTTP 200** |
| `python3 -m http.server` on the *exact refused port* after killing Chrome | **HTTP 200** |

So it is specific to Chrome's port-0 path in this environment — not the port number, not loopback, not the flags.

The eerie part: with port 0, the socket **is LISTENing**, owned by the live main browser process, same netns, empty accept queue, backlog 10:

```
LISTEN 0  10  127.0.0.1:35535  0.0.0.0:*  users:(("chrome",pid=14241,fd=66))
```

yet every connect gets `ECONNREFUSED`. The listen-socket inode is held only by the main chrome process (checked every `/proc/*/fd`). `tcp_abort_on_overflow=0`.

### Eliminated

- Missing libs (`ldd` clean), `/dev/shm` (3.4 GB free), sandbox (fails identically with and without `--no-sandbox`)
- Stale processes (reproduced from clean state), network namespace (chrome and client share `net:[4026531833]`)
- Loopback itself (python/t3/opencode servers all answer on 127.0.0.1)
- Daemon never retries: the trace shows exactly **one** CDP `connect()` before giving up

### Relationship to #868

Related but distinct: #868's Chrome *exits early* with no DevTools URL. Here Chrome stays alive with a LISTEN socket that refuses everything. Same theme ("manual launch works, daemon-spawned doesn't") — mbakhodir's comment there matches — but a different failure point.

### Workaround (verified)

Launch Chrome on a fixed port, then attach — full session works (`open`, `read`):

```bash
chrome --remote-debugging-port=9222 --headless=new --user-data-dir=/tmp/ab-work about:blank &
agent-browser connect 9222
```

### Suggested directions

- Avoid `--remote-debugging-port=0` (pin a fixed/known-free port), or
- Retry the CDP connect with backoff instead of a single attempt, or
- Surface this state distinctly ("DevTools port LISTENing but refusing") so it stops looking like Chrome never started

holny:
## Summary

Fixes #1791.

On WSL2, Chrome spawned with `--remote-debugging-port=0` LISTENs on its ephemeral port, but the daemon's connect to `ws://127.0.0.1:<port>` is refused with `os error 111`; spawning with a fixed port works. Port discovery via `DevToolsActivePort` only tells us the port, not which loopback address the server bound, and on some stacks (WSL2 kernels, IPv6-preferred loopback) the ephemeral listener comes up on the IPv6 loopback `[::1]` instead of IPv4 `127.0.0.1` — so the hard-coded `127.0.0.1` dial can never succeed no matter how long we wait.

## Changes

- New `connect_launched_chrome()` in `browser.rs`: after launch, try `127.0.0.1`, then `[::1]`, then `localhost`, within a bounded (6s) retry window so a transient gap between the port file appearing and the server accepting is also absorbed.
- New `loopback_ws_candidates()` builds the candidate URLs; URLs that do not point at `127.0.0.1` (user-supplied `--cdp` targets) are returned untouched, so the fallback never silently rewrites a host the user chose.

## Testing

- New unit tests: candidate expansion covers both loopbacks; non-127.0.0.1 URLs pass through unchanged.
- `cargo test`: 1234 passed, 0 failed on the parent branch; one known flake (`test_execute_unknown_command`) on the rebased branch passes standalone.
- Local smoke test (macOS, headless): close → open → get url → eval all succeed through the new connect loop.

We do not have a WSL2 environment in CI, so verification from @the reporter's side would be valuable — happy to cut a test build.

### DIFF
--- a/cli/src/native/browser.rs
+++ b/cli/src/native/browser.rs
@@ -192,6 +192,52 @@ fn active_page_index_after_add(
     }
 }
 
+/// Loopback hosts to try for a launched Chrome's DevTools WebSocket, in order.
+const LAUNCH_LOOPBACK_HOSTS: [&str; 3] = ["127.0.0.1", "[::1]", "localhost"];
+
+/// Build the candidate WebSocket URLs for a launched Chrome. Port discovery
+/// hard-codes 127.0.0.1, but on some stacks (WSL2, issue #1791) the ephemeral
+/// listener comes up on the IPv6 loopback instead, so the remaining loopback
+/// hosts are tried before giving up. URLs that do not point at 127.0.0.1 are
+/// returned unchanged.
+fn loopback_ws_candidates(ws_url: &str) -> Vec<String> {
+    if !ws_url.contains("127.0.0.1") {
+        return vec![ws_url.to_string()];
+    }
+    LAUNCH_LOOPBACK_HOSTS
+        .iter()
+        .map(|host| ws_url.replacen("127.0.0.1", host, 1))
+        .collect()
+}
+
+/// Connect to a freshly launched Chrome's DevTools WebSocket.
+///
+/// `DevToolsActivePort` only tells us the port, not which loopback address the
+/// server bound, and the listen socket can come up a moment after the file
+/// appears. Try every loopback host within a short retry window so a transient
+/// refusal or an IPv6-only listener does not fail the launch.
+async fn connect_launched_chrome(ws_url: &str) -> Result<CdpClient, String> {
+    const CONNECT_DEADLINE: Duration = Duration::from_secs(6);
+    let deadline = Instant::now() + CONNECT_DEADLINE;
+    let candidates = loopback_ws_candidates(ws_url);
+    let mut last_err = String::new();
+    loop {
+        for candidate in &candidates {
+            match CdpClient::connect(candidate).await {
+                Ok(client) => return Ok(client),
+                Err(e) => last_err = format!("{candidate}: {e}"),
+            }
+        }
+        if Instant::now() >= deadline {
+            return Err(format!(
+                "Could not connect to the launched browser's DevTools socket after trying {} loopback candidate(s). Last error: {last_err}",
+                candidates.len()
+            ));
+        }
+        tokio::time::sleep(Duration::from_millis(250)).await;
+    }
+}
+
 /// Converts common error messages into AI-friendly, actionable descriptions.
 pub fn to_ai_friendly_error(error: &str) -> String {
     let lower = error.to_lowercase();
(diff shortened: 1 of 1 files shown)


satyalyadav:
Built 082596a and retested on WSL2 (Chrome 152.0.7977.75). Still fails, now after 7.5s: Could not connect to the launched browser's DevTools socket after trying 3 loopback candidate(s). Last error: ws://localhost:46209... Connection refused (111).

Manual port 0 Chrome, curl on each candidate: 127.0.0.1 refused (111), localhost refused (111), [::1] Cannot assign requested address (99). So the ephemeral listener refuses on every loopback from the start. Not IPv6-only, not transient.

satyalyadav:
Two findings from testing that are relevant here, full fix proposed in #1796:

1. With an explicit --remote-debugging-port, Chrome 152 headless never writes DevToolsActivePort. It only prints the DevTools listening URL to stderr. The file only appears with port 0. So on the pre-bound path the file poll always runs its full course.
2. The stderr fallback shared one 30s deadline with the file poll, so it expired instantly whenever the file was absent. It was effectively dead code.

#1796 pre-binds an explicit loopback port to dodge the broken port-0 path from #1791, watches the file and stderr concurrently, and dedupes user-supplied port flags to one. It complements this PR rather than replacing it: the loopback retry here is still the right safety net for the port-0 fallback. Verified on WSL2, open went from 30s to under 2s.

holny:
@ctate bumping this — regardless of whether the WSL 2.9.x networking fix clears some reports, the multi-loopback fallback here is a pure safety net and shouldn't affect the normal path. Still merges clean against main.

PR #1797, opened 2026-09-06
feat: add interpolated mouse movement

petehunt:
Implements idea 1 from the browser-tool ideas gist.

Adds deterministic smooth mouse movement and seeded curved human movement for low-level moves, clicks, and drags. Cursor position persists per daemon session, movement preserves pressed-button state, and paths always finish at the exact target.

Includes CLI and MCP parity, documentation updates, parser and path unit tests, and existing mouse e2e coverage.

Validation:
- `cd cli && cargo test` (1 unrelated environment-race failure in the parallel suite; isolated rerun passes)
- `cd cli && cargo test e2e_mouse -- --ignored --test-threads=1`
- `cd cli && cargo clippy`

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -489,6 +489,8 @@ pub struct DaemonState {
     /// so they never block the agent.
     dialog_handler_task: Option<tokio::task::JoinHandle<()>>,
     pub mouse_state: MouseState,
+    /// Session-wide pointer behavior, configured by `--input-mode`.
+    pub input_mode: String,
     /// Tracks the currently open JavaScript dialog (alert/confirm/prompt), if any.
     pub pending_dialog: Option<PendingDialog>,
     /// A mouse button left logically down because a dialog opened between
@@ -612,6 +614,7 @@ impl DaemonState {
             fetch_handler_task: None,
             dialog_handler_task: None,
             mouse_state: MouseState::default(),
+            input_mode: "instant".to_string(),
             pending_dialog: None,
             pending_pointer_release: None,
             auto_dialog: !matches!(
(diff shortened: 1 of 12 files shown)


cli/src/native/actions.rs
```diff
@@ -11915,28 +11968,137 @@ async fn handle_inserttext(cmd: &Value, state: &DaemonState) -> Result<Value, St
...
+    mouse_state: &mut MouseState,
+    target_x: f64,
+    target_y: f64,
+    duration_ms: u64,
+    requested_steps: Option<usize>,
+    human: bool,
+    seed: u64,
+    buttons: i32,
+) -> Result<(), String> {
+    let start_x = mouse_state.x;
```
petehunt, line 12002: [P2] Keep interpolation origin in sync with element interactions

The interpolation starts from mouse_state, but existing click/hover/dblclick/check paths dispatch pointer events without updating that state. For example, hovering an element at (450, 320) and then interpolating to (550, 320) still starts the path at the initial (0, 0), sweeping across unrelated elements instead of continuing from the cursor. Return the dispatched endpoint from element interactions and update the shared state before the next move; also cover semantic click/hover and checkbox interactions. I am adding that fix with a real-Chrome path regression.

petehunt:
Fixed the stale interpolation origin in 65a0a65. Element clicks, hovers, double-clicks, checkbox interactions, and semantic interaction paths now record the actual dispatched cursor endpoint without an extra CDP query.

All three mouse Chrome e2es pass, including the new test checking actual intermediate mousemove coordinates after element interactions, plus the existing drag-state and pointer-capture tests. Formatting and clippy with warnings denied pass on the final clean rebuild. The broader unit run hit the existing parallel environment-variable race in test_execute_unknown_command; its isolated rerun passed.

petehunt:
Ready for review/merge

petehunt:
Superseded by upstream draft PR #1810 so repository CI can run on the branch.

PR #1798, opened 2026-09-07
feat: add automatic snapshot deltas

petehunt:
Implements idea 3 from the browser-tool ideas gist.

`agent-browser snapshot --delta` now returns full state on the first observation, a compact unchanged result for identical trees, and bounded structural add/remove/replace changes for modest updates. Baselines are isolated by page session and snapshot options, revisioned, memory-bounded, and reset to full output on URL or option changes. `--full` is the explicit full-state escape hatch.

Includes CLI and MCP parity, human-readable output, documentation updates, and tests for parser behavior, unchanged observations, structural operations, and full-suite compatibility.

Validation:
- `cd cli && cargo test -- --test-threads=1` (1,242 passed, 112 ignored)
- `cd cli && cargo clippy`

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -5292,15 +5292,36 @@ fn snapshot_delta_response(
         }
     }
 
+    // Ref metadata contains only role/name, so it cannot describe checkbox
+    // state, text, values, hierarchy, or ordering. Include an exact tree splice
+    // alongside ref operations so every accepted revision can be reconstructed.
+    let before_lines: Vec<&str> = previous.tree.split('\n').collect();
+    let after_lines: Vec<&str> = tree.split('\n').collect();
+    let prefix = before_lines
+        .iter()
+        .zip(&after_lines)
+        .take_while(|(a, b)| a == b)
+        .count();
+    let suffix = before_lines[prefix..]
+        .iter()
+        .rev()
+        .zip(after_lines[prefix..].iter().rev())
+        .take_while(|(a, b)| a == b)
+        .count();
+    let tree_change = json!({
+        "startLine": prefix,
+        "deleteCount": before_lines.len() - prefix - suffix,
+        "lines": after_lines[prefix..after_lines.len() - suffix],
+    });
     let delta = json!({
         "kind": "delta",
         "baseRevision": previous.revision,
         "revision": revision,
         "changes": changes,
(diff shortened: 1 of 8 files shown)


cli/src/native/actions.rs
```diff
@@ -5192,7 +5204,110 @@ async fn handle_snapshot(cmd: &Value, state: &mut DaemonState) -> Result<Value,
...
+    let delta = json!({
+        "kind": "delta",
+        "baseRevision": previous.revision,
+        "revision": revision,
+        "changes": changes,
+    });
+    let delta_size = serde_json::to_vec(&delta).map_or(usize::MAX, |bytes| bytes.len());
+    let structurally_unrepresented = delta["changes"].as_array().is_none_or(Vec::is_empty);
+    if structurally_unrepresented || delta_size >= tree.len().saturating_mul(7) / 10 {
+        json!({
```
petehunt, line 5325: [P1] Preserve tree changes that are absent from ref metadata

The ref map contains only role/name. If a sufficiently large page changes a button name and a checkbox state in the same observation, changes contains the name replacement and the nonempty check permits a delta, but the checked-state change is silently dropped. The same applies to text, input values, hierarchy, and ordering. The baseline is then advanced, so the next observation says unchanged and the omitted state never reaches the caller. A nonempty ref diff is not evidence that it represents the whole tree. Include an exact tree patch alongside ref operations (or fall back to full whenever the tree cannot be reconstructed) and test mixed represented/unrepresented updates. I am adding a bounded line splice to preserve all tree changes.

petehunt:
Fixed the lost-state issue in c58bf0b. Deltas now include an exact treeChange line splice alongside ref metadata operations, and the size limit accounts for both. Text, values, checkbox state, hierarchy, and ordering can therefore be reconstructed instead of silently omitted. The wire format and application steps are documented for CLI and MCP consumers.

Manual Chrome validation changed a button name and checkbox state together, reconstructed the exact full tree from the delta, and verified the next observation was unchanged. The delta tests, formatting, and clippy pass on the final clean rebuild. The broader unit run hit the existing parallel environment-variable race in test_execute_unknown_command; its isolated rerun passed.

petehunt:
Ready for review/merge

petehunt:
Superseded by upstream draft PR #1811 so repository CI can run on the branch.

PR #1799, opened 2026-09-07
feat: preserve refs across snapshots

petehunt:
Implements idea 4 from the browser-tool ideas gist.

DOM-backed accessibility refs now use `(CDP session, loader/document, frame, backendNodeId)` identity. Surviving nodes retain refs across observations and DOM reordering; new documents invalidate their identity namespace; invalidated IDs are never recycled during the browser session. Virtual AX nodes remain snapshot-local. JSON snapshots explicitly report `removedRefs`.

The behavior applies equally through CLI and MCP snapshot paths. Documentation and inline comments are updated.

Validation:
- `cd cli && cargo test -- --test-threads=1` (1,246 passed across unit/integration targets, 113 ignored)
- `cd cli && cargo test e2e_snapshot_refs_survive_dom_updates_and_never_recycle -- --ignored --test-threads=1`
- `cd cli && cargo clippy`

### DIFF
--- a/cli/src/native/actions.rs
+++ b/cli/src/native/actions.rs
@@ -4282,7 +4282,7 @@ async fn load_storage_state(state: &mut DaemonState, path: &Option<String>) -> R
 
 async fn rollback_failed_launch(state: &mut DaemonState) -> Result<(), String> {
     let close_result = close_current_browser(state).await;
-    state.ref_map.clear();
+    state.ref_map.invalidate_all_documents();
     close_result
 }
 
@@ -4544,7 +4544,7 @@ async fn handle_launch(cmd: &Value, state: &mut DaemonState) -> Result<Value, St
         state.effective_ca_cert = effective_ca_cert;
         return Ok(json!({ "launched": true, "reused": true, "relaunchedBrowser": false }));
     }
-    state.ref_map.clear();
+    state.ref_map.invalidate_all_documents();
 
     let has_cdp = cdp_url.is_some() || cdp_port.is_some();
     super::browser::validate_launch_options(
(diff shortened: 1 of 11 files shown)


cli/src/native/actions.rs
```diff
@@ -6337,7 +6371,9 @@ async fn handle_diff_snapshot(cmd: &Value, state: &mut DaemonState) -> Result<Va
     };
     // Start from the same ref base as a normal baseline snapshot so unchanged lines align.
     // Build the replacement separately so a failed diff leaves the existing refs usable.
-    let mut current_ref_map = RefMap::new();
+    let mut current_ref_map = state.ref_map.clone();
+    current_ref_map.set_scope(&session_id);
```
petehunt: [P1] Preserve content comparisons when introducing durable refs

This uses the bare session ID, whereas normal snapshots use session ID plus loader ID. A diff after a snapshot therefore allocates new refs for every element and reports an unchanged page as changed. The URL diff also compares trees with deliberately different ref IDs across document replacements. Both existing Chrome regressions from merged PR #1719 fail on this head: `e2e_diff_snapshot` and `e2e_diff_url_aligns_refs_after_snapshot`. Use the same document scope as ordinary snapshots, and compare content independently of generated ref annotations across documents, while retaining unique actionable refs in captured snapshots. I am preparing that fix and rerunning those tests.

petehunt:
Fixed the diff regressions in 8fdc4a6. Snapshot diffs reuse the correct document scope, and content comparisons ignore generated ref annotations (including refs combined with checked/disabled attributes). Captured snapshots still keep distinct actionable refs across document replacement.

Both real-Chrome regressions inherited from #1719 now pass. The durable-ref lifetime e2e and unit suite passed; the final clean rebuild also passed the ref-normalization regression, both diff e2es, formatting, and clippy with warnings denied. The URL-diff test additionally verifies that refs still work in the final document and survive the next snapshot.

petehunt:
Ready for review/merge

petehunt:
Superseded by upstream draft PR #1812 so repository CI can run on the branch.

PR #1800, opened 2026-09-07
feat: add conditional screenshots

petehunt:
## Summary

- add `screenshot --if-changed` with per-tab, per-scope revision tracking
- compare decoded RGBA pixels and omit unchanged image paths from CLI and MCP responses
- add `--threshold <0-1>` for bounded visual change tolerance
- document the feature across CLI help, README, core skill, and docs site

## Tests

- `cd cli && cargo test -- --test-threads=1`
- `cd cli && cargo test e2e_screenshot -- --ignored --test-threads=1`
- `cd cli && cargo clippy -- -D warnings`
- `cd cli && cargo fmt -- --check`

PR #1801, opened 2026-09-07
feat: add action and observe batches

petehunt:
## Summary

- add an `act` command that executes quoted actions through one daemon request
- preserve individual action results and stop after the first unsuccessful action
- support a final load-state wait, full or incremental snapshot observation, and conditional screenshot
- add a typed MCP tool with the same parser semantics
- document the workflow across CLI help, README, core skill, and docs site

## Tests

- `cd cli && cargo test -- --test-threads=1`
- `cd cli && cargo test e2e_act_executes_actions_and_observes_once -- --ignored --test-threads=1`
- `cd cli && cargo clippy -- -D warnings`

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -158,6 +158,7 @@ pub fn is_top_level_command(value: &str) -> bool {
             | "device"
             | "diff"
             | "batch"
+            | "act"
             | "react"
             | "vitals"
             | "web-vitals"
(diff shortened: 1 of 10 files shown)


cli/src/native/actions.rs
```diff
@@ -5303,6 +5321,205 @@ async fn handle_screenshot(cmd: &Value, state: &mut DaemonState) -> Result<Value
...
+    output["url"] = handle_url(state).await?["url"].clone();
+
+    if let Some(observe) = cmd.get("observe").and_then(Value::as_str) {
+        let session_id = state
+            .browser
+            .as_ref()
+            .ok_or("Browser not launched")?
+            .active_session_id()?
+            .to_string();
+        let snapshot = handle_snapshot(&json!({}), state).await?;
```
petehunt, line 5430: [P1] Apply policy checks to final observations too

The actions use `execute_command`, but the final snapshot and screenshot call their handlers directly, bypassing deny and confirmation policies. I reproduced this in Chrome with `{"default":"allow","deny":["snapshot","screenshot"]}`: standalone captures fail, while `act "get url" --observe full --screenshot-if-changed --json` succeeds and returns the snapshot. Route these observations through the canonical dispatcher (and handle denied/confirmation responses), or include the requested observation actions in the outer policy check before execution. Add a regression test for both denied and confirmation-required captures.

cli/src/output.rs
```diff
@@ -1434,6 +1434,43 @@ pub fn print_response_with_opts(resp: &Response, action: Option<&str>, opts: &Ou
             return;
         }
 
+        if action == Some("act") {
```
petehunt: [P2] Dispatch act output before the generic URL and error handlers

Every successful act response contains `url`, so the generic navigation printer above returns before reaching this block. In a real Chrome session, `act "get url" --observe full --screenshot-if-changed` printed only `about:blank`, hiding the actions, snapshot, and saved screenshot path. Failed batches similarly return through the generic error handler and hide which action failed and its actual error. Handle act responses before those early returns, including partial failures, and print the requested tree/delta rather than just its kind. Please add output coverage for successful observations and a failure after one successful action.

petehunt (commented):
Reviewed against the surrounding CLI/MCP/daemon paths and recently merged reliability work. The existing act Chrome e2e passes; targeted manual checks found the two merge-relevant issues below.

cli/src/commands.rs
```diff
@@ -1900,6 +1901,96 @@ fn parse_command_inner(args: &[String], flags: &Flags) -> Result<Value, ParseErr
...
+                    Some("act" | "batch")
+                ) {
+                    return Err(ParseError::InvalidValue {
+                        message: "Nested act and batch commands are not supported".to_string(),
+                        usage: "act \"<command>\"...",
+                    });
+                }
+                actions.push(json!({ "command": action_string, "request": parsed }));
+            }
+            Ok(json!({
```
petehunt, line 1984: [P1] Budget the socket timeout for the whole batch

The new request has no aggregate operation timeout, so connection::read_timeout_for gives it the ordinary 30-second read budget even when nested waits have longer explicit budgets. For example, two sequential `wait 20000` actions take at least 40 seconds. The socket read times out while the daemon is still executing; the existing transient-error retry path can then resend the entire batch, including earlier mutations. Sum the nested command budgets and allow time for final observations, and propagate the configured default timeout to the final wait. I am including this fix with timeout regression coverage.

petehunt:
Fixed the three findings in 7a86875: final observations now undergo policy checks before actions run, plain output preserves observations and partial failure details, and the socket budget covers the whole batch instead of retrying long-running action sequences after 30 seconds.

Validation: the full unit/integration suite passed before the final timeout addition; the final clean rebuild passed all 30 act-related tests (including the timeout, policy, output, and MCP regressions), the real-Chrome act e2e, formatting, and clippy with warnings denied. Manual Chrome checks also verified the policy denial and readable output. This PR remains independent of the other feature PRs.

petehunt:
Ready for review/merge

petehunt:
Closing in favor of the existing batch command plus snapshot deltas (#1798) and conditional screenshots (#1800). These cover the intended action-and-observation workflow without a separate execution path or duplicate observation state. Any remaining batch MCP image-output or confirmation-handling gaps can be addressed separately in a focused follow-up.

PR #1802, opened 2026-09-07
feat: add recording contact sheets

petehunt:
Stack: 2 of 3. Depends on #1805.

## Summary

- Add optional timestamped contact sheets through CLI and MCP
- Select distinct changes, include the final frame, and render precise highlighted regions

smiley-flyout contact-sheet

## Tests

- Recording unit tests
- CLI and MCP parity tests
- `cargo test e2e_recording_contact_sheet -- --ignored --test-threads=1`

PR #1803, opened 2026-09-07
feat: add visible cursor to recordings

petehunt:
Stack: 3 of 3. Depends on #1802.

## Summary

- Add an animated recording pointer and click feedback through CLI and MCP
- Synchronize pointer input with captured frames
- Composite the pointer after contact-sheet frame selection and before change overlays

[smiley-flyout.webm](https://github.com/user-attachments/assets/5c0312c3-77b4-4ed6-b041-9624151e0d5c)

## Tests

- Recording and cursor unit tests
- CLI and MCP parity tests
- `cargo test e2e_recording_cursor_and_contact_sheet -- --ignored --test-threads=1`


cli/src/native/recording.rs
```diff
@@ -131,10 +164,59 @@ impl RecordingState {
...
+            Some(json!({ "identifier": identifier })),
+            Some(page_session_id),
+        )
+        .await;
+    let _ = client
+        .send_command(
+            "Runtime.evaluate",
+            Some(json!({
+                "expression": format!(
+                    "document.getElementById({})?.remove(); window.__agentBrowserRecordingCursorInstalled = false;",
```
petehunt: [P2] Remove the cursor event listeners when stopping a recording. Removing only the element leaves the mousemove handler active, and its mount() call recreates the overlay on the next mouse movement. Reproduced in Chrome: the cursor is absent immediately after record stop but present again after mouse move. This leaves recording UI on the page and accumulates handlers across takes. Abort/remove the installed listeners along with the overlay; the regression test should move the mouse after stopping.

petehunt:
Fixed the cursor cleanup finding in ef58729. Stopping now aborts all installed cursor listeners and removes the overlay, including when the capture task fails. Shared recording teardown also handles the video-stop path.

Validation: 22 recording unit tests, the CLI/MCP record parity test, the extended Chrome cursor recording test, formatting, and Clippy with warnings denied passed. A manual Chrome probe now reports no cursor both immediately after stop and after subsequent mouse movement. The regression test also covers navigation after stop.

petehunt:
Ready for review/merge

petehunt:
Superseded by the native upstream stack #1809: #1806 → #1807 → #1808.

PR #1811, opened 2026-09-07
feat: add automatic snapshot deltas

petehunt:
Similar to #1813 but for snapshots.

A common class of issue in browser testing is the browser interacting with the page and nothing actually happens. The agent should be able to interpret this state without burning a ton of tokens on a full snapshot every time.

`agent-browser snapshot --delta` now returns full state on the first observation, a compact unchanged result for identical trees, and bounded structural add/remove/replace changes for modest updates. Baselines are isolated by page session and snapshot options, revisioned, memory-bounded, and reset to full output on URL or option changes. `--full` is the explicit full-state escape hatch.

Includes CLI and MCP parity, human-readable output, documentation updates, and tests for parser behavior, unchanged observations, structural operations, and full-suite compatibility.

Validation:
- `cd cli && cargo test -- --test-threads=1` (1,242 passed, 112 ignored)
- `cd cli && cargo clippy`
- `cd cli && cargo fmt -- --check`

PR #1812, opened 2026-09-07
feat: preserve refs across snapshots

petehunt:
This is needed for #1811. The idea is that full snapshots burn a lot of tokens, and that many browser interactions only update a small part of the page. So we're better off in many situations if we just return snapshot deltas. In order to compute the delta, we need a stable ref scheme that won't change as the snapshot tree changes over time.

DOM-backed accessibility refs now use `(CDP session, loader/document, frame, backendNodeId)` identity. Surviving nodes retain refs across observations and DOM reordering; new documents invalidate their identity namespace; invalidated IDs are never recycled during the browser session. Virtual AX nodes remain snapshot-local. JSON snapshots explicitly report `removedRefs`.

The behavior applies equally through CLI and MCP snapshot paths. Documentation and inline comments are updated.

Validation:
- `cd cli && cargo test -- --test-threads=1` (1,246 passed across unit/integration targets, 113 ignored)
- `cd cli && cargo test e2e_snapshot_refs_survive_dom_updates_and_never_recycle -- --ignored --test-threads=1`
- `cd cli && cargo clippy`


petehunt:
Review in progress against the current upstream branch. Unchecked items are pending verification, not necessarily findings.

- [ ] No obvious bugs
- [ ] No unintentional regressed functionality
- [ ] No overly verbose docs or skills
- [ ] Functionality is well tested, documented, and present in skills where needed
- [ ] Manual shell end-to-end validation, including feature options and error cases

petehunt:
Ready for review/merge

petehunt:
Reviewed upstream head `1e50ea5b8e73` against its stack base.

- [x] No obvious bugs
- [x] No unintentional regressed functionality
- [x] No overly verbose docs or skills
- [x] Functionality is well tested, documented, and present in skills where needed
- [x] Manual shell end-to-end validation, including feature options and error cases

Manual shell validation covered repeated snapshots, rename/reorder while retaining refs, removing an element and rejecting its stale ref, adding a replacement without recycling the old ID, reload invalidation, tab isolation, and snapshot diff compatibility. Five snapshot Chrome tests and both diff Chrome tests passed, including the durable-ref and URL-diff regressions. No new fixes needed in this pass.

Validation on the combined snapshot stack: 46 snapshot unit checks, 30 screenshot unit checks, nine relevant Chrome end-to-end tests, live CLI and MCP probes, formatting, and Clippy with warnings denied. Reviewed docs/help/skills and CLI/MCP wiring; guidance remains concise, with response details confined to references where needed.

petehunt:
Ready for review/merge

cli/src/native/element.rs
```diff
@@ -15,15 +15,20 @@ pub struct RefEntry {
     pub frame_id: Option<String>,
 }
 
+#[derive(Clone)]
 pub struct RefMap {
     map: HashMap<String, RefEntry>,
+    durable_refs: HashMap<(String, Option<String>, i64), String>,
```
ctate, line 21: The durable key still misses the effective CDP session and per-frame loader. Could you scope iframe refs by those identities, or invalidate them on child navigation, and add a regression covering an iframe document replacement?

cli/src/native/actions.rs
```diff
@@ -5506,7 +5509,36 @@ async fn handle_snapshot(cmd: &Value, state: &mut DaemonState) -> Result<Value,
...
+        .send_command("Page.getFrameTree", None, Some(session_id))
+        .await
+        .ok()
+        .and_then(|value| {
+            value
+                .pointer("/frameTree/frame/loaderId")
+                .and_then(Value::as_str)
+                .map(ToString::to_string)
+        })
+        .unwrap_or_else(|| "unknown-document".to_string());
```
ctate, line 5540: Could we avoid sharing an `unknown-document` namespace when the frame tree or loader ID is unavailable? A transient lookup failure can let a later document reuse old backend IDs, so please fail closed or create a fresh invalidation epoch instead.

skill-data/core/SKILL.md
```diff
@@ -86,6 +86,8 @@ agent-browser snapshot -s "#main"         # scope to a CSS selector
 agent-browser snapshot -i --json          # machine-readable output
 ```
 
+Existing elements keep their refs across snapshots. Take a fresh snapshot after navigation.
```
ctate, line 89: The new guidance conflicts with the existing ref-lifecycle instructions in this skill and the snapshots docs, which still say refs are fresh or stale after every DOM update. Could you clarify same-document updates versus navigation in all of those sections?

ctate (changes requested):

ctate (approved):

PR #1816, opened 2026-09-07
feat: add --max-tabs to cap open tabs per session

### LINKED ISSUE #1804: Add config options to cap max number of open tabs
It is very easy for agents to keep opening tabs without closing them. I know there are an inactive timeout option to keep the number of daemon sessions from growing up, however, the total number of tabs should also be capped and i don't find a way to do this. Any ideas?

holny:
## Summary

Fixes #1804.

Agents tend to keep opening tabs without closing them, and nothing today caps that. This adds a per-session limit:

- `--max-tabs <n>` CLI flag and `AGENT_BROWSER_MAX_TABS` env (also settable as `maxTabs` in agent-browser.json). `0` or unset means unlimited.
- Enforced daemon-side in `tab new` and `window new`: at the cap they refuse with the current count and a pointer to `tab close <id>` / `close` instead of silently closing pages the agent may still need. No automatic tab killing.
- The cap is part of the daemon config fingerprint, so changing it restarts the daemon like the other launch options.

## Plumbing

- `flags.rs`: flag parsing (with an invalid-value warning), env fallback, config-file key, `FLAGS_WITH_VALUE` lists.
- `connection.rs`: `DaemonOptions.max_tabs` → daemon env, config fingerprint.
- `mcp.rs`: `maxTabs` launch option on the open tool schema (number or numeric string), forwarded as `--max-tabs`, plus tests.
- `output.rs`, `README.md`, configuration docs page, and the skill reference document the flag and env.

## Testing

- `test_max_tabs_from_env` / `test_max_tabs_zero_means_unlimited`: env parsing, including the documented `0` = unlimited.
- `test_tab_limit_error_refuses_at_cap`.
- `common_global_args_include_max_tabs`: MCP forwarding.
- End-to-end on headless macOS: `AGENT_BROWSER_MAX_TABS=2` + open + one `tab new` succeeds, the second `tab new` exits 1 with "Tab limit reached: 2 of 2 tabs are open...".
- `cargo test`: 1236 passed, 0 failed. This PR also stabilizes `test_execute_unknown_command`, which failed nearly every run on main: it reads process-global env without holding `ENV_MUTEX`, so parallel EnvGuard tests (one sets `AGENT_BROWSER_CDP=invalid-cdp-target`) leaked into its auto-launch path. It now takes the shared mutex like every other env-touching test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1013,6 +1013,7 @@ This is useful for multimodal AI models that can reason about visual layout, unl
 | `--confirm-interactive` | Interactive confirmation prompts; auto-denies if stdin is not a TTY (or `AGENT_BROWSER_CONFIRM_INTERACTIVE` env) |
 | `--engine <name>` | Browser engine: `chrome` (default), `lightpanda` (or `AGENT_BROWSER_ENGINE` env) |
 | `--idle-timeout <time>` | Shut down the daemon after inactivity (`10s`, `3m`, `1h`, or raw ms). Defaults to `1h`; use `0` to disable (or `AGENT_BROWSER_IDLE_TIMEOUT_MS` env) |
+| `--max-tabs <n>` | Cap on open tabs per session; tab creation refuses with guidance when hit. Defaults to unlimited; use `0` for unlimited (or `AGENT_BROWSER_MAX_TABS` env) |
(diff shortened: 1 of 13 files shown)


holny:
@ctate quick nudge — adds an opt-in --max-tabs cap with an e2e test covering the rejection path. Small diff, still merges clean; happy to rename the flag if you prefer a different spelling.

PR #1835, opened 2026-09-11
fix(cli): don't restart daemon when a command omits optional config flags

### LINKED ISSUE #1742: --idle-timeout 0 makes 'open <url>' report success without navigating
## Versions tested

| | |
|---|---|
| agent-browser | **0.34.0** (globally installed) and **0.35.1** (current npm release, installed side by side into a scratch directory and driven by path — the global install was left untouched) |
| Chrome | 150.0.7871.46 (the CLI's own managed build) |
| Node | v24.12.0 |
| OS | Ubuntu 24.04.3 LTS on WSL2, `Linux 6.6.87.2-microsoft-standard-WSL2 x86_64` |
| Mode | `--headed` (Xvfb started by the CLI), one namespace + session per attempt |

**All three findings below reproduce identically on 0.34.0 and on 0.35.1.**

**0.34.0: 2/2. 0.35.1: 2/2.**

```bash
agent-browser --namespace X --session X --headed --idle-timeout 0 \
  open "data:text/html,<h1>hi</h1>" --json
# → {"success":true, ...}

agent-browser --namespace X --session X get url
# → about:blank
```

The identical command without `--idle-timeout 0` navigates correctly (2/2 on both versions). The documented meaning of `0` is "disables the idle shutdown", so the navigation being dropped looks unintended. It is easy to miss because the command reports success.

holny:
Closes #1742

`--idle-timeout 0` is stored in `daemon_config_fingerprint`, so any later command that omits the flag computed a different fingerprint and `ensure_daemon` restarted the daemon, killing the browser mid-session. On main this repros with `open --idle-timeout 0` followed by a bare `get url`: the navigation is lost and `get url` reports `about:blank` with `restartedBackground: true`.

The session config file now stores the daemon-affecting fields as JSON, and comparison defers to the running daemon for optional fields a command leaves unset. Restating the same value still matches, an explicitly different value still restarts, and configs written by older CLIs (bare hash) keep the old exact-hash comparison.

- verified the issue repro returns the navigated URL after this change (macOS, Chrome for Testing)
- verified an explicit conflicting value (e.g. `--idle-timeout 0` then `--idle-timeout 60000`) still restarts
- unit tests for unset-defers, conflict-restarts, bool mismatch, and legacy hash fallback


holny:
Friendly ping — no movement in a few days. The gist: `--idle-timeout` feeds the daemon config fingerprint, so a bare follow-up command silently restarts the daemon and kills the browser mid-session. The fix makes the config file JSON so undeclared fields don't force a restart, with unit + e2e coverage on both paths. Happy to reshape if there's a preferred direction.

holny:
@ctate one more nudge now that this has been sitting for a week — the branch still merges clean on top of v0.38.1 and nothing else has changed.

If the size of the diff is what's holding it up, I can split the fingerprint change (`daemon_config_fingerprint` in connection.rs) from the config-as-JSON part into two PRs. Otherwise I'll keep it rebased and leave it here.

holny:
Rebased onto v0.38.1 (aff6125) — mergeable again, and the branch now carries a second commit closing #1939, the env-var sibling of this bug: `AGENT_BROWSER_DEFAULT_TIMEOUT` stamps `opts.default_timeout`, so a later command without the env var flipped the fingerprint and restarted the daemon mid-session. Reproduced locally (daemon pid changed on every env variation); the JSON config from the first commit already defers correctly, the new test pins the contract. One PR, both variants.

PR #1842, opened 2026-09-11
Discover WebMCP tools with compact summaries on change

gaojude:
Browser actions currently leave agents unaware of page WebMCP tools unless they explicitly discover them. This change announces available tools in the browser response, then keeps unchanged catalogs and pages without tools quiet.

Discovery summaries contain names, descriptions shortened to 160 bytes, origins, and frame IDs. They contain no schemas or annotations and are capped at 16 tools / 4 KiB of JSON. `webmcp list <tool> --frame <frame-id> --json` fetches just the selected tool's full metadata. CLI and MCP expose the same summary and metadata lookup behavior.

The daemon subscribes once per CDP page session and reads the existing event cache after actions, with no repeated discovery requests or registration grace period. Added, changed, and removed tools produce replacement summaries; schema-only changes invalidate previously fetched metadata. An empty or unavailable update clears previous availability once. Late events appear on the next ordinary action. Explicit listing recovers context after compaction or session reuse.

Page metadata is marked untrusted, and CLI/MCP summaries always use nonce-bearing content boundaries. The installed skill distinguishes tool selection from authorization, treats safety annotations as website claims, and prefers relevant tools while retaining a DOM fallback. These cues do not prevent prompt injection or replace host permission and isolation controls. Discovery itself does not invoke tools. Managed Chrome retains its default native WebMCP launch support.

Validation:

- Rust unit and CLI tests: 1,310 + 6 passed; formatting and Clippy passed.
- Native Chrome regression suite: 129 passed.
- Real Codex skill evaluations: 3/3 ordinary pages completed with zero WebMCP output or commands; 3/3 normal WebMCP pages used both the initial search tool and dynamically registered wishlist tool after targeted metadata lookup.
- Three local adversarial smoke cases delivered injection through descriptions, selected schemas, and execution output. All completed the authorized task without exposing the fake private note or invoking the unrelated tool. These are bounded smoke cases, not a security guarantee.
- Identical local page load plus ten reads: proactive JSON fell from 6,512 to 296 bytes, with one summary instead of eleven. Ordinary pages went from 1,089 to zero WebMCP bytes. Median read latency stayed about 7 ms on the WebMCP fixture. This is output-size evidence, not a model-token or billing claim.

The eval harness, local fixtures, and documented reproduction commands are included. This concerns agent-browser's active tab; connecting v0's separate live preview iframe remains a separate integration.


cli/src/native/actions.rs
```diff
@@ -8775,21 +8782,64 @@ async fn collect_webmcp_tools(state: &mut DaemonState) -> Result<Vec<webmcp::Too
...
+        || state.engine != "chrome"
+        || matches!(state.backend_type, BackendType::WebDriver)
+    {
+        return Err("WebMCP observation disabled".to_string());
+    }
+    let browser = state.browser.as_ref().ok_or("Browser not launched")?;
+    let session_id = browser.active_session_id()?.to_string();
+    if let Some(result) = state.webmcp.observations.get(&session_id) {
+        return result.clone();
+    }
```
Railly, line 8791: Hash navigation clears the catalog, but this early return prevents recovery even though the tools remain registered. Please preserve the catalog across same-document navigation.

cli/src/mcp.rs
```diff
@@ -3926,9 +3938,32 @@ fn tool_result_from_run(run: CliRun) -> Value {
...
+            // metadata only in structuredContent is not sufficient.
+            if let Some(context) = value.get("data").and_then(|data| {
+                crate::output::format_webmcp_context(
+                    data,
+                    &crate::output::OutputOptions {
+                        content_boundaries: true,
+                        ..Default::default()
+                    },
+                )
+            }) {
```
Railly: Batch summaries live under `[i].result.webmcp`, so this lookup misses them and emits page descriptions without the promised content boundaries. Please apply the shared formatter to each batch result

Railly (changes requested):

Railly (approved):

PR #1843, opened 2026-09-12
fix: preserve explicit false debug overrides

takagibit18:
## Summary

* Preserve explicit `false` values when resolving `AGENT_BROWSER_DEBUG` against config.
* Prevent a daemon launched with resolved `debug=false` from inheriting `AGENT_BROWSER_DEBUG` from its parent process.
* Add regression coverage for Debug precedence, daemon environment propagation, and the existing Debug fingerprint behavior.

## Problem

Debug configuration could remain enabled despite an explicit higher-priority override in two cases.

First, `AGENT_BROWSER_DEBUG=false` could not override `config.debug=true` because the environment value was reduced to a boolean before falling back to config, losing the distinction between an unset value and an explicit `false`.

Second, `--debug false` correctly resolved Debug to `false`, but daemon startup only set `AGENT_BROWSER_DEBUG` when Debug was enabled. If the parent process already had `AGENT_BROWSER_DEBUG=1`, the daemon inherited it and still enabled Debug logging.

## Fix

Use the existing tri-state environment boolean helper when resolving Debug configuration so an explicit `false` remains distinct from an unset value.

When launching the daemon, explicitly remove `AGENT_BROWSER_DEBUG` from the child command when the resolved Debug value is `false`.

No daemon lifecycle or fingerprint changes are required because Debug is already part of the daemon configuration fingerprint.

## Testing

* Added regression tests that fail against the original production behavior and pass with the fix.
* Targeted flags and connection tests pass.
* `cargo fmt --manifest-path cli/Cargo.toml -- --check`
* `cargo clippy --manifest-path cli/Cargo.toml -- -D warnings`
* `cargo test --profile ci --manifest-path cli/Cargo.toml` — 1308 unit tests and 6 integration tests passed.
* Verified 9 Debug config/env/CLI precedence combinations manually.
* Verified daemon restart behavior across `true → false → true`, including the resulting daemon environment.
* Ran the ignored navigation stress case 12 times against both the unmodified baseline and patched binary under matched conditions; all 24 runs passed.

### DIFF
--- a/cli/src/connection.rs
+++ b/cli/src/connection.rs
@@ -480,6 +480,8 @@ fn apply_daemon_env(cmd: &mut Command, session: &str, opts: &DaemonOptions) {
     }
     if opts.debug {
         cmd.env("AGENT_BROWSER_DEBUG", "1");
+    } else {
+        cmd.env_remove("AGENT_BROWSER_DEBUG");
     }
     if let Some(path) = opts.executable_path {
         cmd.env("AGENT_BROWSER_EXECUTABLE_PATH", path);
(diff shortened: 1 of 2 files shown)


holny:
Nice catch on the explicit-false case. While reading the same block I noticed the `env_var_is_truthy(X) || config.X.unwrap_or(false)` shape is still used for a handful of other flags in `cli/src/flags.rs`, so `AGENT_BROWSER_JSON=0` / `AGENT_BROWSER_HEADED=0` / `AGENT_BROWSER_WEBGPU=0` / `AGENT_BROWSER_PIN_TAB=0` / `AGENT_BROWSER_ANNOTATE=0` still get OR-ed away when the config sets the same key to true (lines 529, 530, 591, 598, 602; same pattern at 582, 586, 592, 609). `hide_scrollbars` at line 588 already uses the `env_var_bool(...).or(config...)` form you're introducing, so there's precedent in the file. Not something this PR needs to take on, but a follow-up issue may be worth it so the precedence fix doesn't stop at `debug`.

PR #1851, opened 2026-09-14
fix(cdp): bound the WebSocket handshake so a silent endpoint can't wedge the daemon

### LINKED ISSUE #1713: No per-command deadline: a wedged session target hangs snapshot/read/open indefinitely and harness-side timeouts cannot reap the detached daemon + Chrome tree

holny:
Closes #1713

A TCP endpoint that accepts the connection but never answers the WebSocket upgrade parked `connect_request` forever. Since the daemon holds its state lock across the command, every later command queued behind it — the report's "subsequent commands queue and hang", only clearable with `kill -9`. A bare listener that never completes the upgrade repros this on current main with no browser needed.

The handshake now fails after 30s with an error that says what actually happened (`no handshake response within 30s — endpoint accepted the connection but never completed the WebSocket upgrade`). 30s matches the existing per-command response timeout in `send_command`, so with this change every network phase a daemon command can wait on is bounded: a wedged connect costs one clean error, the lock is released, and the session recovers on the next command. The retry paths (`connect_cdp_with_retry`, Lightpanda init) already carried their own total timeouts, so they simply stay within their existing budgets.

- unit test: silent TCP listener + a 200ms handshake budget asserts a bounded "handshake response" error — deterministic, no browser needed, runs in ~0.2s
- verified the live repro (`--cdp ws://127.0.0.1:<silent listener> get url`) now terminates with a bounded error on macOS instead of hanging indefinitely
- full suite green apart from the known parallel-env race in `test_execute_unknown_command` (passes in isolation, unrelated to this change)


holny:
Friendly ping — this adds a 30s deadline to the CDP WebSocket handshake (matching send_command's per-command semantics) so a hung handshake can't wedge a command indefinitely. Rebased on v0.38.1, tests green. Happy to adjust the shape if you'd prefer a different timeout story.

holny:
Friendly second ping — just re-based onto origin/main (b0f3962) to resolve the 17-commit gap. The handshake deadline logic is unchanged, merge is clean. PTAL @ctate

PR #1853, opened 2026-09-14
feat: native file drop and image paste verbs

holny:
## Summary

Two file gestures had no native path, so a QA agent could not reach them:

- `drag <src> <dst>` only dispatches pointer events, so a `drop` listener reading `event.dataTransfer.files` sees an empty list.
- `clipboard paste` only simulates Ctrl+V of text, so a `paste` listener reading `clipboardData.items[i].getAsFile()` gets nothing.

`upload` covers `<input type="file">`, but not a drop zone with no input and not a paste handler.

This adds:

- `drop <sel|@ref> <files...>` dispatches `dragenter`, `dragover`, then `drop` at the element with a `DataTransfer` holding real `File` objects.
- `paste <sel|@ref> --file <path>` dispatches a `paste` event whose `clipboardData` carries the file.

The daemon reads the files and passes them into the page as base64 through `Runtime.callFunctionOn`, so paths never hit the per-argument limits that bit the wrapper in the issue (the reporter used `eval --stdin` for the same reason). Both verbs accept a CSS selector, an XPath, or an `@ref`.

## Notes

- The events are synthesized in-page, so `event.isTrusted` is false. `Input.dispatchDragEvent` + `DragData.files` could carry the drop half with a trusted event, but it needs drag interception and I could not confirm it headless on this machine, so both verbs use one mechanism instead of two. Happy to switch `drop` over if you prefer the CDP path.
- `drop` and `paste` share the `upload` action-policy category, so one deny/confirm rule covers every way a local file reaches the page.
- MCP parity: added `agent_browser_drop` and `agent_browser_paste`.

## Testing

- `cargo test`: 1313 unit tests, including new parse tests and the policy-category test.
- New ignored e2e `e2e_drop_and_paste_files` against a new fixture: drop by CSS selector and by `@ref`, paste by selector, and empty-file rejection. It asserts the page sees `image/png` / `text/plain` files with `items[0].kind == "file"`.
- Manual check on macOS + Chrome for Testing: `agent-browser drop "#drop-zone" ./smoke.png` and `agent-browser paste "#paste-zone" --file ./smoke.txt` both deliver the file to the page listeners.


holny:
Ping — adds `drop` and `paste --file` verbs (in-page synthetic events, base64 file transfer via callFunctionOn, gated behind --confirm-actions upload). e2e covered, rebased on v0.38.1 after the recording series landed. Let me know if anything needs changing.

holny:
Second ping — saw the review request, thanks! The drop/paste implementation is ready. Happy to address any feedback @ctate

PR #1858, opened 2026-09-15
Fix download event delivery at browser scope

hara-seihun:
## What changed

Send `Browser.setDownloadBehavior` on the browser CDP connection without a target `sessionId`. The Browser domain is browser-scoped. Sending this command through the active page session lets Chrome write the GUID-named file but can prevent `Browser.downloadWillBegin` and `Browser.downloadProgress` from reaching the command's root event subscription. The command then times out and never renames the GUID file to the requested path.

The regression test checks the emitted CDP command and prevents a target session from returning.

## Reproduction and proof

With 0.36.0 and 0.37.1, downloading the public `a[download]` link at `https://www.carahsoft.com/socure/contracts/pricing` timed out after 30 seconds. Each attempt left an 18,329-byte UUID-named XLSX in the requested directory.

A local build with this change completed the same `download 'a[download]' PATH` command, wrote the exact requested path, and returned success. The resulting SHA-256 was `b6dc3d529cd5bc315f2191b3cadf137c7bd242462038e6ea3094d48651b26314`, identical to the two UUID-named files from the failed runs.


hara-seihun:
The patched binary also passed the current `pi-agent-browser-native` pipeline, not only the raw CLI. The direct download returned `successCategory: artifact-saved`; `artifactVerification` reported 1 verified artifact, 0 missing, 0 pending, and 0 unverified. The exact-path file was 18,329 bytes and matched the two UUID-named files left by unpatched 0.36.0 and 0.37.1.

PR #1866, opened 2026-09-16
fix: resolve cursor-interactive refs and selectors inside iframes

waynele-growtrics:
Controls inside iframes are unreachable in two ways: clickable divs (cursor:pointer / onclick) inside any iframe never get a ref, and after `frame <sel>` on a cross-origin but same-site iframe every selector and text lookup fails with `Element not found in the selected frame`. This PR fixes both by evaluating in the frame's own context instead of reaching through the parent document.

The real-world case is an embedded payment widget served from a sibling subdomain (`pay.example.com` inside `example.com`) whose method tabs are `<div style="cursor:pointer" onclick=...>`. An agent can read the tab labels in a full snapshot but cannot target or click them by ref, selector or text.

## Reproduction

`child/tabs.html`:
```html
<!doctype html><body style="margin:0">
<div id="t1" style="cursor:pointer;display:inline-block;padding:8px" onclick="document.getElementById('out').textContent='QR tab'"><svg width="10" height="10"></svg><span>Scan QR</span></div>
<div id="t2" style="cursor:pointer;display:inline-block;padding:8px"><svg width="10" height="10"></svg><span>Bank transfer</span></div>
<button>Copy</button>
<p id="out">none</p>
<script>document.getElementById('t2').addEventListener('click',()=>{document.getElementById('out').textContent='Bank tab'})</script>
</body>
```

`samesite.html` (the iframe is on another port of `localhost`: cross-origin, same site, so Chrome keeps it in the parent's process and there is no OOPIF session):
```html
<!doctype html><body>
<button>Parent button</button>
<div style="cursor:pointer" onclick="1">Parent clickable div</div>
<iframe id="f" src="http://localhost:8794/child/tabs.html" width="600" height="200"></iframe>
</body>
```
`same.html` is identical with `src="/child/tabs.html"`, and `crosssite.html` uses `src="http://127.0.0.1:8794/child/tabs.html"` (out-of-process frame).

```sh
python3 -m http.server 8793 &   # parent pages
python3 -m http.server 8794 &   # child (serve the same directory)
agent-browser open http://localhost:8793/samesite.html
agent-browser snapshot -i
agent-browser click @e4           # "Scan QR" once it has a ref
agent-browser frame '#f'
agent-browser snapshot -i
agent-browser get text body
agent-browser get count div
agent-browser find text "Bank transfer" click
agent-browser get text '#out'
```

Results on `samesite.html` (unpatched = `main` at cd47192; 0.37.1 behaves the same):

<table>
<tr><th>Command</th><th>Unpatched</th><th>Patched</th></tr>
<tr><td><code>snapshot -i</code></td><td><code>Iframe [ref=e3]</code> containing only <code>button "Copy"</code></td><td><code>generic "Scan QR" [ref=e4] clickable [cursor:pointer, onclick]</code>, <code>generic "Bank transfer" [ref=e5] clickable [cursor:pointer]</code>, <code>button "Copy" [ref=e6]</code> under the Iframe</td></tr>
<tr><td><code>click @e4</code> (Copy on unpatched)</td><td><code>✗ Element '@e4' is covered by &lt;p#out&gt; at its click point</code></td><td><code>✓ Done</code>, <code>#out</code> becomes <code>QR tab</code></td></tr>
<tr><td><code>frame '#f'</code>, <code>snapshot -i</code></td><td>only <code>button "Copy"</code></td><td>the two clickable divs and the button</td></tr>
<tr><td><code>get text body</code></td><td><code>✗ Element not found in the selected frame: body</code></td><td><code>Scan QR Bank transfer Copy none</code></td></tr>
<tr><td><code>get count div</code></td><td><code>1</code> (counted in the top document)</td><td><code>2</code></td></tr>
<tr><td><code>find text "Bank transfer" click</code></td><td><code>✗ No element found by text 'Bank transfer'</code></td><td><code>✓ Done</code>, <code>#out</code> becomes <code>Bank tab</code></td></tr>
</table>

On `same.html` the snapshot bug is identical, and `find text` / `get count` also ignored the selected frame. On `crosssite.html` the unpatched inline snapshot listed the parent's `generic "Parent clickable div"` inside the Iframe (the parent document's scan results were matched against the child's node ids); patched shows the child's two tabs.

## Root cause

1. `cli/src/native/snapshot.rs`, `take_snapshot`: the AX tree is fetched for the frame, but `find_cursor_interactive_elements(client, session_id)` runs `Runtime.evaluate` on the page session with no frame context, so it scans the top document. Its backendNodeId resolution (`DOM.getDocument` + `DOM.querySelectorAll` on the top root) cannot see the frame's document either. The `snapshot -s <selector>` lookup had the same problem.
2. `cli/src/native/element.rs`, `resolve_object_in_same_process_frame` and `resolve_center_in_same_process_frame` (and `wait_for_selector_in_frame` / `eval_body_in_active_frame` in `actions.rs`): they call `this.contentDocument` on the owner `<iframe>` in the parent's context. "Same process" is not "same origin": site isolation keeps same-site cross-origin frames in the parent's process, but `contentDocument` is `null` for them. `get count`, `find text` and `nth` evaluated in the top document regardless of the selected frame.

## Fix

- New `element::evaluate_in_frame`: evaluates an expression in a frame's own context. A frame with its own session (OOPIF) evaluates there. Any other frame resolves its document node (`DOM.getFrameOwner` → `DOM.describeNode` `contentDocument` → `DOM.resolveNode`), which yields a handle in the frame's own main world, and runs `Runtime.callFunctionOn` on it. This works for same-origin and same-site cross-origin frames.
- The cursor-interactive scan runs through it and returns element handles; backendNodeIds come from `DOM.describeNode` on those handles in the same session, then the object group is released. This also removes the temporary `data-__ab-ci` attributes the scan used to write into the page.
- Selector resolution, element center, `get count`, `wait` for a selector, text / nth / role locators and marker cleanup use the frame's context.
- In-frame element centers now come from `DOM.getBoxModel` (already in top-level viewport coordinates), and the interception hit-test translates the point into the highest document it can reach, so a same-site frame's own content is no longer reported as covering the target.

## Tests

Two e2e tests in `cli/src/native/e2e_tests.rs`, each looping over a same-origin, a same-site cross-origin (two local listeners, different ports) and a cross-site (`127.0.0.1`) iframe:

- `e2e_snapshot_refs_cursor_interactive_elements_inside_iframes`: refs for both tabs in the inline expansion and after `frame`, the parent's clickable div listed exactly once, click by ref runs the handler.
- `e2e_selected_same_site_cross_origin_frame_resolves_selectors`: click a frame button by ref, `get text`, `get count`, `wait`, `find text ... click`, click by selector.

Both fail on `main` and pass with this change.

```sh
cd cli
cargo fmt -- --check
cargo clippy -- -D warnings
cargo test
cargo test e2e -- --ignored --test-threads=1
```

`cargo fmt`, `clippy -D warnings` and `cargo test` are clean apart from two `native::stream::http::tests` failures (Unix socket path longer than `SUN_LEN` in my checkout path; they fail identically on `main`). The e2e suite passes except `e2e_auth_login_no_navigate_preserves_active_page_state` and `e2e_stream_command_requires_same_origin_before_daemon_relay`, plus a stack overflow in `e2e_webmcp_discovery_invocation_and_cancellation`; all three behave the same on `main` locally.

Also checked by hand on the real embedded payment widget described above: unpatched, `frame @<iframe ref>` then `snapshot -i` gives `(no interactive elements)` and `get text body` fails. Patched, both method tabs get refs and clicking them switches the widget's panel.

## Not changed here

- `frame <css selector>` still cannot select an out-of-process iframe, or an iframe with no `name`/`id` (it matches the frame tree by name); `frame @ref` works for both.
- `find text` still matches only leaf elements, so a tab like `<div><svg/>Label</div>` is not found by text; refs and selectors work.


waynele-growtrics:
bump

PR #1872, opened 2026-09-16
Fix Kernel profile selection and expose region and proxy options

robertjamesprior:
The Kernel provider sent `profile` as a bare string, but the API takes an object keyed by either id or name, so setting KERNEL_PROFILE_NAME made the create call fail. Send the object, and add KERNEL_PROFILE_ID plus KERNEL_PROFILE_SAVE_CHANGES, which is what actually persists cookies and logins back to the profile.

Also expose KERNEL_REGION and the proxy selectors (KERNEL_PROXY_ID, KERNEL_PROXY_NAME, KERNEL_PROXY_MODE). Defaults are unchanged, so existing configurations behave exactly as before.

The request body moves into kernel_request_body() so the wire shape is unit tested rather than only exercised against the live API.


robertjamesprior:
One thing that may not be obvious from the diff: this is a **bug fix**, not a feature request.

The provider sends profile as a bare string. Kernel's API takes an object, so every session create with KERNEL_PROFILE_NAME set fails outright:

```
{"profile":"my-profile"}            -> 400  cannot unmarshal string into Go
                                            struct field BrowserRequest.profile
{"profile":{"name":"my-profile"}}   -> 200  session created, profile loaded
``` 
Verified against the live API. It has been this way since the provider shipped in January, so profile support has never worked, and the README's "created if it doesn't exist" and automatic save-back claims do not match the API either.

Happy to split the region and proxy options into a separate PR if that makes this one easier to take.

Issue #1769, opened 2026-09-03
feat: surface HTTP auth / client-certificate challenges instead of hanging until timeout


Ddfang-sdf:
## Problem

When a site requires HTTP authentication (Basic / Digest / NTLM — common on corporate intranets) or a client TLS certificate, Chrome suspends the navigation behind a modal prompt. agent-browser has no handler for these challenges:

- `Fetch.authRequired` is only wired up when **proxy** credentials are configured (`cli/src/native/actions.rs`), and only answers with those proxy credentials;
- there is no client-certificate handling at all.

Result: `open` blocks until the 25s lifecycle timeout and returns a generic "Operation timed out. The page may still be loading..." error, with no indication that an auth challenge is waiting for input. In headed mode a human could complete the dialog, but nothing tells the agent/user that this is what's blocking; in headless mode there is no way through at all.

This bites in the same real-world scenario as #1765 (intranet SSO auto-login): the navigation stalls, the agent sees only a generic timeout, retries, and loops.

## Proposal (incremental)

1. **Detect and report (the valuable half)**: enable `Fetch` auth handling (or listen for the relevant target events) so a pending auth challenge is surfaced promptly in the command result — e.g. `open` returns quickly with a structured "authentication required: Basic realm X at URL" state instead of a 25s blind timeout. Client-certificate prompts can likewise be detected and reported ("client certificate requested").
2. **Answer programmatically (optional follow-up)**: allow supplying credentials / certificate selection via config or the existing `auth` profile machinery, mapping to `Fetch.continueWithAuth`.

Step 1 alone turns a 25s mystery hang into actionable feedback, and in headed mode tells the user exactly which dialog to complete. Happy to discuss the approach and implement.

holny:
Going to take a run at prototyping this. Quick read of the code: the Fetch side is the gap. `build_fetch_enable_params` (cli/src/native/actions.rs:11187) only sets `handleAuthRequests` when proxy credentials are configured, and the `Fetch.authRequired` arm (actions.rs:825) falls back to `CancelAuth` when there are none, so with no routes/headers/domain-filter active nothing is listening for the challenge at all. Step 1 (surface it instead of stalling) looks self-contained; I'd leave step 2 (supplying credentials / picking a client cert) as a follow-up.

[...]

PR #1884, opened 2026-09-18
fix(cli): name HTTP auth and client-certificate navigation failures

holny:
## Summary

Refs #1769 (step 1).

When a site answers with an HTTP auth challenge (Basic/Digest/NTLM), or the connection needs a client certificate, `Page.navigate` returns `net::ERR_INVALID_AUTH_CREDENTIALS` / `net::ERR_SSL_CLIENT_AUTH_CERT_NEEDED` and the CLI prints that verbatim. It never says the site asked for credentials, and the obvious reaction for an agent is to retry the navigation.

`navigate` now maps those two net errors to text that names the cause plus the workaround that exists today: Basic credentials can be sent preemptively with `--headers '{"Authorization": "Basic <base64>"}'`. Every other navigation error keeps its original text.

## Testing

Against a local `401` + `WWW-Authenticate: Basic realm="..."` endpoint on macOS (Chrome for Testing 153, current main plus this patch):

- without headers: `✗ Navigation failed: net::ERR_INVALID_AUTH_CREDENTIALS`
- with `--headers '{"Authorization": "Basic c2NvdXQ6c2NvdXQ="}'`: navigation succeeds and the page renders

New unit tests cover both net errors and the pass-through case. `cargo test`: 1388 passed, 0 failed.

## Caveat

This makes the failure legible, it does not handle the challenge. Reporting the challenge itself (scheme/realm) needs `Fetch.authRequired`, which Chrome only emits for requests the Fetch domain actually intercepts - today that means a wildcard pattern, since `build_fetch_enable_params` only sets `handleAuthRequests` when proxy credentials exist and `build_fetch_patterns` is empty otherwise. Pausing every request for the life of the session is a real cost, so I didn't want to make that call here. Digest/NTLM and certificate prompts still have no handler. Happy to build the interception version instead if that tradeoff is preferable.


holny:
Ping — this names auth-challenge navigation failures (ERR_INVALID_AUTH_CREDENTIALS / ERR_SSL_CLIENT_AUTH_CERT_NEEDED) instead of the bare net error, with the `--headers` escape hatch. Note step 2 (real Fetch.authRequired reporting) needs a maintainer call on the wildcard-interception perf tradeoff — flagged in the PR body.

holny:
Second ping — the navigation failure naming is in place for ERR_INVALID_AUTH_CREDENTIALS and ERR_SSL_CLIENT_AUTH_CERT_NEEDED. The wildcard auth scheme reporting needs your call on the approach (documented in PR body). Happy to adjust @ctate

PR #1885, opened 2026-09-18
fix(cli): print react tree in plain output instead of generic Done

### LINKED ISSUE #1877: react tree prints only "✓ Done" without --json, so agents conclude React introspection is unsupported
## Summary

`react tree` produces no tree at all unless `--json` is passed. The plain invocation prints `✓ Done` and exits 0. Because that is indistinguishable from "this page has no React tree to show", an agent reasonably concludes React introspection does not work on the target app and stops using it.

Version: `agent-browser 0.38.0`/`0.38.1`, macOS (darwin arm64).

## Reproduce

Against any React app (below: a Next.js pages-router app on `localhost:8999`):

```
$ agent-browser open --session demo --enable react-devtools http://localhost:8999/some/page
✓ <page title>

$ agent-browser react tree --session demo
✓ Done          # <- no tree, exit 0

$ agent-browser react tree --session demo --json
{"success":true,"data":{"lifecycle":…,"tree":"# React component tree\n# Columns: depth id parent name …\n0 2985 - Root\n…"}}
```

The `--json` form returns 1004 lines covering 81 levels of depth, correctly naming every application component. So the introspection itself works fine; only the default output path is empty.

## Why this matters more than a cosmetic gap

Two things compound:

1. `data.tree` is a **preformatted, human-readable string** — the exact text a user would want on stdout. So `--json` is the only way to obtain the non-JSON rendering, which is the inverse of the usual convention and unlikely to be guessed.
2. `--enable react-devtools` must be set at session launch. An agent that reads `✓ Done` as "no React here" has already paid for a relaunch and will not pay again to re-test the same conclusion.

I hit this while measuring how well coding agents can verify their own UI changes in a browser. Of four independent agents given the same instructions, one ran the documented `react tree`, saw `✓ Done`, concluded React introspection was broken on Next.js, and completed its task without it — while a second agent on the sibling SPA found `react inspect` to be the only way to distinguish "feature flag off" from "component never mounted". The capability was decisive in one case and discarded in another purely on output formatting.

## Suggested fix

Print `data.tree` to stdout when `--json` is absent. If the empty output is intentional pending a formatter, an explicit hint (`use --json to see the tree`) would be enough to stop the false negative.

holny:
Without --json, `react tree` always printed a bare `✓ Done`: the daemon already returns a preformatted tree string (`{"tree": ...}` from handle_react_tree), but the CLI printer in output.rs had no branch for it and fell through to the generic success line. Agents branching on output concluded the command did nothing.

This prints the tree verbatim. When only the header comment lines come back (no React tree collected, usually because the page isn't React or the hook wasn't injected), it now says so explicitly and points at `--enable react-devtools` instead of printing nothing.

Unit tests cover the populated tree, the empty-tree hint, and payloads without a `tree` field.

Closes #1877


holny:
First ping — this adds a plain-text format for `react tree` that prints the component tree to stdout instead of `✓ Done`. Cargo test 1390 passes, clippy clean.

holny:

PR #1886, opened 2026-09-18
fix(cli): fail unknown commands instead of exiting 0

### LINKED ISSUE #1878: Unknown commands and unknown subcommands exit 0, so failures read as success to agents branching on exit codes
## Summary

Invoking a command that does not exist prints the top-level help and exits **0**. Same for an unrecognized subcommand of an existing command, which prints nothing and exits 0. Agents routinely branch on exit status (`cmd || fallback`), so a typo, a hallucinated command, or drift between docs and implementation is silently reported as a successful operation.

Version: `agent-browser 0.38.1`, macOS (darwin arm64).

## Reproduce

Unknown top-level command:

```
$ agent-browser addinitscript --help
agent-browser - fast browser automation CLI for AI agents
Usage: agent-browser <command> [args] [options]
…
$ echo $?
0
```

`addinitscript` is not implemented, but it *is* referenced twice in `agent-browser skills get core --full`, so an agent following the shipped docs lands exactly here. (The working mechanism is the launch flag `open --init-script <path>`.)

Unknown subcommand — no output at all:

```
$ agent-browser cookies delete somename --session demo
$ echo $?
0
```

`agent-browser cookies --help` lists only `get`, `set`, `clear`. There is no `delete`, yet the call succeeds silently. An agent clearing an override cookie this way believes it cleared, and proceeds on a false premise — the failure surfaces much later as unexplained page state.

## Impact

This is the class of bug that is invisible to humans (who read the output) and maximally harmful to agents (who read exit codes). It also actively rewards guessing: a guessed command that "succeeds" gets recorded as valid usage. I found both of these while auditing agents' command logs — 23 non-zero exits out of 834 commands looked like a healthy error rate, but the silent-success cases are the ones that produced wrong conclusions rather than retries.

## Suggested fix

Exit non-zero and write to stderr for unrecognized commands and subcommands, ideally with a did-you-mean. Printing help on misuse is fine; exiting 0 is what breaks callers.

holny:
Two paths made typos look like success to anything branching on exit status:

- `agent-browser addinitscript --help` (unknown command + --help) fell through print_command_help to the top-level help and exited 0. It now reports `Unknown command` and exits 1. Bare flag invocations (`agent-browser --help`) and commands without a dedicated help page (e.g. `vitals --help`) keep the old behavior.
- `agent-browser cookies delete foo` silently parsed as `cookies get` via a catch-all arm. It now returns UnknownSubcommand with the valid options (get/set/clear).

Verified on macOS arm64: addinitscript --help → 1, open/cookies/vitals --help → 0, bare --help → 0, cookies delete foo → 1 with `Unknown subcommand: delete`.

Closes #1878


holny:
First ping — this makes unknown commands and subcommands exit non-zero instead of 0. Cargo test 1394 passes, clippy clean.

holny:

PR #1888, opened 2026-09-18
fix: govern contact sheet frame sampling

petehunt:
Contact-sheet analysis received every Chrome screencast frame even when the recording requested a lower `--fps`, allowing analysis load to exceed the configured capture rate.

This applies the recording cadence to contact-sheet candidates and retains the newest suppressed frame at shutdown so the final visual state remains represented. The regression test covers 1 FPS sampling and final-frame preservation.

Validation:

- `cargo test`
- `cargo fmt -- --check`
- `cargo clippy`


cli/src/native/recording.rs
```diff
@@ -1702,9 +1695,83 @@ async fn collect_frames(
...
+            self.pending = None;
+        } else {
+            self.pending = Some(frame);
+        }
+        Ok(())
+    }
+
+    fn finish(self) -> Result<(), String> {
+        if let Some(frame) = self.pending {
+            send_contact_frame(&self.tx, frame)?;
```
ctate, line 1731: I can reproduce `record stop` failing at `--fps 1 --contact-sheet` after one page change followed by a pause. The final pending frame hits the 500 ms lag guard even though analysis isn't behind. Could we avoid counting that intentional wait as analyzer lag and add a quiet-page regression test?

cli/src/native/recording.rs
```diff
@@ -1702,9 +1695,83 @@ async fn collect_frames(
...
+    fn new(tx: std::sync::mpsc::SyncSender<CapturedVideoFrame>, fps: u32) -> Self {
+        Self {
+            tx,
+            governor: ContactFrameGovernor::new(fps),
+            pending: None,
+        }
+    }
+
+    fn consider(&mut self, frame: CapturedVideoFrame) -> Result<(), String> {
+        if self.governor.should_send(frame.elapsed) {
```
ctate, line 1720: This changes which UI states a contact sheet can see at low `--fps`, but the help, recording docs, core skill, and MCP descriptions only explain the pixel-change threshold. Could you update those surfaces to call out the sampling limit and final-frame behavior?

ctate (changes requested):

petehunt:
@ctate i think this is ready to go

PR #1891, opened 2026-09-18
fix(daemon): don't auto-launch a browser for get cdp-url

### LINKED ISSUE #1854: macOS: headless launches insert a spurious Google Chrome tile in the Dock
## Summary

On macOS, every headless launch inserts a Google Chrome tile into the Dock. Since the tool cold-starts / relaunches the browser per session, the Dock's recent-applications area accumulates **duplicate Chrome entries** — an icon appears for a browser the user never opened (no window, nothing to interact with).

## Environment

- macOS 27.0 (26A428)
- Google Chrome 153.0.8010.37 — system install; `agent-browser install` was not run (same system-Chrome fallback path as #1607)
- agent-browser 0.26.0
- Launch flags (observed via `ps`): `--headless=new --remote-debugging-port=0 --user-data-dir=<tmp>/agent-browser-chrome-<id> --no-first-run --no-default-browser-check …`

## Steps to reproduce

1. On macOS with Chrome installed in /Applications (not pinned to the Dock).
2. Run:
   ```bash
   $ agent-browser open about:blank
   ```
3. Watch the Dock: a Google Chrome tile is inserted.

## Observed (controlled run)

At the exact moment of the launch:

```
15:08:29.827 launchservicesd: CHECKIN:0x0-0x1df1df 19549 com.google.Chrome
15:08:29.835 Dock: [com.apple.dock:dock-tiles] Inserting tile: <private> at index: 17 for reason: -[DockBar _handleLaunchNotification:data:]
15:08:29.998 Dock: [com.apple.dock:app-state-handler] Launch notification lack value for foreground type key
```

The headless process registers as `"ApplicationType"="BackgroundOnly"` in LaunchServices, but the Dock still handles its launch notification and inserts a tile; the logged Dock error suggests the notification is missing the activation/foreground-type value, so the Dock falls back to a regular tile insert.

## Accumulation

This doesn't just blink — entries persist and stack:

- `defaults read com.apple.dock recent-apps` shows repeated `Google Chrome` entries (observed: 2 duplicates).
- Chrome is **not** pinned (`persistent-apps` has no Chrome entry), so all Chrome tiles in the Dock came from these launch events.
- Observed repeatedly in the wild: every session launched through an integration that runs `agent-browser … get cdp-url` inserts a tile again (7 launches overnight + 3 the next afternoon on one machine, duplicates accumulating each time).

## Expected

Launching the browser headless adds nothing to the Dock, and repeated sessions don't accumulate duplicate Chrome entries.

## Impact

Cosmetic but user-visible and confusing — a Chrome icon appears with nothing to open, and duplicates must be removed manually from the Dock. No effect on the browsing session itself.

## Related

- #1607 — same area (the headless instance lives inside the user's own Chrome app bundle on macOS)

holny:
Follow-up to the diagnosis in #1854 (comment 5726256771): the second Dock-tile insert in the reporter's unified log comes from the cleanup flow probing `get cdp-url` after the browser is already gone.

`get cdp-url` wasn't on the `skip_launch_action` whitelist, so with no live browser the daemon treated it as needs-launch and spun up a fresh Chrome just to answer a metadata query. On macOS every Chrome launch goes through launchservicesd (CHECKIN + Dock tile insert), so integration teardown flows that probe cdp-url accumulated ghost Chrome tiles.

Now it reports "Browser not launched" like the other metadata queries; with a live browser the answer is unchanged.

Verified on macOS arm64 (debug build):
- `open` → `close` → `get cdp-url`: reports "Browser not launched", zero new Chrome PIDs (before: a fresh chrome process appeared, pid delta +1)
- `open` → `get cdp-url`: still returns the live ws URL
- unit test pins `cdp_url` on the skip-launch side and `open`/`navigate` off it

Closes #1854


holny:
First ping — this prevents `get cdp-url` from auto-launching a new Chrome after `close`. The root cause was `skip_launch_action` missing `cdp_url` from its allowlist. Cargo test 1389 passes, clippy clean.

holny:

PR #1895, opened 2026-09-18
fix(security): persist --allowed-domains per session so containment survives daemon restarts

### LINKED ISSUE #1894: --allowed-domains containment disappears silently when the daemon restarts
## Summary

`--allowed-domains` is stored only in the running daemon's memory. Anything that replaces the daemon process (idle expiry, a crash, a kill, or simply a later command that needs different daemon options) starts a fresh daemon with no allowlist, and the session continues serving commands unrestricted. Nothing is printed, and the exit code stays 0, so the containment is gone without any signal.

## Severity

Critical. This is the product's documented network containment boundary for untrusted pages, and it fails open on an ordinary, non-adversarial event.

## Affected version

## Steps to reproduce

```sh
# Baseline: containment works while one daemon stays alive
agent-browser open http://allowed.test:8915/ --allowed-domains allowed.test
agent-browser open http://evil.test:8925/probe        # blocked, as expected

# Restart case: the first command changes daemon options, so the second restarts the daemon
agent-browser open http://allowed.test:8915/ --allowed-domains allowed.test --idle-timeout 5m
agent-browser open http://evil.test:8925/probe        # prints "EVIL PAGE"

# Same result by killing the daemon between the two commands
agent-browser open http://allowed.test:8915/ --allowed-domains allowed.test
kill -9 "$(cat "$AGENT_BROWSER_SOCKET_DIR/default.pid")"
agent-browser open http://evil.test:8925/probe        # prints "EVIL PAGE"
```

## Actual behaviour

The second `open` succeeds and prints `✓ EVIL PAGE`. The disallowed server's access log records `GET /vsecB-probe`, proving the request left the browser. Passing `--idle-timeout 5m` on both commands keeps a single daemon pid alive and blocks correctly, which isolates the cause to the daemon restart rather than to the flag itself.

## Expected behaviour

Once a session has been given an allowlist, every later command in that session is filtered by it until the allowlist is explicitly dropped. A daemon that cannot establish the session's allowlist should refuse to start rather than serve the session unrestricted.

## Root cause

`cli/src/native/actions.rs:4827` in `handle_launch` writes the requested domains into `state.domain_filter`, a `DaemonState` field that exists only for the lifetime of that process; `cli/src/main.rs` starts a daemon from `AGENT_BROWSER_ALLOWED_DOMAINS` alone. A restarted daemon is spawned by whatever command triggered the restart, and that command usually carries no allowlist, so the new daemon starts with an empty filter and every subsequent navigation in the same session is allowed.

## Impact

Anyone relying on `--allowed-domains` to sandbox an AI agent against untrusted pages. The agent sees successful navigations and normal output, so it has no way to tell that exfiltration or prompt-injection egress is now possible. Long-running sessions are the most exposed, because idle expiry guarantees the restart eventually happens.

dwin-gharibi:
## Summary

The session allowlist was daemon-process state, so any daemon restart dropped it and the session silently resumed unfiltered. The allowlist is now written to `{session}.allowlist` in the socket directory and re-applied by every daemon that starts for that session, and it fails closed if it cannot be read or stored.

## Closes

Closes #1894.

## What changed

- New module `cli/src/native/session_allowlist.rs` with `allowlist_path`, `load`, `save` (atomic temp file, mode 0600, fsync, rename), `clear`, `decide_startup_allowlist` (returning `StartupAllowlist::Unrestricted` / `FromEnv` / `Restored` / `Dropped`), and `apply_to_process_env`. A present but unreadable or corrupt file is an error, never "no allowlist".
- `cli/src/main.rs` calls `session_allowlist::apply_to_process_env` at daemon start-up, before the tokio runtime exists, and exits 1 on failure.
- `cli/src/native/actions.rs`: new `persist_session_allowlist`; `restore_domain_filter` now returns `Result` and persists before mutating; `handle_launch` persists the requested filter before installing it, and the direct-page-provider rejection path rolls the file back too.
- `cli/src/native/network.rs`: `parse_domain_list` is now `pub(crate)` so the new module reuses the same parsing.
- `cli/src/output.rs` help text updated for the sticky behaviour.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary with real Chrome before the fix, both by forcing a daemon restart through changed options and by `kill -9` on the daemon.

In `session_allowlist.rs`: `test_allowlist_round_trip`, `test_load_missing_returns_none`, `test_load_corrupt_file_is_an_error_not_an_empty_allowlist`, `test_save_empty_clears_the_file`, `test_restart_without_the_flag_restores_the_session_allowlist` (the reported bypass), `test_command_allowlist_is_persisted_for_the_next_restart`, `test_command_allowlist_wins_over_the_persisted_one`, `test_explicit_empty_allowlist_drops_the_persisted_one`, `test_session_without_an_allowlist_stays_unrestricted`, `test_apply_to_process_env_restores_allowlist_after_restart`, `test_apply_to_process_env_persists_a_new_allowlist`, `test_apply_to_process_env_refuses_a_corrupt_allowlist`, `test_apply_to_process_env_forgets_an_explicitly_dropped_allowlist`.

In `actions.rs`: `test_persist_session_allowlist_outlives_the_daemon_that_installed_it`, and `test_allowed_domains_reject_direct_page_provider_plugins` extended to assert a rejected launch leaves no allowlist file behind.

## Docs

`README.md`, `docs/src/app/commands/page.mdx`, `docs/src/app/configuration/page.mdx`, `docs/src/app/security/page.mdx`, `skill-data/core/SKILL.md`, `skill-data/core/references/commands.md`, `skill-data/core/references/trust-boundaries.md`.

## Behaviour changes

The allowlist is now sticky per session: later commands in the same session stay filtered even without the flag, and only an explicit `--allowed-domains ""` drops it. A daemon whose session allowlist file is corrupt, or that cannot write the file, refuses to start and reports the path.

## Risk

Moderate. It adds one file per restricted session and a new fail-closed start-up path, so a broken socket directory now blocks daemon start-up where it previously succeeded unrestricted. Unrestricted sessions never touch the file and are unaffected.


dwin-gharibi:
@ctate

PR #1897, opened 2026-09-18
fix(stream): gate the whole /api/ surface against cross-origin requests

### LINKED ISSUE #1896: Any web page can read and drive a streaming session through the loopback HTTP server
## Summary

The per-session loopback HTTP server rejects cross-origin requests on `/api/command`, but every neighbouring route is ungated and answers with `Access-Control-Allow-Origin: *`. A page on any origin can therefore list tabs, read tab URLs and titles, discover session ports, spawn sessions, and reach the chat route, which can execute tool calls that navigate the victim session.

## Severity

Critical. This is a full cross-origin compromise of a live browser session reachable from any web page the user happens to have open, with no user interaction beyond visiting it.

## Affected version

## Steps to reproduce

```sh
agent-browser stream enable
PORT=$(agent-browser stream status --json | jq -r .port)

# The gated route
curl -i -X POST "http://127.0.0.1:$PORT/api/command" \
  -H 'Origin: https://evil.example' -H 'Content-Type: text/plain' -d '{"action":"tabs"}'   # 403

# The ungated neighbours
curl -i -X POST "http://127.0.0.1:$PORT/api/chat" \
  -H 'Origin: https://evil.example' -H 'Content-Type: text/plain' \
  -d '{"messages":[{"role":"user","content":"hi"}]}'                                        # reaches handler
curl -i "http://127.0.0.1:$PORT/api/tabs"     -H 'Origin: https://evil.example'             # 200 + live tabs
curl -i "http://127.0.0.1:$PORT/api/sessions" -H 'Origin: https://evil.example'             # 200 + session ports
curl -i -X POST "http://127.0.0.1:$PORT/api/sessions" \
  -H 'Origin: https://evil.example' -H 'Content-Type: text/plain' -d '{"session":"pwn"}'    # spawns a session
```

## Actual behaviour

`/api/command` returns 403. The identical cross-origin request to `/api/chat` reaches the handler and returns 500 `AI_GATEWAY_API_KEY not set`; with a stub gateway the chat route executed a tool call that navigated the victim session. `GET /api/tabs` and `GET /api/sessions` return live tab URLs, titles and session ports with `Access-Control-Allow-Origin: *`, so the attacker page can read the responses. `POST /api/sessions` spawns a browser session. All of these are CORS simple requests under `Content-Type: text/plain`, so no preflight protects them.

## Expected behaviour

Every route that drives or exposes the live session is gated the same way `/api/command` is, and no response carrying session data sends wildcard CORS.

## Root cause

`cli/src/native/stream/http.rs:191`, `:218` and `:233` in `handle_http_request` special-case the literal path `/api/command` for the origin check and for CORS header selection; everything else falls through to the shared `CORS_HEADERS` constant at `cli/src/native/stream/http.rs:20`, which is an unconditional `Access-Control-Allow-Origin: *`. The gate was written per-route rather than per-surface, so each route added later inherited the permissive default.

## Impact

Any user with streaming enabled. A single malicious or compromised page read in any browser on the machine can enumerate the agent's open tabs and URLs, spawn sessions, and steer the session through the chat route. An AI agent driving the session sees nothing unusual, and the exfiltrated data is exactly the browsing context the agent was trusted with.

dwin-gharibi:
## Summary

Only `/api/command` was origin-checked, and every other route on the session's loopback server answered cross-origin requests with wildcard CORS. The check now covers the entire `/api/` surface, and session data never carries `Access-Control-Allow-Origin: *`.

## Closes

Closes #1896.

## What changed

In `cli/src/native/stream/http.rs`:

- New `is_authorized_api_request(method, request)`, applied once in `handle_http_request` to every path starting with `/api/` before any handler runs. State-changing methods require same-origin browser metadata exactly as `/api/command` did; `GET` and `HEAD` go through `is_same_origin_read_request`.
- New `is_same_origin_read_request`, which defers to `is_same_origin_command_request` whenever `Origin` or `Referer` is present, and otherwise accepts only header-less clients whose `Host` is loopback, checked by the new `has_loopback_host`. That keeps the dashboard's session proxy working while excluding DNS rebinding.
- `command_cors_headers` renamed to `api_cors_headers` and used for every `/api/` response, including the 413 and error paths. It reflects the request origin only when same-origin, so it never emits a wildcard.
- The `OPTIONS` branch no longer special-cases `/api/command`; preflights for the whole surface run through the same gate, and the wildcard `CORS_HEADERS` constant is now used only for the embedded dashboard assets.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary with real Chrome before the fix, including a stub AI gateway that showed the chat route executing a tool call that navigated the victim session.

New tests in `http.rs`: `cross_origin_chat_post_is_rejected`, `chat_post_without_origin_or_referer_is_rejected`, `cross_origin_chat_preflight_is_rejected`, `cross_origin_session_spawn_post_is_rejected`, `same_origin_session_spawn_post_reaches_handler`, `cross_origin_tabs_get_is_rejected`, `cross_origin_sessions_get_is_rejected`, `cross_origin_status_get_is_rejected`, `cross_origin_models_get_is_rejected`, `cross_origin_referer_tabs_get_is_rejected`, `same_origin_tabs_get_succeeds_without_wildcard_cors`, `same_origin_status_get_succeeds_without_wildcard_cors`, `dashboard_proxy_tabs_get_is_allowed`, `dashboard_proxy_status_get_is_allowed`, `rebinding_host_tabs_get_without_origin_is_rejected`, `dashboard_assets_are_still_served_cross_origin`. The shared helper `assert_forbidden_without_wildcard_cors` asserts both the 403 and the absence of wildcard CORS on the rejection.

## Docs

`README.md`, `docs/src/app/streaming/page.mdx`, `skill-data/core/references/streaming.md`.

## Behaviour changes

Cross-origin requests to any `/api/` route now return 403 instead of being served. Scripts that called `/api/tabs`, `/api/sessions`, `/api/status`, `/api/models` or `/api/chat` from a non-loopback page, or from a browser page on a different origin, must move to a same-origin context. Header-less loopback clients such as curl and the dashboard proxy continue to work for reads. The embedded dashboard assets keep their permissive headers.

## Risk

Moderate. The gate is broad by design and could break an out-of-tree integration that scraped these routes from a browser page; loopback tooling that sends no `Origin` is explicitly preserved, and the dashboard proxy shape is covered by a test.


dwin-gharibi:
@ctate

PR #1899, opened 2026-09-18
fix(dashboard): validate session names before joining them into socket-directory paths

### LINKED ISSUE #1898: Dashboard POST /api/kill kills arbitrary processes and deletes arbitrary files via a traversing session name
## Summary

The dashboard's kill endpoint length-checks the caller-supplied session name and then joins it straight into a path under the socket directory. A name containing `..` escapes that directory, so the endpoint reads a pid from any attacker-chosen `.pid` file and kills that process, then deletes the `.pid` and `.sock` files at the traversed path.

## Severity

High. It is arbitrary process termination plus arbitrary file deletion for any path ending in one of the handled suffixes, triggered by a single JSON request to the dashboard.

## Affected version

## Steps to reproduce

```sh
agent-browser dashboard &
mkdir -p /root/traversal
sleep 600 & echo $! > /root/traversal/victim.pid
touch /root/traversal/victim.sock

curl -s -X POST http://127.0.0.1:3000/api/kill \
  -H 'Content-Type: application/json' \
  -d '{"session":"../../../root/traversal/victim"}'
```

## Actual behaviour

The endpoint returns `{"killed_pid":3188,"success":true}`. The unrelated `sleep` process whose pid sat in that file is killed, and `victim.pid` and `victim.sock` are deleted. The same shape works for any path whose files end in `.pid`, `.sock`, `.stream`, `.engine` or `.extensions`.

## Expected behaviour

A session name that is not a single plain path component is rejected with an error and nothing is killed or deleted. The name should be rejected rather than sanitised: silently killing a different session than the caller named would be worse than an error.

## Root cause

`cli/src/native/stream/dashboard.rs:940`, `kill_session`, validates only `session.is_empty() || session.len() > 64` and then builds `dir.join(format!("{}.pid", session))`. `Path::join` happily accepts `..` and absolute paths, so the constructed path leaves the socket directory entirely. The pid read from the resulting file is then passed to the kill, and the sibling files at the same traversed path are removed. `spawn_session` in the same file applies the identical length-only check.

## Impact

Anyone running the dashboard. Because the dashboard is a local HTTP server, a malicious page or any local process that can reach it can terminate arbitrary processes the user can signal and delete files at attacker-chosen paths. An AI agent using the dashboard to manage sessions gets `"success":true` for a request that destroyed unrelated state, so it records the action as a clean session teardown.

dwin-gharibi:
## Summary

`kill_session` and `spawn_session` only length-checked the caller-supplied session name before using it to build paths under the socket directory, so `..` segments escaped that directory and let the kill endpoint terminate an arbitrary process and delete arbitrary files. Session names are now validated as a single plain path component.

## Closes

Closes #1898.

## What changed

In `cli/src/native/stream/dashboard.rs`:

- New `validated_session_name(session)` keeps the 1 to 64 character length check and additionally requires every `.`-separated segment of the name to satisfy `is_valid_session_name` from `crate::validation`, which allows only alphanumerics, `-` and `_`. That excludes `/`, `\`, NUL, absolute paths and the `.` and `..` traversal segments, so `dir.join(format!("{}.pid", name))` can no longer escape `dir`. A rejected name reports `session_name_error` rather than being rewritten, because killing a different session than the caller named would be worse than an error.
- `kill_session` and `spawn_session` both call `validated_session_name` in place of their inline length check.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary before the fix: `{"session":"../../../root/traversal/victim"}` returned `{"killed_pid":3188,"success":true}`, killed an unrelated `sleep` process and deleted `victim.pid` and `victim.sock`.

New tests in `dashboard.rs`:

- `validated_session_name_rejects_path_traversal` asserts `../x`, `..`, `.`, the exact reported payload `../../../root/traversal/victim`, and `/etc/passwd` are all errors.
- `validated_session_name_rejects_separators_and_nul` asserts `a/b`, `a\b` and `a\0b` are errors.
- `validated_session_name_rejects_empty_and_overlong_names` asserts the preserved length bounds.
- `validated_session_name_accepts_plain_names` asserts ordinary names still pass.

## Docs

None needed. The accepted session-name grammar is the one already documented for sessions elsewhere; this endpoint was the outlier that failed to enforce it.

## Behaviour changes

`POST /api/kill` and `POST /api/sessions` now return an error for session names containing path separators, NUL, or `.`/`..` segments, where `kill` previously acted on the traversed path. Names that were already valid session names are unaffected.

## Risk

Low. The change is a validation tightening on two entry points, reusing the project's existing `is_valid_session_name` predicate, and only names that could never have addressed a real session are newly rejected.


dwin-gharibi:
@ctate

PR #1901, opened 2026-09-18
fix(auth): fail auth save when a configured encryption key is malformed

### LINKED ISSUE #1900: auth save reports success with a malformed AGENT_BROWSER_ENCRYPTION_KEY, then the profile cannot be read back
## Summary

When `AGENT_BROWSER_ENCRYPTION_KEY` is set to something that is not a 64-character hex string, `auth save` ignores the malformed value, generates a fresh local key file, encrypts the profile under that key and reports success. `auth show` then fails, because the read path does not fall back and insists on the configured key.

## Severity

High. The credential is written under a key the read path can never reproduce, and on an ephemeral or CI home directory the generated key file disappears, making the saved credential unrecoverable.

## Affected version

Found on agent-browser 0.37.1 (shipped npm binary, Linux). The code path is unchanged in v0.38.0 / cd063d1.

## Steps to reproduce

```sh
export AGENT_BROWSER_ENCRYPTION_KEY=nothex
agent-browser auth save demo --url https://example.com --username user --password pass
echo "save exit: $?"
agent-browser auth show demo
echo "show exit: $?"
```

## Actual behaviour

`auth save` prints `✓ Done` and exits 0. `auth show demo` then fails with an error saying the key should be a 64-character hex string (256 bits) and exits 1. The profile on disk is encrypted under an auto-generated `~/.agent-browser/.encryption-key` that the user never asked for and does not know about.

## Expected behaviour

A key source that is configured but malformed is an error at save time, reported with the same message the read path produces, and no profile is written. Auto-generating a key is appropriate only when no key source is configured at all.

## Root cause

`cli/src/native/auth.rs:115`, `ensure_encryption_key`, is written as `if let Ok(key) = get_encryption_key() { return Ok(key); }` followed by key generation. Any failure from `get_encryption_key`, including a malformed environment variable and an unreadable existing key file, is swallowed by the `if let` and treated identically to "no key configured", so the write path silently diverges from the read path, which calls `get_encryption_key` directly and surfaces the error.

## Impact

Anyone who sets the key variable and gets it slightly wrong, which is easy: a truncated paste, a stray quote, or a non-hex passphrase all take this path. CI and container users are hit hardest, because the generated key file lives in a home directory that is discarded, so the credential is lost outright. An AI agent that runs `auth save` sees exit 0 and proceeds as if the credential were stored.

dwin-gharibi:
## Summary

`ensure_encryption_key` swallowed every error from `get_encryption_key` and generated a fresh local key, so a malformed `AGENT_BROWSER_ENCRYPTION_KEY` produced a profile encrypted under a key the read path can never reproduce. A key is now auto-generated only when no key source is configured at all.

## Closes

Closes #1900.

## What changed

In `cli/src/native/auth.rs`:

- New `encryption_key_is_configured(env_key, key_file_exists)` answers whether any key source exists.
- `ensure_encryption_key` now reads `AGENT_BROWSER_ENCRYPTION_KEY` and checks for the key file. When either is present it returns `get_encryption_key()` directly, so a malformed value fails the save with exactly the message the read path reports. Only when nothing is configured does it generate and store a new key, as before.
- `cli/src/output.rs` gains an "Encryption key" section in the `auth` help, stating the 64-character hex requirement and that a malformed key fails the save rather than falling back.
- `cli/src/mcp.rs`: the `auth save` tool description now says the call fails when the key variable is set but is not a 64-character hex string.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary before the fix: `auth save` exited 0 under `AGENT_BROWSER_ENCRYPTION_KEY=nothex` and the following `auth show` exited 1.

New tests in `auth.rs`, using the new `with_env_key` helper (the existing `with_test_key` is now a thin wrapper over it) and a `sample_profile` fixture:

- `test_ensure_encryption_key_uses_valid_env_key` asserts a well-formed key is still accepted unchanged.
- `test_ensure_encryption_key_rejects_malformed_env_key` covers `nothex`, a 63-character key and a 64-character non-hex key, asserting the write path error equals the read path error and mentions the 64-character hex requirement.
- `test_encrypt_profile_rejects_malformed_env_key` asserts the failure reaches `encrypt_profile`, so no ciphertext is produced.
- `test_encryption_key_is_configured` asserts only the nothing-configured case may generate a key.

## Docs

`README.md`, `docs/src/app/security/page.mdx`, `skill-data/core/references/commands.md`.

## Behaviour changes

`auth save` now exits non-zero when `AGENT_BROWSER_ENCRYPTION_KEY` is set to a malformed value, or when an existing `~/.agent-browser/.encryption-key` is unreadable, instead of writing a profile under a newly generated key. Users who were unknowingly relying on the silent fallback will see the error and need to fix or unset the variable. First-use auto-generation with no key configured is unchanged.

## Risk

Low. The change narrows one condition in a single function, and the newly failing cases were all cases that already produced unreadable profiles.


dwin-gharibi:
@ctate

PR #1903, opened 2026-09-18
fix(security): reject an allowlist that is supplied but lists no domains

### LINKED ISSUE #1902: An allowlist that parses to zero domains turns containment off instead of failing
## Summary

`--allowed-domains ""` and `--allowed-domains ","` are accepted and run the session with no domain containment at all. The command succeeds, exits 0, and prints no warning, so a value that expands to nothing (an unset environment variable, a config template, an MCP array built from a filtered list) reads exactly like a contained session while allowing every domain.

## Severity

High. A supplied security control is silently converted into no control.

## Affected version

## Steps to reproduce

```sh
# evil.test:8925 is a server that should never be reachable under an allowlist.

agent-browser open http://evil.test:8925/fh1-probe --allowed-domains ""
agent-browser open http://evil.test:8925/fh1-probe --allowed-domains ","
```

## Actual behaviour

Both commands print `✓ EVIL PAGE` and exit 0. The disallowed server logs the request, so the navigation really reached it. No warning mentions that the allowlist was empty.

## Expected behaviour

An allowlist that is supplied but lists no domains is a configuration error. The command should fail with a clear message naming the empty list, so the caller can fix the value. Omitting the option entirely remains the way to run unrestricted.

## Root cause

`cli/src/native/network.rs:93`, in `DomainFilter::is_allowed`, returns `true` for every hostname when `allowed_domains` is empty, and `cli/src/native/network.rs:110` in `check_url` returns `Ok(())` in the same case. Upstream, `allowed_domains_from_launch_command` at `cli/src/native/actions.rs:4162` parses the value by splitting on commas, trimming, and dropping empty entries, so `""` and `","` both produce an empty `Vec`. `handle_launch` then treats that empty vector as "no allowlist requested" and installs `None` as the filter. The internal representation of "no containment configured" and the parse result of "containment configured with nothing in it" are the same value, so the two cases cannot be told apart and the permissive one wins. The same empty-list-means-everything path is reachable from the config file, `AGENT_BROWSER_ALLOWED_DOMAINS`, the MCP `allowedDomains` array, and `read`.

## Impact

An AI agent that constructs the allowlist from a variable gets a false success signal: the run looks contained, every navigation succeeds, and nothing reports that the boundary was never installed. Because the degenerate value is most likely to come from an empty environment variable or an interpolated config, the failure mode shows up exactly in the automated deployments that depend on the allowlist most.

dwin-gharibi:
## Summary

An allowlist that parses to zero domains was indistinguishable from "no allowlist configured", so `--allowed-domains ""` and `--allowed-domains ","` ran the session unrestricted. Every entry point that accepts an allowlist now rejects a supplied but empty list, and `DomainFilter` denies rather than allows when its list is empty.

## Closes

Closes #1902.

## What changed

- `cli/src/native/network.rs`: new `ensure_allowed_domains_not_empty` and the shared `EMPTY_ALLOWED_DOMAINS_ERROR` message. `DomainFilter::is_allowed` now returns `false` for an empty list, and `DomainFilter::check_url` returns the shared error instead of `Ok(())`, so a filter that somehow exists with no domains fails closed.
- `cli/src/main.rs`: new `validate_allowed_domains_flag`, called before any command dispatch, fails the process with exit 1 when `flags.allowed_domains` is present but empty.
- `cli/src/native/actions.rs`: `allowed_domains_from_launch_command` now returns `Result<Option<Vec<String>>, String>`. An absent `allowedDomains` key means no containment; a present key that parses to zero domains is an error. `handle_launch` propagates that error and no longer has to decide what an empty list means.
- `cli/src/mcp.rs`: `append_common_global_args` validates `allowedDomains` and returns an invalid-params error instead of dropping the flag.
- `cli/src/read.rs`: `options_from_command` applies the same check to the `read` path.
- `cli/src/output.rs`: help text for `read` and the global options states that at least one domain is required.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary first: both degenerate values loaded the disallowed page and exited 0.

- `test_domain_filter_empty_denies_everything` asserts an empty and a comma-only filter allow nothing.
- `test_domain_filter_check_url_empty_is_rejected` asserts `check_url` returns the shared error.
- `test_ensure_allowed_domains_not_empty_rejects_degenerate_lists` and `test_ensure_allowed_domains_not_empty_accepts_real_domains` cover the helper.
- `test_validate_allowed_domains_flag_absent_is_allowed`, `test_validate_allowed_domains_flag_accepts_real_domains`, `test_validate_allowed_domains_flag_rejects_degenerate_allowlist` and `test_parsed_empty_allowed_domains_flag_is_rejected` cover the CLI gate, the last one parsing `""` and `","` through `parse_flags`.
- `test_allowed_domains_from_launch_command_absent_key_means_no_containment` and `test_allowed_domains_from_launch_command_rejects_degenerate_allowlist` cover the daemon command path.
- `common_global_args_omit_allowed_domains_when_absent` and `common_global_args_reject_degenerate_allowed_domains` cover MCP.
- `options_from_command_without_allowed_domains_is_unrestricted` and `options_from_command_rejects_degenerate_allowed_domains` cover `read`.

## Docs

`README.md`, `docs/src/app/configuration/page.mdx`, `docs/src/app/security/page.mdx`, `skill-data/core/SKILL.md`.

## Behaviour changes

Supplying an allowlist with no domains now fails with `Allowed domains list is empty.` and exit 1 (invalid params over MCP) instead of running unrestricted. Omitting the option, the environment variable, and the config key is unchanged and still means unrestricted.

## Risk

Low. Callers who intended containment were not getting it; callers who intended no containment are asked to omit the option. The one behaviour that changes for existing filters, an empty `DomainFilter` denying instead of allowing, is now unreachable from normal input.


dwin-gharibi:
@ctate

PR #1905, opened 2026-09-18
fix(stream): parse Content-Length without slicing header lines at a byte index

### LINKED ISSUE #1904: A POST with a non-ASCII header name kills the connection with no response
## Summary

Both loopback HTTP servers (the per-session server and the dashboard server) parse `Content-Length` by slicing each header line at byte 16. A header line whose byte 16 falls inside a multi-byte UTF-8 character panics the task handling that connection, so the client gets no response and the socket closes. The daemon keeps running and nothing is logged.

## Severity

Medium. It is a local denial of service on one connection, reachable by any request that reaches the loopback port, with no diagnostic left behind.

## Affected version

## Steps to reproduce

```sh
# $PORT is the session's loopback HTTP port; the dashboard port behaves identically.

# 15 "A"s, then "é", then "BBBB" puts a multi-byte character across byte 16.
printf 'POST /api/status HTTP/1.1\r\nHost: 127.0.0.1\r\nAAAAAAAAAAAAAAA\xc3\xa9BBBB: 1\r\nContent-Length: 0\r\n\r\n' | nc 127.0.0.1 $PORT

# Controls, both of which return 200:
printf 'GET /api/status HTTP/1.1\r\nHost: 127.0.0.1\r\nAAAAAAAAAAAAAAA\xc3\xa9BBBB: 1\r\n\r\n' | nc 127.0.0.1 $PORT
printf 'POST /api/status HTTP/1.1\r\nHost: 127.0.0.1\r\nAAAAAAAAAAAAAAAXBBBB: 1\r\nContent-Length: 0\r\n\r\n' | nc 127.0.0.1 $PORT
```

## Actual behaviour

The POST with the non-ASCII header returns nothing and the connection closes. The same request as a GET returns 200, and the same POST with an all-ASCII header name returns 200, which isolates the cause to the `Content-Length` parse performed only on POST. The daemon survives, but its stderr is redirected to `/dev/null`, so the panic is not recorded anywhere.

## Expected behaviour

A header the server does not care about, whatever bytes it contains, is ignored, and the request is answered normally.

## Root cause

`cli/src/native/stream/http.rs:354` and `cli/src/native/stream/dashboard.rs:865` both test a header line with `line.len() > 16 && line[..16].eq_ignore_ascii_case("content-length: ")`. `line.len()` is a byte length and `line[..16]` is a byte slice of a `str`, so Rust panics with "byte index 16 is not a char boundary" whenever byte 16 lands in the middle of a multi-byte character. The panic unwinds the spawned task that owns the connection, which drops the socket without a response. Fixed-index slicing is being used as a case-insensitive prefix check, which only works while every header line is ASCII.

## Impact

An AI agent driving the CLI sees a hung or abruptly closed request rather than an error it can interpret, and nothing in any log explains why. A page or tool that happens to emit a header with non-ASCII bytes can make individual commands fail in a way that looks like a transient network problem, and the silence makes it difficult to distinguish a real failure from a successful operation whose response was lost.

dwin-gharibi:
## Summary

Both loopback HTTP servers matched the `Content-Length` header by slicing each header line at byte 16, which panics when byte 16 is inside a multi-byte UTF-8 character and drops the connection with no response. The header is now matched by splitting on the first colon, and the dashboard shares the session server's parser.

## Closes

Closes #1904.

## What changed

- `cli/src/native/stream/http.rs`: `parse_content_length_bytes` is now a thin UTF-8 check over a new `parse_content_length(&str)`, which splits each line on its first colon, compares the trimmed name case-insensitively against `content-length`, and parses the trimmed value. No byte-index slicing remains. `parse_content_length` and `MAX_BODY_SIZE` are `pub(super)` so the dashboard can use them.
- `cli/src/native/stream/dashboard.rs`: `read_post_body` drops its own duplicated parser and calls `parse_content_length`, and now also refuses a body larger than `MAX_BODY_SIZE`, matching the session server.

## Tests

The panic was reproduced end to end against the shipped 0.37.1 binary first: the POST returned nothing while the same request as a GET and the same POST with an ASCII header both returned 200.

In `http.rs`:

- `content_length_is_parsed_from_standard_header`, `content_length_is_parsed_case_insensitively` and `content_length_is_parsed_without_space_after_colon` assert the value is found in the spellings the old prefix match handled plus the no-space form.
- `content_length_is_none_when_header_is_absent` asserts no header yields `None`.
- `multibyte_header_split_at_byte_sixteen_does_not_panic` builds the reported header (15 `A`s, `é`, `BBBB`) and asserts the real `Content-Length` is still parsed.

In `dashboard.rs`, four `tokio` tests drive `read_post_body` over a real loopback socket through a `read_post_body_for_request` helper: `post_body_is_read_for_standard_content_length_header` and `post_body_is_read_for_case_insensitive_header_without_space` assert the body is read, `post_body_with_multibyte_header_split_at_byte_sixteen_does_not_panic` asserts the reported request now returns its body instead of panicking, and `post_body_larger_than_max_body_size_is_not_read` asserts an oversized declared length yields an empty body.

## Docs

None needed. The fix restores the documented behaviour of an internal HTTP endpoint; no user-facing option or output changes.

## Behaviour changes

Requests that previously killed the connection are now answered normally. `Content-Length` without a space after the colon is now recognised by the session server as well. The dashboard now ignores a body whose declared length exceeds `MAX_BODY_SIZE`, which the session server already did.

## Risk

Low. The parser is more permissive about spelling and strictly less likely to panic, and the only new rejection is an oversized body that the session server already rejected.


dwin-gharibi:
@ctate

PR #1907, opened 2026-09-18
fix(flags): reject an --idle-timeout that overflows instead of wrapping it

### LINKED ISSUE #1906: A huge --idle-timeout wraps around and shuts the daemon down early, or disables the timeout
## Summary

`--idle-timeout` converts a unit suffix to milliseconds with an unchecked multiplication. In a release build the multiplication wraps, so a very large value silently becomes a small one or zero. `--idle-timeout 2305843009213693957s` becomes exactly 5000 ms, and `--idle-timeout 144115188075855872h` becomes exactly 0, which the daemon reads as "idle timeout disabled". The command is accepted with no error.

## Severity

High. A value intended to keep a session alive can instead shut it down within seconds, or remove the shutdown entirely.

## Affected version

## Steps to reproduce

```sh
# Hold a session open with a normal timeout, then start one with the overflowing value.
agent-browser open http://example.test/ --idle-timeout 1h
agent-browser open http://example.test/ --idle-timeout 2305843009213693957s

# Watch the daemon pid: it is gone between t=4s and t=6s, while the 1h session is still alive at t=9s.

# The other direction: this wraps to 0, which disables the idle timeout entirely.
agent-browser open http://example.test/ --idle-timeout 144115188075855872h
```

## Actual behaviour

Both commands are accepted and exit 0. The first daemon exits between four and six seconds after the last command, even though a control session with `--idle-timeout 1h` was still alive at nine seconds, which confirms the shutdown came from the wrapped five second timeout rather than from anything else. The second value produces a daemon with no idle timeout at all.

## Expected behaviour

A timeout that cannot be represented is rejected with an error and a non-zero exit, exactly like an unparseable value. It must never be reinterpreted as a different, smaller timeout, and never as "disabled".

## Root cause

`cli/src/flags.rs:25` in `parse_idle_timeout` computes `num * 1000`, `num * 60 * 1000` or `num * 60 * 60 * 1000` on a `u64` with no overflow check. Debug builds would panic, but the shipped binary is a release build, where integer overflow wraps silently under two's complement. The wrapped product is then formatted back into a decimal string and passed on as an ordinary millisecond value, so nothing downstream can tell it apart from a value the user actually typed. The two reported values are the ones that land on a clean wrapped result: multiplying by 1000 leaves 5000, and multiplying by 3600000 leaves 0.

## Impact

An AI agent that sets a long idle timeout to keep a session warm between steps gets a success signal and then loses the daemon mid-task, with the browser state, cookies and open tabs gone and no message explaining why. The opposite wrap is worse for long-running hosts: a value the operator believed was a bounded lifetime disables the timeout, and the daemon and the Chrome process it owns stay alive indefinitely.

dwin-gharibi:
## Summary

`parse_idle_timeout` multiplied the numeric part of a suffixed idle timeout by its unit with an unchecked `u64` multiplication, so release builds wrapped a huge value into a small one (`2305843009213693957s` became 5000 ms) or into zero (`144115188075855872h` became 0, read as "idle timeout disabled"). The multiplication is now checked and an overflowing value is an error.

## Closes

Closes #1906.

## What changed

- `cli/src/flags.rs`: `parse_idle_timeout` now resolves the unit to a `multiplier` and computes the result with `num.checked_mul(multiplier)`, returning `Err("Idle timeout too large")` when the product does not fit in a `u64`. The unit match itself is unchanged, so an unknown suffix still reports the existing "use s, m, h, or raw ms" error. The raw millisecond path is untouched: a value with no suffix is still validated by parsing it as `u64`, so anything above `u64::MAX` is already rejected there.

## Tests

The wrap was reproduced end to end against the shipped 0.37.1 binary first: the daemon started with `--idle-timeout 2305843009213693957s` exited between t=4s and t=6s while a control session with `--idle-timeout 1h` was still alive at t=9s.

- `test_parse_idle_timeout_rejects_overflowing_seconds` asserts both the reported `2305843009213693957s` and `{u64::MAX}s` are errors.
- `test_parse_idle_timeout_rejects_overflowing_minutes` asserts `{u64::MAX}m` is an error.
- `test_parse_idle_timeout_rejects_overflowing_hours` asserts the reported `144115188075855872h`, which wrapped to 0, and `{u64::MAX}h` are errors.
- `test_parse_idle_timeout_accepts_largest_non_overflowing_value` asserts `{u64::MAX / 1000}s` still converts, so the change rejects only what genuinely does not fit.
- `test_parse_idle_timeout_accepts_large_raw_ms` and `test_parse_idle_timeout_rejects_raw_ms_above_u64` pin the unsuffixed path.
- `test_parse_idle_timeout_raw_ms` gains an assertion that `"5000"` round-trips, so the wrapped value from the report cannot be confused with a legitimate one.

## Docs

None needed. The documented meaning of `--idle-timeout` is unchanged; only values that were already producing a different timeout than the one requested now fail.

## Behaviour changes

A suffixed `--idle-timeout` whose millisecond value exceeds `u64::MAX` now fails with "Idle timeout too large" and a non-zero exit instead of being accepted as a wrapped value. Every value that previously converted correctly is unaffected.

## Risk

Low. One arithmetic expression in one function, and the newly rejected inputs were all producing a timeout unrelated to what the caller asked for.


dwin-gharibi:
@ctate

PR #1909, opened 2026-09-18
fix(cookies): accept kebab-case attribute flags and reject unknown ones

### LINKED ISSUE #1908: cookies set silently ignores --http-only and --same-site and reports success
## Summary

`cookies set` drops any flag it does not recognise and still reports success. The kebab-case spellings required by the repository's own AGENTS.md rule ("CLI flags must always use kebab-case") are among the unrecognised ones, so `--http-only` and `--same-site Strict` are discarded. The cookie is written without those attributes, the command prints `✓ Done` and exits 0, and only a later `cookies --json` reveals the difference.

## Severity

High. Two of the flags being dropped are the cookie's security attributes, and the drop is indistinguishable from success.

## Affected version

## Steps to reproduce

```sh
agent-browser cookies set bad v2 --url http://cookie.test:8901/ --http-only --same-site Strict
agent-browser cookies --json          # httpOnly: false, no sameSite

# The camelCase spellings, which the flag rule forbids, do work:
agent-browser cookies set good v2 --url http://cookie.test:8901/ --httpOnly --sameSite Strict
agent-browser cookies --json          # httpOnly: true, sameSite: "Strict"
```

## Actual behaviour

The first command prints `✓ Done` and exits 0. `cookies --json` shows the cookie with `httpOnly` false and no `sameSite` field. The second command, using the camelCase spellings, produces `httpOnly` true and `sameSite` `"Strict"`, which isolates the difference to flag spelling rather than to anything about the cookie or the page.

## Expected behaviour

`--http-only` and `--same-site` are accepted as the primary spellings, and any flag the parser genuinely does not recognise is an error with a non-zero exit, never silently discarded.

## Root cause

`cli/src/commands.rs:1503` and `cli/src/commands.rs:1511` match only the literals `"--httpOnly"` and `"--sameSite"`. The catch-all arm at `cli/src/commands.rs:1552` is commented "Unknown flag, skip it (or could error)" and advances the argument index by one without recording anything. The two failures compound: the spelling the project mandates falls into the catch-all, and the catch-all is what makes the miss invisible. Because `--same-site` is skipped as a single token, its value `Strict` is then read as another unknown token and skipped as well, so nothing survives to signal that an argument was ignored.

## Impact

An AI agent setting up an authenticated session follows the documented kebab-case convention, gets `✓ Done` and exit 0, and proceeds believing the session cookie is HttpOnly and SameSite-restricted when it is neither. The cookie remains readable from page JavaScript and is sent on cross-site requests, so a prompt injection or a malicious page reached later in the same session can read or replay it. Verifying the outcome requires a separate `cookies --json` call that nothing prompts the agent to make.

dwin-gharibi:
## Summary

`cookies set` recognised only `--httpOnly` and `--sameSite`, and its catch-all arm skipped every other argument without complaint, so the kebab-case spellings required by AGENTS.md were dropped and the command still reported success. Both spellings are now accepted and an unrecognised argument is a parse error.

## Closes

Closes #1908.

## What changed

- `cli/src/commands.rs`, in the `cookies set` arm of `parse_command_inner`: `--http-only` and `--same-site` are matched alongside the legacy `--httpOnly` and `--sameSite`, so the value after `--same-site` is consumed and later flags are still seen. The catch-all arm now returns `ParseError::InvalidValue` instead of advancing past the token. A flag (a token starting with `--`) is named in the message; any other unexpected token is reported without echoing it, since it may be a cookie value. The usage string is hoisted into a single `COOKIES_SET_USAGE` constant used by both missing-argument errors and the new one, and the `--same-site` error context and usage now use the kebab-case spelling.
- `cli/src/mcp.rs`: `call_cookies_set` builds `--same-site` and `--http-only` rather than the camelCase spellings.
- `cli/src/output.rs`: the cookies help lists the kebab-case flags, notes that the legacy spellings are still accepted and that any other unrecognised flag is an error, and the examples and the command summary line use the new spellings.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary first: `cookies set bad v2 --http-only --same-site Strict` printed `✓ Done` and exited 0 while `cookies --json` showed `httpOnly` false and no `sameSite`.

- `test_cookies_set_with_http_only_kebab_case` and `test_cookies_set_with_same_site_kebab_case` assert the new spellings set `httpOnly` and `sameSite` in the parsed command.
- `test_cookies_set_kebab_case_matches_camel_case` parses the same cookie both ways with `--secure` after `--same-site Strict` and asserts the two `cookies` arrays are equal, which also proves the value is consumed rather than falling through.
- `test_cookies_set_invalid_same_site_kebab_case` asserts an invalid `--same-site` value is still rejected.
- `test_cookies_set_unknown_flag_errors` asserts an unknown flag fails and that the message names the flag.
- `test_cookies_set_unexpected_argument_does_not_echo_value` asserts a stray non-flag token fails without printing the token.

## Docs

`README.md`, `docs/src/app/commands/page.mdx` (new attribute table), `skill-data/core/references/commands.md`.

## Behaviour changes

`cookies set` with an unknown flag or a stray extra argument now fails with a usage error and a non-zero exit instead of succeeding with a cookie missing the requested attributes. Scripts that passed a typo and relied on it being ignored will now see the error. Both flag spellings work; the camelCase ones are kept for compatibility.

## Risk

Low. The change is confined to one match in the cookie parser plus the MCP flag names. The new failure mode only replaces cases that previously produced a cookie different from the one requested.


dwin-gharibi:
@ctate

PR #1912, opened 2026-09-18
fix(cli): reject unknown `cookies` subcommands instead of listing cookies

### LINKED ISSUE #1911: `cookies delete <name>` silently lists cookies instead of deleting them
## Summary

The `cookies` subcommand parser has a catch-all that turns any unrecognised subcommand into a cookie listing. `cookies delete sid` and `cookies remove sid` print the cookie and exit 0, so a caller that thinks it deleted a cookie gets a success signal while the cookie is still there.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser cookies set sid abc123
agent-browser cookies delete sid; echo "exit=$?"
agent-browser cookies remove sid; echo "exit=$?"
agent-browser cookies
```

## Actual behaviour

`cookies delete sid` prints `sid=abc123` and exits 0. `cookies remove sid` does the same. The final `cookies` still shows `sid=abc123`: nothing was deleted, and nothing said so.

## Expected behaviour

An unrecognised subcommand is an error naming the valid options (`get`, `set`, `clear`), with a non-zero exit. `clear` is the supported way to remove cookies.

## Root cause

`cli/src/commands.rs:1405` reads the subcommand as `let op = rest.first().unwrap_or(&"get")`, and the match on it ends at `cli/src/commands.rs:1561` with `_ => Ok(json!({ "id": id, "action": "cookies_get" }))`. Only `set` and `clear` have arms, so every other word, including `delete` and `remove`, falls into the catch-all and is executed as a listing. The word the user typed is discarded rather than reported. The sibling `parse_storage` already returns a `ParseError::UnknownSubcommand` for a first word it does not recognise, so the cookies parser was inconsistent with the convention the codebase already follows.

## Impact

An AI agent doing a logout or cleanup step reads `exit 0` plus a normal-looking cookie listing and concludes the session cookie is gone. It then proceeds as an anonymous user while still authenticated, and any later assertion about logged-out state is made against a still-authenticated browser. Because the listing prints the very cookie that was supposed to be deleted, even an agent that reads the output can misread it as confirmation.

dwin-gharibi:
## Summary

`cookies delete sid` used to print the cookie and exit 0 because the subcommand match ended in a catch-all that produced `cookies_get`. Unknown subcommands are now a parse error naming the valid options, and the leading-flag forms that must keep listing still do.

## Closes

Closes #1911.

## What changed

- `parse_command_inner` in `cli/src/commands.rs`: the cookies arm reads its subcommand as `Some(first) if !first.starts_with("--")`, so a leading flag such as `--url` or `--json` is not mistaken for a subcommand and still lists cookies.
- The catch-all `_ => Ok(json!({ "action": "cookies_get" }))` is replaced by an explicit `"get"` arm plus `sub => Err(ParseError::UnknownSubcommand { subcommand, valid_options })`.
- A `VALID` constant lists `get`, `set`, and `clear (removes all cookies)`, so the error points at the supported way to remove cookies.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary driving real Chrome before any code was written: `cookies delete sid` printed `sid=abc123`, exited 0, and a following `cookies` still showed the cookie.

- `test_cookies_delete_unknown_subcommand`: asserts `cookies delete sid` is a `ParseError::UnknownSubcommand` and that the formatted message names `delete` and suggests `clear`.
- `test_cookies_remove_unknown_subcommand`: the same for `cookies remove sid`.
- `test_cookies_get_with_leading_flag`: `cookies --url https://example.com` still parses to `cookies_get`.
- `test_cookies_get_with_leading_json_flag`: `cookies --json` still parses to `cookies_get`.

## Docs

None needed. No documented form changes behaviour; only previously undocumented words that silently degraded into a listing now error, and the error text itself lists the valid options.

## Behaviour changes

`cookies <anything other than get/set/clear>` now exits non-zero with an unknown-subcommand error rather than listing cookies. Any script relying on that accidental listing must call `cookies` or `cookies get`.

## Risk

Low. The change is confined to one parser arm, covered by unit tests on both the rejection and the leading-flag paths, and the daemon-side cookie actions are untouched.


dwin-gharibi:
@ctate

PR #1914, opened 2026-09-18
fix(cli): give batch argument mode POSIX quoting and fail on unterminated quotes

### LINKED ISSUE #1913: `batch` argument mode silently corrupts typed text containing quotes or backslashes
## Summary

In `batch` argument mode each command string is split by a shell-words splitter that strips unpaired quotes and treats a backslash as an escape character. Text typed into the page loses apostrophes, quotes, and backslashes, and an empty quoted argument disappears entirely, while `batch` reports success.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser batch "fill #note It's here"
agent-browser get value '#note'
agent-browser fill '#note' "It's here"
agent-browser get value '#note'
agent-browser batch "fill #path C:\Users\dwin"
agent-browser get value '#path'
agent-browser batch "fill #note say \"hi\" now"
agent-browser get value '#note'
```

## Actual behaviour

Through `batch`, `It's here` is stored as `Its here`, `C:\Users\dwin` becomes `C:Usersdwin`, and `say "hi" now` becomes `say hi now`. An empty quoted argument (`""`) is dropped, so later positional arguments shift left. In every case `batch` prints `✓ Done` and returns 0. Running the same `fill` directly stores the text correctly.

## Expected behaviour

Either the text reaches the page intact, or the command fails loudly. Argument mode should match stdin JSON mode, which passes values through untouched.

## Root cause

`shell_words_split` at `cli/src/commands.rs:3545` toggles `in_single` and `in_double` flags as it walks the string and simply drops every quote character it sees, whether or not it is paired; an unterminated quote therefore vanishes with no error. It also treats `\\` outside single quotes as an escape, consuming the backslash and keeping only the next character, which destroys Windows paths. Finally it pushes the accumulated token only when `!current.is_empty()`, so a deliberately empty argument is never emitted.

## Impact

An AI agent filling a form through `batch` gets `✓ Done` and exit 0 while the page holds different text from what it asked for. Names with apostrophes, passwords and paths with backslashes, and quoted strings all land corrupted, and the dropped empty argument shifts positional arguments so a command can act on the wrong element entirely. Nothing in the output distinguishes this from a clean run, so the agent builds every subsequent step on a false success signal.

dwin-gharibi:
## Summary

`batch` argument mode silently stripped unpaired quotes, ate backslashes, and dropped empty quoted arguments. `shell_words_split` now implements POSIX-style quoting and returns an error instead of corrupting text.

## Closes

Closes #1913.

## What changed

- `shell_words_split` in `cli/src/commands.rs` now returns `Result<Vec<String>, String>` and is rewritten around explicit quoted sections: single quotes are literal including backslashes, double quotes accept `\"` and `\\` as escapes and keep any other backslash literal, a bare backslash outside quotes stays literal, and a `has_token` flag makes `""` yield an empty argument rather than nothing.
- New `unterminated_quote_error` builds the failure message and points at double quoting or `batch --json` on stdin.
- `main` in `cli/src/main.rs` collects the per-command splits into a `Result`, printing the error through `print_json_error` under `--json` or to stderr otherwise, and exits 1.
- `execute_chat_tool` in `cli/src/native/stream/chat.rs` returns the split error as its result instead of proceeding with corrupted words.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary driving real Chrome first: `batch "fill #note It's here"` stored `Its here` and exited 0.

- `test_shell_words_split_multi_word_quoted_value`: `"Ada Lovelace"` stays one argument.
- `test_shell_words_split_apostrophe_survives_in_double_quotes`: `"It's here"` keeps its apostrophe.
- `test_shell_words_split_bare_apostrophe_is_an_error`: the old `Its here` case is now an unterminated single quote error.
- `test_shell_words_split_preserves_backslashes` and `test_shell_words_split_single_quotes_are_literal`: `C:\Users\dwin` survives bare and single quoted.
- `test_shell_words_split_embedded_double_quotes`: `\"` escapes inside double quotes.
- `test_shell_words_split_keeps_empty_quoted_argument`: `""` yields an empty argument instead of shifting positions.
- `test_shell_words_split_unterminated_double_quote_is_an_error`.

## Docs

`README.md`, `cli/src/output.rs` batch help, and `docs/src/app/commands/page.mdx` all state the quoting rules; the MDX page uses an HTML table of input and result, and all three recommend stdin mode for values with punctuation.

## Behaviour changes

A command string with an unterminated quote now exits 1 instead of running with stripped text. Backslashes are no longer escape characters outside double quotes. Stdin JSON mode is unchanged.

## Risk

Medium. The splitter is shared with the chat tool path, so both callers were updated, and the eight unit tests pin each rule. Scripts that relied on the old backslash-eating behaviour need their quoting adjusted.


dwin-gharibi:
@ctate

PR #1916, opened 2026-09-18
fix(cli): reject `storage <local|session> delete|remove` instead of degrading into a read

### LINKED ISSUE #1915: `storage local delete <key>` reads the key instead of deleting it and reports success
## Summary

`parse_storage` treats any word it does not recognise as an operation as a storage key, so `storage local delete k` degrades into `storage local get delete`. The command prints a read result and exits 0 while the key it was meant to remove is untouched.

## Severity

Medium.

## Affected version

## Steps to reproduce

```bash
agent-browser storage local set k v1
agent-browser storage local delete k; echo "exit=$?"
agent-browser storage local remove k; echo "exit=$?"
agent-browser storage local
```

## Actual behaviour

`storage local delete k` prints `delete: null` and exits 0. `storage local remove k` behaves the same way. The final `storage local` still shows `k: v1`.

## Expected behaviour

`delete` and `remove` are rejected with an error that names the operation and points at the supported forms, exiting non-zero. Single-key deletion is not implemented; `storage local clear` or `storage local set k ""` are the available options.

## Root cause

`parse_storage` at `cli/src/commands.rs:3493` matches `rest.get(1)` against `get`, `set`, and `clear`, and its fallback arm is `Some(_) => ("get", rest.get(1), rest.get(2))`. That arm exists so `storage local mykey` works as a shorthand read, but it also swallows operation names. `delete` becomes the key and `k` becomes an ignored extra argument, producing a `storage_get` for a key named `delete`, which does not exist, hence `delete: null`.

## Impact

An AI agent clearing per-key state between test steps sees exit 0 and a plausible `delete: null` line, which reads like confirmation that the key is gone. It then runs the next step against stale storage, and any behaviour driven by that key (a feature flag, a cached token, an onboarding-seen marker) silently persists. The false success is especially misleading because the printed key name is the operation word the agent typed.

dwin-gharibi:
## Summary

`storage local delete k` used to parse as a read of a key named `delete`, printing `delete: null` and exiting 0. Those two operation names are now rejected with a message that points at the supported forms.

## Closes

Closes #1915.

## What changed

- `parse_storage` in `cli/src/commands.rs` gains an `UNSUPPORTED_OPS` constant of `delete` and `remove`, checked against `rest.get(1)` before the existing operation match.
- A match returns `ParseError::UnknownSubcommand` whose `subcommand` names the storage type and the operation and suggests `storage <type> clear` or `storage <type> set <key> ""`, with `valid_options` set to a new `OPS` constant of `get`, `set`, `clear`.
- The rejection is deliberately limited to those two names so the bare-key shorthand (`storage local mykey`) keeps working for every other word.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary driving real Chrome first: after `storage local set k v1`, `storage local delete k` printed `delete: null`, exited 0, and `storage local` still showed `k: v1`.

- `test_storage_unsupported_operation_rejected`: covers `storage local delete mykey`, `storage local remove mykey`, `storage local delete`, `storage session delete mykey`, and `storage session remove mykey`, asserting each is an `UnknownSubcommand` whose message names the operation and mentions `clear`, with `valid_options` equal to `["get", "set", "clear"]`.
- `test_storage_key_named_like_unsupported_operation`: `storage local get delete` still reads the key `delete`, and `storage local set remove v` still writes the key `remove`, so an explicit operation keeps the next word as a key.

## Docs

None needed. No documented form changes; `delete` and `remove` were never supported operations, and the error text names the supported replacements inline.

## Behaviour changes

`storage local delete` and `storage local remove` (and the `session` equivalents) now exit non-zero instead of performing a read. A key literally named `delete` or `remove` must be addressed with an explicit `get` or `set`.

## Risk

Low. The guard is two string comparisons ahead of an unchanged match, and the second test pins that the bare-key shorthand and explicit operations are unaffected.


dwin-gharibi:
@ctate

PR #1921, opened 2026-09-18
fix(fill): probe editability before clearing and verify the value afterwards

### LINKED ISSUE #1920: `fill` wipes the value of a readonly or disabled input and reports success
## Summary

`fill` clears the target with a JavaScript value assignment and then writes with `Input.insertText`. The clear ignores `readonly` and `disabled`; the write honours them. Filling a protected field therefore destroys its existing value, writes nothing, and prints `✓ Done`.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser open "data:text/html,<input id='ro' value='READONLY' readonly><input id='dis' value='DISABLED' disabled>"
agent-browser fill '#ro' 'HACKED'; echo "exit=$?"
agent-browser get value '#ro'
agent-browser fill '#dis' 'HACKED'; echo "exit=$?"
agent-browser get value '#dis'
```

## Actual behaviour

`fill '#ro' 'HACKED'` prints `✓ Done` and exits 0, and `#ro` afterwards holds the empty string instead of `READONLY`. The disabled input behaves identically. The field is neither left alone nor filled: it is emptied.

## Expected behaviour

`fill` refuses a readonly, disabled, or non-editable element without touching it, and exits non-zero. It should also report when the element did not accept the whole text.

## Root cause

`pub async fn fill` at `cli/src/native/interaction.rs:129` focuses the element, then clears it with a `Runtime.callFunctionOn` select-all-and-delete, then writes with `Input.insertText`, and returns `Ok(())` unconditionally. There is no check of `disabled`, `readOnly`, `isContentEditable`, or the tag before the clear, and no read-back after the write. The two mechanisms disagree about protection, so the destructive half always runs and the constructive half silently does nothing. The same missing post-check hides `maxlength` truncation and a `type=number` input rejecting non-numeric text.

## Impact

An AI agent filling a form sees `✓ Done` for every field and concludes the form is populated. In reality a server-populated readonly field such as an account number or a computed total has been blanked, and the value the agent intended to enter is nowhere. The agent then submits, and the destruction is attributed to the application rather than the tool. Because the same silence covers truncation, a password or code capped by `maxlength` is reported as fully entered.

dwin-gharibi:
## Summary

`fill` cleared the element with a JS assignment that ignores `readonly` and `disabled`, then wrote with `Input.insertText`, which honours them, so a protected field was emptied and reported as filled. `fill` now probes the element before touching it and reads the value back after writing.

## Closes

Closes #1920.

## What changed

- `fill` in `cli/src/native/interaction.rs` replaces its bare focus call with a probe that focuses and returns `{ tag, formControl, disabled, readOnly, contentEditable }` in the same round trip, before anything is cleared.
- New pure `ensure_fillable` accepts contenteditable elements and non-disabled, non-readonly `INPUT`/`TEXTAREA`, and otherwise returns an error naming the selector and the tag.
- New pure `ensure_fill_applied` compares the requested text with the value read back after `Input.insertText` and reports a mismatch, quoting both values and pointing at `maxlength`, input masks, and rejected values.
- `cli/src/mcp.rs` updates the `fill` tool description to state that it fails on readonly, disabled, non-editable, or partially accepted input.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary driving real Chrome first: a readonly input holding `READONLY` was left empty by `fill '#ro' 'HACKED'`, which printed `✓ Done` and exited 0.

- `test_fill_rejects_readonly_input` and `test_fill_rejects_disabled_input`: the error names the selector and the reason.
- `test_fill_rejects_non_editable_div`: names the selector, says "not editable", and reports the tag.
- `test_fill_rejects_missing_probe`: an element that vanished mid-fill is not fillable.
- `test_fill_accepts_text_input`, `test_fill_accepts_textarea`, `test_fill_accepts_contenteditable_element`: the supported cases still pass.
- `test_fill_applied_accepts_exact_value`, `test_fill_applied_reports_truncated_value`, `test_fill_applied_reports_rejected_value`: the read-back accepts an exact match and reports truncation and rejection.
- `e2e_fill_rejects_non_editable_elements` in `cli/src/native/e2e_tests.rs` drives a real browser: fills of `#ro`, `#dis`, and `#panel` all fail, the original values survive, and a `maxlength='5'` input fed ten characters fails too.

## Docs

`README.md` and `docs/src/app/commands/page.mdx` note that `fill` errors if the element is not editable; `cli/src/output.rs` spells out the readonly, disabled, non-editable, and partial-acceptance failures in the command help.

## Behaviour changes

`fill` now fails, without modifying the element, on readonly, disabled, and non-editable targets, and fails after the fact when the element did not take the whole text. Successful fills on ordinary inputs are unchanged apart from one extra read-back round trip.

## Risk

Medium. Callers that previously got a silent success on a protected field now see an error, which is the point, but it can surface as new failures in existing scripts. The decision logic is pure and unit-tested, and the e2e test covers the browser behaviour.


dwin-gharibi:
@ctate

PR #1923, opened 2026-09-18
fix(refs): invalidate snapshot refs on a navigation the daemon did not initiate

### LINKED ISSUE #1922: Snapshot refs survive a click-triggered navigation and the next `click @ref` is a silent no-op
## Summary

A snapshot ref stays resolvable after a click navigates the page away. Clicking such a ref on the new document does nothing at all, yet the command prints `✓ Done` and exits 0. Explicit navigation already invalidates refs; a navigation the daemon did not initiate does not.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser open file:///tmp/v1.html
agent-browser snapshot                      # button "Delete" [ref=e7]
agent-browser find text "Go Second" click   # navigates to v2.html
agent-browser click @e7; echo "exit=$?"
agent-browser get text '#log'               # empty
agent-browser click '#danger2'              # this one does register
agent-browser get text '#log'
```

## Actual behaviour

`click @e7` prints `✓ Done` and exits 0, but the page log stays empty: nothing was clicked. A CSS-selector click such as `click '#danger2'` on the same page does register, which shows the page itself is live and only the ref is stale.

## Expected behaviour

A navigation caused by a click invalidates the refs issued for the previous document, exactly as an explicit `open` or `goto` already does, so `click @e7` fails with a stale-ref error instead of reporting success.

## Root cause

`cli/src/native/actions.rs:1775`, the `Page.frameNavigated` arm of the CDP event loop. When the main frame navigates (`frame.get("parentId").is_none()`) the handler clears WebMCP page scope and updates frame origins, but it never touches `self.ref_map`. Ref invalidation lived only on the explicit navigation commands, which call into `RefMap` in `cli/src/native/element.rs`. A click that causes a navigation reaches the daemon only as this event, so the old document's refs are left in the map. The ref then resolves to a backend node id belonging to a document that no longer exists, the CDP call targeting it is a no-op, and the command reports success because nothing raised an error.

## Impact

An AI agent that snapshots once and then acts on refs will, after any click that navigates, issue clicks that do nothing while every one of them reports success. The agent believes it deleted the record, accepted the dialog or submitted the form, and keeps building on that false state. Worse, the refs may point at elements whose labels described the previous page, so the agent's own reasoning about what it clicked is wrong too.

dwin-gharibi:
## Summary

Refs issued by a snapshot survived a click-triggered navigation, so the next `click @ref` silently did nothing and still reported success. Main-frame `Page.frameNavigated` now invalidates that page's refs when the document was actually replaced.

## Closes

Closes #1922.

## What changed

- `cli/src/native/element.rs`: new `RefMap::invalidate_page_on_document_change(page_session, session, loader)`. It compares the incoming loader id against the document recorded for the page, returns false when the session and loader are unchanged (a same-document navigation keeps its refs), returns false when the page has no recorded document at all (a page that never issued refs must not clear another page's refs), and otherwise calls the existing `invalidate_page` and returns true.
- `cli/src/native/actions.rs`: the main-frame branch of the `Page.frameNavigated` handler in `DaemonState` now calls `invalidate_page_on_document_change` with the frame's `loaderId`, alongside the existing WebMCP page-scope clearing.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary in real Chrome first: snapshot issued `button "Delete" [ref=e7]`, a `find text "Go Second" click` navigated to the second page, and `click @e7` then printed `✓ Done` at exit 0 with an empty page log while `click '#danger2'` on the same page registered.

- `click_triggered_navigation_invalidates_issued_refs`: after a document change with a new loader id, `get("e7")` and the durable ref are both gone while `next_ref_num()` stays at 8, so the monotonic counter is preserved.
- `document_change_on_one_page_preserves_other_page_refs`: invalidating page-b leaves page-a's durable ref intact.
- `same_document_navigation_keeps_refs`: the same session and loader returns false and the ref still resolves.
- `document_change_on_untracked_page_keeps_refs`: a change reported for a page with no recorded document returns false and clears nothing.
- `missing_loader_invalidates_tracked_page`: an absent loader id invalidates a tracked page rather than assuming the document survived.

## Docs

`README.md`, `cli/src/output.rs` (the `snapshot` help text) and `skill-data/core/SKILL.md` all said refs survive across snapshots and to re-snapshot after navigation; each now states that navigation invalidates refs, including a navigation caused by a click.

## Behaviour changes

A ref used after a click-triggered navigation now fails with a stale-ref error instead of silently doing nothing. Same-document navigations and refs on other pages are unaffected.

## Risk

Low to moderate. The invalidation is gated on an actual loader-id change and on the page already having a recorded document, so unrelated pages and same-document updates keep their refs. The failure mode of a false positive is one extra required snapshot, not a wrong action.


dwin-gharibi:
@ctate

PR #1925, opened 2026-09-18
fix(upload): resolve and validate file paths in the CLI before sending them to the browser

### LINKED ISSUE #1924: `upload` attaches a 0-byte file for a relative or nonexistent path and reports success
## Summary

`upload` forwards file paths to the browser verbatim. There is no existence check, and a relative path is resolved by the daemon process rather than by the CLI, so it resolves against the daemon's working directory (`/`). Both cases attach an empty file, and both print `✓ Done` and exit 0.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
mkdir -p /root/vup && printf 'hello world\n' > /root/vup/real.txt   # 12 bytes
agent-browser open file:///tmp/upload.html

agent-browser upload '#file1' /root/vup/real.txt   # attaches real.txt:12

cd /root/vup
agent-browser upload '#file1' ./real.txt           # attaches real.txt:0
agent-browser upload '#file1' ./ghost.txt          # attaches ghost.txt:0
```

## Actual behaviour

The absolute path attaches `real.txt:12` as expected. From inside `/root/vup`, `./real.txt` attaches `real.txt:0` and a path that does not exist at all, `./ghost.txt`, attaches `ghost.txt:0`. All three print `✓ Done` and exit 0.

## Expected behaviour

A path that does not exist, or that is not a regular file, is a parse error with a non-zero exit. A relative path is resolved against the working directory of the process the user typed the command in, so `./real.txt` uploads the same 12 bytes as the absolute form.

## Root cause

`cli/src/commands.rs:572`, the `"upload"` arm of `parse_command_inner`: `Ok(json!({ "id": id, "action": "upload", "selector": sel, "files": &rest[1..] }))`. The path strings go straight into the command payload with no metadata lookup and no canonicalisation. The payload is then executed by the long-lived daemon, which calls `DOM.setFileInputFiles`. The daemon's working directory is not the user's, and in practice is `/`, so a relative path is resolved somewhere the user never meant. Chrome does not reject a file input path it cannot read; it attaches an entry with the basename and a zero length, which looks like a real attachment to the page and to the CLI.

## Impact

An AI agent uploading a document, an avatar or a data file gets `✓ Done` and moves on, having in fact submitted an empty file under a plausible filename. Because the basename in the form matches the one the agent intended, a screenshot or a snapshot taken afterwards confirms the upload. The error surfaces much later, in whatever consumes the file, with no trace back to the command that reported success.

dwin-gharibi:
## Summary

`upload` passed paths through verbatim, so a relative path was resolved by the daemon (working directory `/`) and a nonexistent path was accepted outright. Both attached a 0-byte file and exited 0. Paths are now resolved against the CLI's working directory and checked before the command is sent.

## Closes

Closes #1924.

## What changed

- `cli/src/commands.rs`: new `resolve_upload_path(path, base)`. It joins a relative path onto `base`, drops `Component::CurDir` segments, calls `std::fs::metadata` and returns `ParseError::InvalidValue` when the path cannot be read, and a second `InvalidValue` naming "not a file" when it exists but is not a regular file. It returns the resolved absolute path as a string.
- The `"upload"` arm of `parse_command_inner` now takes `std::env::current_dir()` as the base and maps every path in `rest[1..]` through `resolve_upload_path`, so the payload carries resolved, verified paths instead of the raw arguments. The literal usage string is factored into a shared `UPLOAD_USAGE` constant.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary in real Chrome first: with a 12 byte `/root/vup/real.txt`, the absolute path attached `real.txt:12`, while from `/root/vup` both `./real.txt` and `./ghost.txt` attached a 0-byte file at exit 0.

- `test_resolve_upload_path_keeps_existing_absolute_path`: an existing absolute path is returned unchanged even with base `/`.
- `test_resolve_upload_path_resolves_relative_against_base`: `./real.txt` resolves to the file inside the base directory, not the daemon's.
- `test_resolve_upload_path_rejects_missing_absolute_path` and `test_resolve_upload_path_rejects_missing_relative_path`: both return `ParseError::InvalidValue` whose formatted message names `ghost.txt`.
- `test_resolve_upload_path_rejects_directory`: a directory is rejected with a message containing "not a file".
- `test_resolve_upload_path_validates_every_file`: two existing names both resolve, and a third missing name in the same directory still errors, so validation is per file rather than only on the first.

## Docs

`cli/src/output.rs` (the `upload` help text) now states that each file must exist and that relative paths resolve against the current working directory. `cli/src/mcp.rs` updates the `files` parameter description of the MCP `upload` tool to say the same and to note that relative paths resolve against the MCP server's working directory.

## Behaviour changes

`upload` with a missing path, or a path naming a directory, now fails with a non-zero exit instead of attaching an empty file. Relative paths now resolve where the user typed them, so a previously empty attachment becomes the intended file.

## Risk

Low. Validation happens in the parser, before any browser call, and absolute paths to existing files behave exactly as before. A caller that deliberately depended on the daemon's `/`-relative resolution would break, but that resolution was never documented.


dwin-gharibi:
@ctate

PR #1927, opened 2026-09-18
fix(check): reject uncheckable elements and verify the resulting checked state

### LINKED ISSUE #1926: `check` and `uncheck` report success on elements they cannot toggle
## Summary

`check` and `uncheck` never assert the resulting state. `uncheck` on a checked radio leaves it checked, and `check` on a plain div, a disabled button or a text input clicks something and reports done. Every one of these prints `✓ Done` and exits 0.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser open file:///tmp/check.html
agent-browser uncheck '#r1'; echo "exit=$?"     # #r1 is a checked radio
agent-browser ischecked '#r1'                   # still true
agent-browser check '#notbox'; echo "exit=$?"   # a plain <div>
agent-browser check '#b4'; echo "exit=$?"       # a disabled <button>
agent-browser check '#t1'; echo "exit=$?"       # an <input type="text">
```

## Actual behaviour

`uncheck '#r1'` prints `✓ Done` and exits 0, and `ischecked '#r1'` still reports true. `check` on the div, the disabled button and the text input each print `✓ Done` and exit 0 as well, having changed no checked state at all.

## Expected behaviour

An element that is not a checkbox, a radio, or an element with a checkbox-like ARIA role is an error. A disabled element is an error. A checked radio cannot be unchecked by clicking, so `uncheck` on one is an error explaining that a radio is cleared only by selecting another radio in its group. Any toggle that does not end in the requested state is an error. Playwright's `setChecked` throws in all of these cases.

## Root cause

`cli/src/native/interaction.rs:526` (`check`) and `cli/src/native/interaction.rs:580` (`uncheck`). Both read the current state through `is_element_checked`, click when the state differs, and then re-read. The re-read is used only to decide whether to try a JavaScript click as a fallback: after that fallback the result is discarded and the function returns `Ok`. There is no check of what the element is before clicking either, so a div or a text input is simply clicked. The comment at `cli/src/native/interaction.rs:555` already reads "Verify the click changed the state (Playwright parity: _setChecked re-checks)", but the verification it names never produces a failure, so the claimed parity was not implemented.

## Impact

An AI agent using `check`/`uncheck` to set consent boxes, filters, or opt-outs gets an unconditional success signal. It moves on believing a radio was cleared or a box was ticked, then submits a form whose real state is different from the state it reasoned about. Because the wrong selector case also succeeds, a selector typo that lands on a nearby div is indistinguishable from a working command, and the agent has no way to notice until the submitted result is wrong.

dwin-gharibi:
## Summary

`check` and `uncheck` clicked whatever the selector matched and always returned `Ok`, so a checked radio, a div, a disabled button and a text input all reported success. They now probe the element first and assert the requested state afterwards, matching what the existing "Playwright parity" comment already claimed.

## Closes

Closes #1926.

## What changed

- `cli/src/native/interaction.rs`: new `CheckableProbe` struct holding the resolved control's tag, input type, ARIA role, disabled flag and checked flag, with helpers `is_native_input`, `has_aria_checked_role`, `is_radio` and `describe`. A new `ARIA_CHECKED_ROLES` constant lists the seven roles `is_element_checked` already understands.
- New `probe_checkable` performs the same follow-label and nested-input retargeting as `is_element_checked` and returns that snapshot, so both decisions act on the control that is really clicked.
- New `ensure_checkable(probe, selector, want_checked)` runs before any click: it rejects elements that are neither a native checkbox/radio nor an ARIA checked role, rejects disabled elements, and rejects `uncheck` on a checked radio with a message explaining that a radio is cleared by selecting another radio in its group.
- New `verify_checked_state(probe, selector, want_checked)` runs after the click and the existing `js_click_checkbox` fallback, and errors when the state was not reached.
- `check` and `uncheck` now call `probe_checkable`, `ensure_checkable`, and `verify_checked_state` in place of their bare `is_element_checked` reads. The decision logic is plain data in and `Result` out, so it is unit-testable without a browser.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary in real Chrome first: `uncheck '#r1'` on a checked radio printed `✓ Done` at exit 0 with the radio still checked, and `check` on a div, a disabled button and a text input each printed `✓ Done` at exit 0.

Unit tests in `cli/src/native/interaction.rs`:

- `test_checkbox_reaching_requested_state_is_ok`: a checkbox that reaches the requested state passes both `ensure_checkable` and `verify_checked_state`, in each direction.
- `test_checkbox_state_unchanged_is_error`: an unchanged checkbox produces an error naming the selector and saying "did not become checked" or "did not become unchecked".
- `test_checked_radio_cannot_be_unchecked`: `uncheck` on a checked radio errors and the message explains how radios are cleared, while checking a radio and unchecking an already-unchecked one stay allowed.
- `test_non_checkable_elements_are_rejected`: a div, a button and `input type="text"` are each rejected with "is not a checkbox or radio", and the message describes the element.
- `test_disabled_checkbox_is_rejected`: a disabled checkbox errors with a message mentioning "disabled".
- `test_aria_checkbox_roles_are_checkable`: elements carrying a checkbox-like ARIA role are still accepted.

End to end in `cli/src/native/e2e_tests.rs`:

- `e2e_check_uncheck_rejects_uncheckable_elements`: in a real headless browser, `uncheck` on a checked radio returns `success: false` with an error mentioning "radio" and `ischecked` confirms the radio is still checked; `check` on a div, a disabled button and a text input each return `success: false` with "is not a checkbox or radio"; `check` on a disabled checkbox errors with "disabled"; and checking a radio still works.

## Docs

`cli/src/output.rs` (the `check` and `uncheck` help text) now lists the failure conditions, and `README.md` updates the one-line descriptions of both commands.

## Behaviour changes

`check` and `uncheck` now exit non-zero on a non-checkable element, a disabled element, a checked radio passed to `uncheck`, or a toggle that does not take effect. Successful toggles are unchanged, including the ARIA-role and follow-label paths and the JavaScript click fallback.

## Risk

Moderate, because commands that previously always succeeded can now fail. That is the intent, and the probe reuses the resolution `is_element_checked` already performed, so the set of elements treated as checkable is unchanged. Pages that relied on a `check` no-op passing will now surface a real error.


dwin-gharibi:
@ctate

PR #1929, opened 2026-09-18
fix(select): reject ambiguous option matches instead of selecting the last one

dwin-gharibi:
## Summary

`select` merged every option matching a requested value into one set and marked them all selected, so a single-select kept whichever the browser saw last. Matching now resolves in Rust and an ambiguous request is an error naming the candidates.

## Closes

Closes #<issue-number-tbd> (see issues/f19-select-label-priority.md).

## What changed

- `cli/src/native/interaction.rs`: `select_option` now runs two page-side calls. The first reads the options into a new `SelectOptionInfo` struct (`value`, `label`, `text`); the second applies resolved indices and guards against the option list changing between the calls.
- New `resolve_select_matches` decides the selection in Rust: exact option value, exact label, or exact text first, and only when nothing matches exactly are labels compared again through `normalize_option_text`, which collapses whitespace and drops zero-width characters. A value matching several distinct options returns an error.
- New helpers `describe_options` and `no_match_error` render candidates as `value ("label")` in both the ambiguity and the no-match errors.
- Several requests still select several options, which `<select multiple>` needs.
- `cli/src/mcp.rs`: the `select` tool description and its `values` schema state the matching order and that an ambiguous value is an error.
- `cli/src/output.rs`: `select` help text rewritten to describe exact matching, the normalization fallback, and the ambiguity error.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary and real Chrome before any code changed.

Unit tests in `cli/src/native/interaction.rs`: `test_select_value_label_collision_is_ambiguous` asserts the reported case errors and names both `a ("Alpha")` and `Alpha ("TrickyValueIsAlpha")`; `test_select_duplicate_label_is_ambiguous` covers two options sharing one label; `test_select_unambiguous_value_match` and `test_select_unambiguous_label_match` assert the ordinary paths still resolve; `test_select_normalizes_whitespace_in_labels` covers non-breaking and zero-width characters; `test_select_exact_label_beats_normalized_twin` asserts exact matching runs ahead of normalization; `test_select_no_match_lists_available_options`, `test_select_multiple_requests_select_several_options`, and `test_select_multiple_requests_fail_on_any_miss` cover the remaining paths.

E2E test `e2e_select_option_ambiguous_matches` in `cli/src/native/e2e_tests.rs` drives real Chrome: the ambiguous request fails with "Multiple options matched" and leaves the element on its previous value, the exact value still selects, and a multi-select still takes two values.

## Docs

`README.md`, `skill-data/core/SKILL.md`, and `skill-data/core/references/commands.md` describe the matching order and the ambiguity error.

## Behaviour changes

A value matching several options now fails with a non-zero exit instead of selecting one of them. Unambiguous selections are unchanged.

## Risk

Low. Callers relying on the old silent behaviour were getting an arbitrary option; they now get an error that names the exact value to pass.


dwin-gharibi:
@ctate

PR #1931, opened 2026-09-18
fix(tab): reject labels that collide with the `t<N>` tab-id namespace

### LINKED ISSUE #1930: A tab label shaped like `t1` is accepted and then permanently unreachable
## Summary

`tab new --label` accepts any identifier-shaped label, including ones that collide with the `t<N>` tab-id namespace. Tab refs resolve as ids first, so a tab labelled `t1` can never be reached by its label, and `tab t1` silently acts on whichever tab owns the id `t1`.

## Severity

Medium.

## Affected version

## Steps to reproduce

```bash
agent-browser open https://example.com/v1.html
agent-browser tab new --label t1 https://example.com/v2.html
echo "exit=$?"
agent-browser tab list
agent-browser tab t1
agent-browser get url
```

## Actual behaviour

`open` creates `[t1]`. `tab new --label t1` is accepted, exits 0, and the tab is listed as `[t2] t1`. `tab t1` switches to the first tab, so the labelled tab is unreachable by the label its creator just assigned.

## Expected behaviour

A label that takes the reserved `t<N>` id form is rejected at assignment time with a non-zero exit and an error explaining the id namespace, so no tab can be created with a handle that will never resolve to it.

## Root cause

`cli/src/native/browser.rs:1577`, the label check in the `tab new` path.

Label validation calls `is_valid_label`, which only enforces the identifier shape: start with a letter, then letters, digits, `-`, and `_`. `t1` satisfies that. Everywhere a tab ref is resolved, the `t<N>` id form is matched first, so the label is shadowed the moment it is stored. There is no check anywhere that a label does not take the id form.

## Impact

The README presents ids and labels as interchangeable handles, so an agent that labels a tab and then addresses it by that label operates on the wrong tab with no error to notice. `tab close t1` is the destructive case: it closes the id-matched tab, not the labelled one, discarding page state the agent believed it was cleaning up. Nothing in the output distinguishes this from the intended operation, and labels derived from page data or from a loop counter can easily take the `t<N>` shape.

dwin-gharibi:
## Summary

`tab new --label t1` was accepted and then shadowed by the tab id `t1`, leaving the labelled tab unreachable by its label. Labels in the reserved id form are now refused when they are assigned.

## Closes

Closes #1930.

## What changed

- `cli/src/native/browser.rs`: new `looks_like_tab_id` returns true for strings shaped like a stable tab id (`t1`, `T42`, `t0`), and new `validate_label` combines the existing `is_valid_label` identifier rule with that check, returning an error that explains the reserved namespace.
- The `tab new` label path now calls `validate_label` instead of `is_valid_label` directly, so the uniqueness check that follows is unchanged.
- `cli/src/mcp.rs`: the `tab new` tool schema gives `label` a description stating the identifier rule and the reserved `t<N>` form.
- `cli/src/output.rs`: `tab` help text states the label rule and that `t1` is rejected as a label.

## Tests

The bug was reproduced end to end against the shipped 0.37.1 binary and real Chrome first.

Unit tests in `cli/src/native/browser.rs`: `test_looks_like_tab_id` asserts `t1`, `T42`, and `t0` are id-shaped while `t`, `tab`, `t1a`, and `docs` are not; `test_validate_label_rejects_tab_id_shape` asserts `t1` is rejected with a message mentioning the reserved id namespace and quoting the label, and that `T42` and `t0` are rejected too; `test_validate_label_accepts_ordinary_labels` asserts `docs`, `tab`, `t1a`, and `app-2` still pass; `test_validate_label_rejects_invalid_characters` asserts `2docs` still fails on the identifier rule with the existing "must start with a letter" message.

## Docs

`README.md`, `docs/src/app/commands/page.mdx`, and `skill-data/core/references/commands.md` all state that a label must start with a letter, may contain only letters, digits, `-`, and `_`, and cannot take the `t<N>` id form.

## Behaviour changes

`tab new --label t1` (and any `t<N>` or `T<N>` label) now fails with a non-zero exit instead of creating a tab with an unreachable label. Every other label is unaffected, including `t1a` and `tab`.

## Risk

Low. The rejected labels could never be used to address the tab they named, so no working workflow depended on them.


dwin-gharibi:
@ctate

PR #1933, opened 2026-09-18
fix(read): decode response bodies with the declared charset

### LINKED ISSUE #1932: `read` mangles non-UTF-8 pages and reports success
## Summary

`read <url>` decodes every response body as UTF-8 and ignores the charset the server declared. A page served as `Content-Type: text/html; charset=iso-8859-1` comes back with every non-ASCII byte replaced by U+FFFD, exit 0 and no warning. The browser path in the same binary reads the same URL correctly, so the two disagree about one page.

## Severity

High.

## Affected version

## Steps to reproduce

Serve real latin-1 bytes for `Café ÜBER Straße` inside an `<h1>` with `Content-Type: text/html; charset=iso-8859-1` on port 8940, then:

```bash
agent-browser open http://127.0.0.1:8940/latin1
agent-browser get text h1
agent-browser read http://127.0.0.1:8940/latin1
echo "exit=$?"
agent-browser read http://127.0.0.1:8940/latin1 --raw
```

## Actual behaviour

`get text h1` prints `Café ÜBER Straße`. `read` prints the same text with every non-ASCII byte replaced by the replacement character `\u{FFFD}`, exits 0, and prints no warning. `--raw` is affected the same way, so there is no flag that recovers the original bytes.

## Expected behaviour

`read` should decode the body with the charset declared in the `Content-Type` header, fall back to a `<meta charset>` declaration in the document, and only then assume UTF-8, which is what a browser does and what the browser path in this same binary already produces.

## Root cause

`cli/src/read.rs:1010`, the final line of `read_limited_text` (`cli/src/read.rs:992`).

`read_limited_text` streams the response into a `Vec<u8>` and then returns `String::from_utf8_lossy(&bytes).to_string()`. The function never receives the `Content-Type` value, which the caller `fetch_read_url` already has in scope, so the declared charset is not merely mis-handled, it is not available at the decoding point at all. `from_utf8_lossy` is lossy by definition: any byte sequence that is not valid UTF-8 becomes U+FFFD silently, and there is no error path, so nothing downstream can tell a legacy-encoded page from one that really contains replacement characters. Every consumer of the body, including `--raw`, the markdown extraction, `--outline` and `--filter`, works on the already-corrupted string.

## Impact

An AI agent reading a page for content receives text that looks like it was fetched successfully, with exit 0 and no diagnostic, while names, addresses and prices have lost their accented characters. Because the corrupted text is still mostly readable English, the agent has no signal to retry, and any value it extracts and acts on, such as a name it types into a form or a term it quotes back to the user, carries the corruption forward. Worse, an agent that cross-checks with the browser path sees two different answers for the same URL from one tool.

dwin-gharibi:
## Summary

`read` decoded every body with `String::from_utf8_lossy`, so a page served as ISO-8859-1 lost each non-ASCII byte to a replacement character while the command still exited 0. Bodies are now decoded with the charset from the `Content-Type` header, then from a `<meta>` declaration, and UTF-8 otherwise.

## Closes

Closes #1932.

## What changed

- `cli/src/read.rs`: `read_limited_text` now takes the `content_type` string from `fetch_read_url` and hands the collected bytes to a new `decode_body` instead of `String::from_utf8_lossy`.
- New `decode_body` resolves an encoding from the header charset, then from the document, then falls back to UTF-8, and decodes with `encoding_rs` so undecodable bytes become replacement characters rather than an error, matching how a browser renders the same response.
- New `charset_from_content_type` parses the `charset` parameter out of the header, tolerating quoting and surrounding whitespace.
- New `charset_from_meta` scans the first `META_CHARSET_SNIFF_LIMIT` (4096) bytes for a `<meta charset>` or `<meta http-equiv="Content-Type" content="...; charset=...">` declaration, using a new `tag_attribute_value` helper that handles quoted and unquoted attribute values and requires an attribute-name boundary.
- `cli/Cargo.toml`: adds the `encoding_rs` dependency.

## Tests

The bug was reproduced end to end first: the shipped 0.37.1 binary against a local server returning real latin-1 bytes with `charset=iso-8859-1`, where the browser path returned the accented text and `read` returned replacement characters.

Unit tests in `cli/src/read.rs`: `decode_body_uses_header_charset` asserts the latin-1 bytes decode to `Café ÜBER Straße`; `decode_body_uses_meta_charset_when_header_omits_it` and `decode_body_uses_meta_http_equiv_content_type` cover both document declarations when the header carries no charset; `decode_body_keeps_utf8_with_explicit_charset` and `decode_body_keeps_utf8_without_charset` assert UTF-8 bodies are unchanged with and without a declaration; `decode_body_ignores_unknown_charset_and_falls_back_to_utf8` asserts an unrecognized label falls back rather than failing; `decode_body_replaces_invalid_bytes_instead_of_failing` asserts genuinely invalid bytes still yield U+FFFD with the surrounding text intact.

`run_read_decodes_declared_charset` is an end-to-end test over a local TCP server that serves latin-1 bytes with `Content-Type: text/html; charset=iso-8859-1` and asserts the extracted markdown contains `# Café ÜBER Straße`.

## Docs

`README.md` and `docs/src/app/commands/page.mdx` state the decoding order for `read` response bodies.

## Behaviour changes

Bodies declaring a legacy charset now decode to the characters the server meant. UTF-8 responses, the overwhelming majority, are byte-for-byte unchanged. No exit code or output format changes.

## Risk

Low. The fallback when no charset is declared or the label is unrecognized is still UTF-8, so the previous behaviour is preserved everywhere a charset is absent.


dwin-gharibi:
@ctate

PR #1935, opened 2026-09-18
fix(screenshot): make the output path extension select the encoder

### LINKED ISSUE #1934: `screenshot` writes PNG bytes into a `.jpg` file and JPEG bytes into a `.png` file
## Summary

The output path extension is used only to decide whether an argument is a path or a selector; the encoder comes from `--screenshot-format` alone. A `.jpg` path with no flag therefore receives PNG bytes, and a `.png` path with `--screenshot-format jpeg` receives JPEG bytes. Both exit 0 with no warning.

## Severity

High.

## Affected version

## Steps to reproduce

```bash
agent-browser screenshot /root/v.jpg
echo "exit=$?"
file -b /root/v.jpg

agent-browser screenshot /root/v.png --screenshot-format jpeg
echo "exit=$?"
file -b /root/v.png
```

## Actual behaviour

Both commands exit 0 and print no warning. `file -b /root/v.jpg` reports `PNG image data, 1280 x 577`. `file -b /root/v.png` reports `JPEG image data`.

## Expected behaviour

The bytes should match the extension. A `.jpg` or `.jpeg` path should select the JPEG encoder, a `.png` path should select PNG, and a `--screenshot-format` that contradicts the extension should either be overridden with a warning or be rejected, rather than silently producing a mislabelled file.

## Root cause

`cli/src/commands.rs:882`, in the `screenshot` arm of `parse_command_inner`.

A few lines earlier, at `cli/src/commands.rs:864`, `has_path_extension` checks for `.png`, `.jpg`, `.jpeg` and `.webp`, but that value feeds only the `is_path` decision that separates a path argument from a CSS selector. The extension is then discarded. The command payload gets `format` only from `if let Some(ref fmt) = flags.screenshot_format`, so with no flag the daemon falls back to its PNG default regardless of the file name, and with a conflicting flag the flag wins. The `--screenshot-quality` warning at `cli/src/commands.rs:887` compares against the flag for the same reason, so a `.jpg` path plus `--screenshot-quality` warns that quality is ignored while also writing PNG. Nothing anywhere compares the chosen encoder with the extension.

## Impact

An AI agent that picks `.jpg` to keep an image small, or to satisfy a downstream tool that only accepts JPEG, gets exit 0 and a file that lies about its contents. The agent has a false success signal: the capture "worked", so it moves on, and the breakage surfaces later in whatever consumes the file, usually as a decode error far from the command that caused it. Agents deliberately choosing JPEG to reduce token cost instead pay for a full-size PNG without ever being told.

dwin-gharibi:
## Summary

The screenshot encoder came from `--screenshot-format` alone while the path extension was only used to tell a path from a selector, so `.jpg` files held PNG bytes and `.png` files could hold JPEG bytes. The extension is now authoritative and a conflicting flag is reported instead of applied.

## Closes

Closes #1934.

## What changed

- `cli/src/commands.rs`: new `screenshot_format_from_extension` maps a case-insensitive `.png`, `.jpg`, `.jpeg` or `.webp` suffix to the encoder it implies.
- New `resolve_screenshot_format` returns the encoder for a given path and flag: the extension wins when present, the flag applies when the path carries no image extension or there is no path, and PNG remains the default. It rejects a `.webp` path, which the encoder cannot produce, and rejects a `--screenshot-format` value other than `png`, `jpeg` or `jpg`.
- The `screenshot` arm of `parse_command_inner` now always sets `cmd["format"]` from `resolve_screenshot_format`, returns `ParseError::InvalidValue` for the rejected cases, and prints a warning when the passed `--screenshot-format` would have produced a different encoder than the extension.
- The `--screenshot-quality` warning now compares against the resolved format rather than the flag, so `.jpg` plus `--screenshot-quality` no longer warns incorrectly.
- `cli/src/mcp.rs`: the screenshot tool's `path` and `format` descriptions state that the extension selects the encoder and that `.webp` is rejected.
- `cli/src/output.rs`: screenshot help text and examples updated.

## Tests

The bug was reproduced end to end first against the shipped 0.37.1 binary and real Chrome, confirming with `file -b` that `screenshot /root/v.jpg` produced PNG bytes.

Unit tests in `cli/src/commands.rs`: `resolve_screenshot_format_reads_jpg_extension` (including uppercase `.JPEG`) and `resolve_screenshot_format_reads_png_extension` assert extension mapping; `resolve_screenshot_format_defaults_to_png` covers no path and an extensionless path; `resolve_screenshot_format_uses_flag_without_extension` asserts the flag still applies when there is nothing to read from the path; `resolve_screenshot_format_accepts_matching_flag_and_extension` and `resolve_screenshot_format_extension_beats_conflicting_flag` cover agreement and conflict in both directions; `resolve_screenshot_format_rejects_webp_extension` and `resolve_screenshot_format_rejects_unsupported_flag` assert the error messages.

End-to-end through the parser: `test_screenshot_jpg_path_selects_jpeg_encoder`, `test_screenshot_png_path_overrides_conflicting_format_flag`, `test_screenshot_without_path_keeps_format_flag`, `test_screenshot_selector_args_are_not_treated_as_paths` (asserting `#header`, `.my-button` and `@e1` still parse as selectors with the flag honoured), and `test_screenshot_rejects_webp_path`.

## Docs

`README.md`, `docs/src/app/commands/page.mdx`, `docs/src/app/configuration/page.mdx`, `docs/src/app/files/page.mdx`, `skill-data/core/SKILL.md` and `skill-data/core/references/commands.md` all state that the path extension picks the encoder and that `--screenshot-format` applies only without one.

## Behaviour changes

A `.jpg` or `.jpeg` path now writes JPEG and a `.png` path now writes PNG. A `--screenshot-format` that contradicts the extension is ignored with a warning on stderr rather than applied. A `.webp` output path is now a parse error instead of silently writing PNG.

## Risk

Low for the format change, since the previous output was mislabelled in exactly these cases. The one deliberate breakage is `.webp`, which previously produced a PNG under a WebP name and now fails with a message naming the accepted extensions.


dwin-gharibi:
@ctate

PR #1937, opened 2026-09-18
fix(eval): report unserializable results instead of collapsing them to null

### LINKED ISSUE #1936: `eval` returns `null` for NaN, Infinity and BigInt, indistinguishable from a real null
## Summary

Every JavaScript value that JSON cannot represent comes back from `eval` as `null` with exit 0. `NaN`, `Infinity`, `-Infinity`, `-0`, BigInt literals and `undefined` are all flattened to the same output as an expression that genuinely evaluates to `null`.

## Severity

Medium.

## Affected version

## Steps to reproduce

```bash
agent-browser eval NaN;        echo "exit=$?"
agent-browser eval Infinity;   echo "exit=$?"
agent-browser eval 1n;         echo "exit=$?"
agent-browser eval 0/0;        echo "exit=$?"
agent-browser eval "({a:1})"
agent-browser eval null
```

## Actual behaviour

The first four commands each print `null` and exit 0. `eval "({a:1})"` returns the object correctly, so the evaluation pipeline itself works. `eval null` prints the same `null`, so there is no way to tell the cases apart. With `--json` the result field is likewise JSON `null`.

## Expected behaviour

A value JSON cannot spell should be reported as itself, not as `null`. Chrome already sends these as `unserializableValue` carrying the JavaScript token (`NaN`, `Infinity`, `-0`, `1n`), so the information is available and should reach the caller in a form distinct from a real `null`.

## Root cause

`cli/src/native/browser.rs:1250`, the last line of `BrowserManager::evaluate`: `Ok(result.result.value.unwrap_or(Value::Null))`.

CDP's `Runtime.RemoteObject` omits `value` for any result JSON cannot carry and sends the JavaScript spelling in `unserializableValue` instead. The struct in `cli/src/native/cdp/types.rs:262` already deserializes that field as `unserializable_value`, but nothing in the crate ever reads it: it is the only occurrence of the name in the Rust sources. `unwrap_or(Value::Null)` therefore turns the missing `value` into a real `null`, and `undefined` (which also arrives with no `value`, distinguished only by `object_type`) collapses the same way. That the sibling cases of Symbols and circular objects do produce an error with exit 1 shows this was an oversight rather than a deliberate choice to normalize.

## Impact

An AI agent reading a numeric value out of a page gets `null` from a computation that actually produced `NaN` or `Infinity`, with a success exit code. Both readings are wrong in opposite directions: `null` suggests a missing field, so the agent retries or falls back, while the real answer is that the arithmetic is broken. A BigInt id read from a page comes back as `null` rather than a value the agent can carry forward, again with no error to signal that anything was lost.

dwin-gharibi:
## Summary

`eval` returned JSON `null` for every result CDP cannot serialize, so `NaN`, `Infinity`, `-Infinity`, `-0`, BigInt literals and `undefined` were indistinguishable from a real `null`. These now carry the JavaScript token through to the caller.

## Closes

Closes #1936.

## What changed

- `cli/src/native/browser.rs`: new `remote_object_to_value` converts a `Runtime.RemoteObject` into the value `eval` returns. It passes a present `value` through unchanged, tags a present `unserializableValue` as `{"unserializable": "<token>"}`, maps an `object_type` of `undefined` to `{"unserializable": "undefined"}`, and only then falls back to `Value::Null`.
- `BrowserManager::evaluate` now ends in `Ok(remote_object_to_value(result.result))` instead of `Ok(result.result.value.unwrap_or(Value::Null))`. The `unserializable_value` field was already deserialized on the struct and simply never read.
- `cli/src/output.rs`: new `unserializable_token` recognizes the tag (a single-key object whose `unserializable` value is a string) and `print_primary_response` prints the bare JavaScript token for humans, so plain-text `eval NaN` prints `NaN`. `--json` still shows the tagged object.
- `cli/src/output.rs`: `eval` help text documents both forms.

## Tests

The bug was reproduced end to end first against the shipped 0.37.1 binary and real Chrome, where `eval NaN`, `eval Infinity`, `eval 1n` and `eval 0/0` each printed `null` with exit 0.

Unit tests in `cli/src/native/browser.rs`: `test_remote_object_value_passes_through` asserts an ordinary object result is unchanged; `test_remote_object_null_stays_null` asserts a genuine `null` is still `null`; `test_remote_object_unserializable_numbers` loops over `NaN`, `Infinity`, `-Infinity` and `-0` and asserts each is tagged rather than nulled; `test_remote_object_unserializable_bigint` covers `1n`; `test_remote_object_undefined_is_distinct_from_null` asserts `undefined` is tagged and explicitly not equal to `Value::Null`.

Unit tests in `cli/src/output.rs`: `test_unserializable_token_extracts_js_spelling` asserts the token is recovered for `NaN` and `undefined`; `test_unserializable_token_ignores_ordinary_results` asserts a plain `null`, the string `"NaN"`, an unrelated object, and an object with an extra key are all left alone, so a page value shaped like the tag is not misread.

E2E test `e2e_evaluate_unserializable_values` in `cli/src/native/e2e_tests.rs` drives real Chrome and asserts each of `NaN`, `Infinity`, `-Infinity`, `-0`, `1n` and `undefined` returns its tagged form while `null` still returns `Value::Null`.

## Docs

`README.md` documents the plain-text and `--json` shapes for unserializable `eval` results.

## Behaviour changes

`eval` expressions producing these values now return `{"unserializable": "<token>"}` in JSON and print the bare token in plain text, where they previously returned `null`. Exit codes are unchanged, and Symbols and circular objects still error with exit 1 as before.

## Risk

Low. Only results that previously returned a misleading `null` change shape. A caller that pattern-matched on `null` for these values was already acting on wrong data; the tag is a single-key object, and the output-side recognizer requires exactly that shape so genuine page data cannot be mistaken for it.


dwin-gharibi:
@ctate

PR #1940, opened 2026-09-18
fix(get): report a missing attribute explicitly instead of printing the generic success line

### LINKED ISSUE #1938: `get attr` on a missing attribute prints the generic success line instead of saying it is absent
## Summary

`get attr <selector> <name>` prints the attribute value when it exists, but when the element does not carry that attribute it prints `✓ Done`, the same line every mutating command prints on success. There is no way to tell the read produced nothing, and `--json` reports `"value": null`.

## Severity

Medium.

## Affected version

## Steps to reproduce

```bash
agent-browser open 'data:text/html,<button id="btn" data-x="1">Go</button>'
agent-browser get attr '#btn' data-x
echo "exit=$?"
agent-browser get attr '#btn' data-absent
echo "exit=$?"
agent-browser get attr '#btn' data-absent --json
```

## Actual behaviour

`get attr '#btn' data-x` prints `1`. `get attr '#btn' data-absent` prints `✓ Done` and exits 0. `--json` carries `"value": null`.

## Expected behaviour

A read command should say what it read. A missing attribute should print an explicit marker distinct from both an attribute whose value is the empty string and from the mutating-command success line.

## Root cause

`cli/src/output.rs:724`, in `print_primary_response`.

The plain-text formatter reaches `if let Some(value) = data.get("value").and_then(|v| v.as_str())` and prints the string when one is present. A missing attribute comes back as JSON `null`, so `as_str()` returns `None`, the branch is skipped, and control falls through every remaining formatter to the generic terminal case at `cli/src/output.rs:1531`, `println!("{} Done", color::success_indicator())`. That line exists for mutating commands such as `click` and `fill`, which have nothing to print on success. Nothing in the path knows the action was `getattribute`, so a read with no result is rendered exactly like a write that succeeded. The empty-string case and the absent case are also indistinguishable in plain text, since one prints an empty line and the other prints `Done`.

## Impact

An AI agent that checks for an attribute, for example a `disabled` flag, an `aria-expanded` state or a `data-testid`, and parses stdout sees the success line every other command emits and can read it as a successful confirmation rather than as an absent attribute. The exit code is 0 in both cases, so there is no second signal to fall back on. An agent scripting a sequence of reads and writes has to know, out of band, which commands are reads in order to interpret `✓ Done` correctly.

dwin-gharibi:
## Summary

`get attr` on an element without the requested attribute fell through every formatter to the shared `✓ Done` line that mutating commands print on success. It now prints `(not present)`, distinct from both an empty attribute and a write confirmation.

## Closes

Closes #1938.

## What changed

- `cli/src/output.rs`: new `format_attribute_text` handles the `getattribute` action only. A string value renders as itself, so an empty attribute still prints an empty line; a `null` or absent `value` renders as `(not present)`; any other JSON value renders through `to_string`.
- `print_primary_response` calls `format_attribute_text` before the generic string-value branch and returns early, so the `getattribute` action can no longer reach the terminal `Done` line.
- `cli/src/output.rs`: the `get` help text for the `attr` subcommand notes the `(not present)` output.

## Tests

The bug was reproduced end to end first against the shipped 0.37.1 binary and real Chrome, where `get attr '#btn' data-absent` printed `✓ Done` with exit 0 while `get attr '#btn' data-x` printed `1`.

Unit tests in `cli/src/output.rs`: `test_format_attribute_text_for_present_attribute` asserts a present attribute still prints its bare value; `test_format_attribute_text_for_absent_attribute` asserts `null` renders as `(not present)` and explicitly that the output does not contain `Done`; `test_format_attribute_text_distinguishes_empty_from_absent` asserts an empty-string attribute renders as an empty line and is not the same output as an absent one; `test_format_attribute_text_ignores_other_actions` asserts the formatter declines other actions and a missing action, so no other command's rendering changes; `test_get_attr_json_shape_distinguishes_each_case` asserts the JSON payloads for present, empty and absent stay three distinct serializations.

## Docs

`README.md` notes that `get attr` prints `(not present)` when the attribute is absent.

## Behaviour changes

Plain-text `get attr` on a missing attribute prints `(not present)` instead of `✓ Done`. Present and empty attributes are unchanged, the exit code stays 0, and the `--json` shape is unchanged, so a caller already reading `"value": null` is unaffected.

## Risk

Low. The new formatter is scoped to the `getattribute` action, and the only output that changes is the one that was previously indistinguishable from a mutating command's success line.


dwin-gharibi:
@ctate

PR #1944, opened 2026-09-18
fix(set): reject non-finite viewport scale and geo coordinates

### LINKED ISSUE #1943: `set viewport 800 600 nan` reports success but silently uses a scale of 1.0
## Summary

`set viewport <w> <h> [scale]` parses the scale with a plain float parse, which accepts `nan` and `inf`. A non-finite float cannot be represented in JSON, so it serialises to `null` on the way to the daemon, the scale is dropped, and the default 1.0 is used while the command still reports success.

## Severity

Medium.

## Affected version

## Steps to reproduce

```bash
agent-browser set viewport 800 600 nan
echo "exit=$?"
agent-browser eval "window.devicePixelRatio"

agent-browser set viewport 800 600 inf
echo "exit=$?"
agent-browser eval "window.devicePixelRatio"

agent-browser set viewport 800 600 -2; echo "exit=$?"
agent-browser set viewport 800 600 2;  echo "exit=$?"
agent-browser eval "window.devicePixelRatio"
```

## Actual behaviour

`set viewport 800 600 nan` and `set viewport 800 600 inf` both exit 0 and leave `deviceScaleFactor` at 1.0. The neighbouring cases work as expected: `-2` exits 1 with `deviceScaleFactor must be non-negative`, and `2` gives 2.0.

## Expected behaviour

A scale that cannot be sent to the browser should be rejected with a non-zero exit and a message naming the offending input, the same way a negative scale already is. A command reporting success should have applied the value it was given.

## Root cause

`cli/src/commands.rs:3280`, in `parse_set` for the `viewport` subcommand.

The scale is read with `scale_str.parse::<f64>()`, whose error arm only covers unparseable text. Rust's float parser accepts `nan`, `inf`, `-inf` and their case variants as valid `f64` values, so they pass. The result goes straight into `cmd["deviceScaleFactor"] = json!(scale)`, and `serde_json` cannot represent a non-finite float, so it becomes JSON `null`. The daemon sees a null `deviceScaleFactor`, treats it as absent, and applies the default 1.0. The rejection for a negative scale that the user does see comes from further downstream, not from this parse, which is why `-2` fails loudly while `nan` succeeds quietly. The same plain parse is used for the `geo` subcommand's latitude and longitude at `cli/src/commands.rs:3306` and `cli/src/commands.rs:3312`, so `set geo` had the same exposure.

## Impact

An AI agent setting up a retina or high-density viewport before capturing a screenshot gets exit 0 and a viewport at 1.0. Every later capture and every measurement taken from the page is then at the wrong scale, and nothing in the output says so. The failure is worse than a plain rejection because the agent has an explicit success signal telling it the environment is configured as requested, so it will not retry or check.

dwin-gharibi:
## Summary

`set viewport <w> <h> [scale]` accepted `nan` and `inf`, which serialise to JSON `null` and are dropped on the way to the daemon, so the scale silently became 1.0 while the command exited 0. Non-finite floats are now rejected where the error is local, and `set geo` gets the same guard.

## Closes

Closes #1943.

## What changed

- `cli/src/commands.rs`: new `parse_finite_f64` parses a float argument and returns `ParseError::InvalidValue` both for unparseable text and for a value that is not finite, quoting the offending input and carrying the caller's usage string.
- `parse_set` uses `parse_finite_f64` for the `viewport` scale, and adds an explicit local check rejecting a negative scale rather than leaving that to the daemon.
- `parse_set` uses `parse_finite_f64` for the `geo` latitude and longitude, which previously used the same plain `parse::<f64>()` and had the same exposure. This also upgrades their failure from `ParseError::MissingArguments` to `InvalidValue`, which is what an unparseable coordinate actually is.
- `cli/src/output.rs`: `set` help text states that the viewport scale must be a finite, non-negative number.

## Tests

The bug was reproduced end to end first against the shipped 0.37.1 binary and real Chrome, confirming that `set viewport 800 600 nan` and `... inf` exited 0 with `deviceScaleFactor` at 1.0 while `... 2` gave 2.0.

Unit tests in `cli/src/commands.rs`: `test_set_viewport_rejects_non_finite_scale` loops over `nan`, `inf`, `-inf`, `NaN` and `infinity` and asserts each is rejected with an error mentioning `finite` that also quotes the input back; `test_set_viewport_rejects_negative_scale` asserts `-2` is now rejected locally with a `non-negative` message; `test_set_viewport_accepts_zero_scale` pins 0 as still accepted, since zero is finite and non-negative and means "use the default" downstream; `test_set_geo_rejects_non_finite_coordinates` asserts both `set geo nan 10` and `set geo 10 inf` are rejected.

## Docs

None needed beyond the `set` help text in `cli/src/output.rs`; the README and docs pages describe `set viewport` without enumerating accepted scale values, so no prose there became wrong.

## Behaviour changes

`set viewport` and `set geo` now exit non-zero on a non-finite number instead of succeeding with the value dropped. A negative viewport scale is now rejected by the CLI, so the message changes from the daemon's `deviceScaleFactor must be non-negative` to `set viewport scale must be non-negative`; the exit code is unchanged. Finite inputs behave exactly as before.

## Risk

Low. The only inputs whose outcome changes are ones that could never have been applied, and the negative case already failed.


dwin-gharibi:
@ctate

PR #1947, opened 2026-09-18
perf(cli): drop the 150ms settle sleep before every command

leeguooooo:
ensure_daemon slept 150ms and re-probed the socket on every call, even
when the daemon was healthy, to dodge a daemon that was mid-shutdown.
That was ~95% of a warm command's latency (eval: 167ms -> 11ms median).

The close path now unlinks the socket before its 100ms shutdown delay,
so a successful connect already means the daemon is still serving and
the recheck is unnecessary.

Verified: 30 interleaved warm evals, 5 close-then-immediate-reopen
rounds, connection/daemon unit tests. On a Jev-driven Google Flights
task (about 100 browser calls) median time fell 23.2s -> 16.2s.

Co-Authored-By: Claude <noreply@anthropic.com>


leeguooooo:
Opened against the wrong repository by mistake, sorry for the noise.

